PR #2, opened 2019-06-05
mac: Set a valid interface number on hid_device_info for USB HID dev…

z3ntu:
…ices

Previously the interface would never be set on Mac.

This presents a big pain because retrieving interface numbers can be the
only way to distinguish between the interfaces returned by HIDAPI.

This change makes it possible to retrieve interface number from an
hid_device_info on Mac for USB HID devices only.

It is unclear if the Mac OS IOKit library returns valid interface numbers for non-HID USB devices. Because of this, I have opted to simply skip that case - leave it initialised to `-1`.

In the future, we can easily relax this restriction if it turns out IOKit correctly returns interface number with non-HID USB devices. For now, this PR brings 90% of the value at 5% of the risk.

Re-open of signal11/hidapi#380

### DIFF
--- a/mac/hid.c
+++ b/mac/hid.c
@@ -25,6 +25,7 @@
 #include <IOKit/hid/IOHIDManager.h>
 #include <IOKit/hid/IOHIDKeys.h>
 #include <IOKit/IOKitLib.h>
+#include <IOKit/usb/USBSpec.h>
 #include <CoreFoundation/CoreFoundation.h>
 #include <wchar.h>
 #include <locale.h>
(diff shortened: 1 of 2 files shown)


mac/hid.c
```diff
@@ -479,7 +483,16 @@ struct hid_device_info  HID_API_EXPORT *hid_enumerate(unsigned short vendor_id,
 			cur_dev->release_number = get_int_property(dev, CFSTR(kIOHIDVersionNumberKey));
 
 			/* Interface Number (Unsupported on Mac)*/
```
```suggestion
```

Qbicz:
Thanks @todbot for merging this. However, can we use "rebase & merge" strategy to keep linear history? There are many advantages, the most important for me is possibility to `git bisect` and `git revert` or `git cherry-pick` any commit. Actually I think we can only support "rebase & merge" in settings of repo, I will change that.

Youw:
> I assume this doc

A good start. Try also:
https://www.git-tower.com/learn/git/ebook/en/command-line/advanced-topics/rebase

@Qbicz I suggest also allowing `squash-merge`, so `force-push` is not mandatory for PRs.

Qbicz:
@todbot yes, it's a good place. Basically, you apply your commits on top of the existing history and resolve any conflicts in your commit that introduces the change.

I have changed settings to allow `squash-merge` and `rebase-merge`.

PR #3, opened 2019-06-05
mac: fix behaviour of hid_get_feature_report (was writing on the repo…

z3ntu:
…rt id byte) and return value was not right)

Re-open of signal11/hidapi#219


Qbicz:
It looks reasonable to make the behaviour similar on Mac as it is on Windows and Linux.

PR #4, opened 2019-06-05
Explicit cast from void* to corresponding type

iakov:
Fix compilation without -fpermissive
Originally was PR https://github.com/signal11/hidapi/pull/430

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -737,7 +737,7 @@ static void read_callback(struct libusb_transfer *transfer)
 	if (transfer->status == LIBUSB_TRANSFER_COMPLETED) {
 
 		struct input_report *rpt = (struct input_report*) malloc(sizeof(*rpt));
-		rpt->data = (uint8_t *) malloc(transfer->actual_length);
+		rpt->data = (uint8_t*) malloc(transfer->actual_length);
 		memcpy(rpt->data, transfer->buffer, transfer->actual_length);
 		rpt->len = transfer->actual_length;
 		rpt->next = NULL;
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -182,7 +182,7 @@ static int return_data(hid_device *dev, unsigned char *data, size_t length);
 
 static hid_device *new_hid_device(void)
 {
-	hid_device *dev = calloc(1, sizeof(hid_device));
+	hid_device *dev = (hid_device*) calloc(1, sizeof(hid_device));
 	dev->blocking = 1;
 
 	pthread_mutex_init(&dev->mutex, NULL);
@@ -430,7 +430,7 @@ static wchar_t *get_usb_string(libusb_device_handle *dev, uint8_t idx)
 
 	   Skip over the first character (2-bytes).  */
 	len -= 2;
-	str = malloc((len / 2 + 1) * sizeof(wchar_t));
+	str = (wchar_t*) malloc((len / 2 + 1) * sizeof(wchar_t));
 	int i;
 	for (i = 0; i < len / 2; i++) {
 		str[i] = buf[i * 2 + 2] | (buf[i * 2 + 3] << 8);
@@ -563,7 +563,7 @@ struct hid_device_info  HID_API_EXPORT *hid_enumerate(unsigned short vendor_id, 
 							struct hid_device_info *tmp;
 
 							/* VID/PID match. Create the record. */
-							tmp = calloc(1, sizeof(struct hid_device_info));
+							tmp = (struct hid_device_info*) calloc(1, sizeof(struct hid_device_info));
 							if (cur_dev) {
 								cur_dev->next = tmp;
 							}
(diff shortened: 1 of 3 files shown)


Qbicz:
In C, void* should be promoted to any pointer type without problem. Do you really compile this code with C++ compiler?

mac/hid.c
```diff
@@ -226,7 +226,7 @@ static int get_string_property(IOHIDDeviceRef device, CFStringRef prop, wchar_t
 	if (!len)
 		return 0;
 
-	str = IOHIDDeviceGetProperty(device, prop);
+	str = (CFStringRef)IOHIDDeviceGetProperty(device, prop);
```
Qbicz: when casting, can you keep convention `(type) variable_name;` <- please keep the space at all places

Youw:
> In C, void* should be promoted to any pointer type without problem

At certain compilation flags, even C compiler can start giving warning for the implicit pointer casts, which becomes a real problem with `-Werror` or smth.
For some projects that's a big dial, and they might use builds systems of their own.

LudovicRousseau:
Please provide a copy of a C (not C++) compiler warning for this case.
See also https://en.wikipedia.org/wiki/C_dynamic_memory_allocation#Type_safety

For my own code I always include the warning or error message reported by the compiler in the commit message. So that it is clear what problem is solved when reviewing the PR or the commit 10 years later.

mac/hid.c
iakov: Done

Youw:
My first thought was about `-Wconversion` which, apparently isn't the case for `malloc` (i.e. casting from `void*`) and C compiler.

There is `-Wc++-compat` which is quite verbose regarding compatibility with C++, and it gives me a second thought.

@iakov why do you need have the ability to compile this source using C++ compiler?

iakov:
Well, we use C++ as a main language for https://github.com/trikset/trik-studio , and `*/hid.c` files were included into `.cpp`, so it is a C++ related issue. Nevertheless, if this does not blow up code, and looks correct from pure C point of view, I hope these edits are useful for those like me (whoe includes .c files into C++ code).

mac/hid.c
```diff
@@ -296,7 +296,8 @@ static wchar_t *dup_wcs(const wchar_t *s)
 static io_service_t hidapi_IOHIDDeviceGetService(IOHIDDeviceRef device)
 {
 	static void *iokit_framework = NULL;
-	static io_service_t (*dynamic_IOHIDDeviceGetService)(IOHIDDeviceRef device) = NULL;
+	typedef io_service_t (*dynamic_IOHIDDeviceGetService_t)(IOHIDDeviceRef device);
+	static dynamic_IOHIDDeviceGetService_t dynamic_IOHIDDeviceGetService;
```
```suggestion
	static dynamic_IOHIDDeviceGetService_t dynamic_IOHIDDeviceGetService = NULL;
```

mac/hid.c
```diff
@@ -682,6 +683,7 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
 {
 	hid_device *dev = NULL;
 	io_registry_entry_t entry = MACH_PORT_NULL;
+	IOReturn ret;
```
```suggestion
	IOReturn ret = kIOReturnSuccess;
```

libusb/hid.c
```diff
@@ -182,7 +182,7 @@ static int return_data(hid_device *dev, unsigned char *data, size_t length);
 
 static hid_device *new_hid_device(void)
 {
-	hid_device *dev = calloc(1, sizeof(hid_device));
+	hid_device *dev = (hid_device*) calloc(1, sizeof(hid_device));
```
Qbicz: That would be good if you stick to one notation - either use `(hid_device*)` or `(hid_device *)` at all places. Otherwise, this looks good

mac/hid.c
iakov: Committed initialization, but with kIOReturnInvalid to prevent occasional future misuse.

libusb/hid.c
```diff
@@ -736,8 +736,8 @@ static void read_callback(struct libusb_transfer *transfer)
 
 	if (transfer->status == LIBUSB_TRANSFER_COMPLETED) {
 
-		struct input_report *rpt = malloc(sizeof(*rpt));
-		rpt->data = malloc(transfer->actual_length);
+		struct input_report *rpt = (struct input_report*) malloc(sizeof(*rpt));
+		rpt->data = (uint8_t *) malloc(transfer->actual_length);
```
Qbicz: `(uint8_t*)` to make it aligned with other usages

mac/hid.c
Youw: yes, that's fine too

libusb/hid.c
Qbicz:
```suggestion
		rpt->data = (uint8_t*) malloc(transfer->actual_length);
```

PR #7, opened 2019-06-05
configure.ac: don't use +=

z3ntu:
The += operator is bashism and won't work with non-bash shells, such as
ash on Alpine Linux.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,14 +63,14 @@ case $host in
 
 	# HIDAPI/hidraw libs
 	PKG_CHECK_MODULES([libudev], [libudev], true, [hidapi_lib_error libudev])
-	LIBS_HIDRAW_PR+=" $libudev_LIBS"
-	CFLAGS_HIDRAW+=" $libudev_CFLAGS"
+	LIBS_HIDRAW_PR="${LIBS_HIDRAW_PR} $libudev_LIBS"
+	CFLAGS_HIDRAW="${CFLAGS_HIDRAW} $libudev_CFLAGS"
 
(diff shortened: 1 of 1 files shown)


z3ntu:
There are still 4 instances of a += behind a `if darwin` which maybe also should be fixed. The default shell in macOS is bash but I've read something that you will be able to choose e.g. dash to be the default shell in an upcoming macOS release. But for that probably more stuff has to be fixed anyways.

Youw:
> There are still 4 instances of a += behind a if darwin

So, maybe, update that too?

z3ntu:
Changed for macOS too. I didn't want to change it first because I don't have this fox stuff installed in my VM but it should be fine

Youw:
@LudovicRousseau, @Qbicz are we going to follow any specific guide on commit messages?

LudovicRousseau:
I don't plan to work on hidapi. I just helped to move the project to a new home.
Do what you want.

Qbicz:
I would stick to
`module: What this change does`, for example `mac: Fix error in foo`. Also we should require to always provide description - why this change is needed and how it fixes the problem.

@todbot @Youw @jdk I was also thinking about how long the PR needs to be in review before we merge it - I would opt for 24 hours at least for small fixes and maybe 3 days for bigger changes. This is to let all interested people take a look and requested changes if the new code is not good enough. If you agree, where do you think is the proper place to put such guidelines?

Youw:
> If you agree, where do you think is the proper place to put such guidelines?

CONTRIBUTIONS.md ?

PR #8, opened 2019-06-05
Windows: enumerate simple devices

Qbicz:
Right now, in hid_open it is checked if the connected devices have
"HIDClass" driver name. This is true for composite devices, but some
simple HID devices only have "Mouse" or "Keyboard" driver name.

One of devices that cannot be opened is Dell MS116t, but any mouse which
is not part of composite device can be affected.

With this change, one can enumerate and open these devices.

Signed-off-by: Kubicz, Filip <filip.kubicz@nordicsemi.no>


Qbicz:
Originally pull request https://github.com/signal11/hidapi/pull/418

PR #13, opened 2019-06-08
README.txt: update number of backends

emaste:
FreeBSD support was added in commit 74440e2d as the fifth back-end, but
the list was still introduced as "HIDAPI has four back-ends."


Youw:
Technically FreeBSD is not a backend - it is a running platform.

The 4 backends:
- WinAPI(hid.dll);
- libusb(Linux/FreeBSD)
- hidraw(Linux);
- IOHidManager(macOS);

I think such description would be more accurate.

Issue #6, opened 2019-06-05
Continuous integration


z3ntu:
I'm working on CI scripts for:

| Target | CI provider | Duration | Notes |
|--------|----------| -- | -- |
|msbuild|AppVeyor|20 sec||
|cygwin|AppVeyor|1 min 45 sec||
|mingw|builds.sr.ht|70 sec|Cross-compiled on Fedora|
|freebsd|builds.sr.ht|55 sec||
|macos|Travis CI|53 sec||
|alpine|builds.sr.ht|30 sec||
|archlinux|builds.sr.ht|28 sec||

This makes sure at least that hidapi builds for most/all supported platforms.

Qbicz:
That's very interesting, please share the results when you have the CI setup ready.

Which of these platforms can be integrated with GitHub, e.g. to run when a pull-request is opened?
Also, would it be possible to download artifacts from them? (ideally shared libraries for Windows and Mac)

z3ntu:
You can see the results for all the platforms listed above at https://github.com/z3ntu/hidapi/commits/ci . All of them are fully integrated into GitHub ;)

Artifacts are definitely possible on AppVeyor, but I'm not sure if Travis supports artifacts natively. But honestly, it's really easy to compile hidapi on macOS, so I wouldn't actually bother with that. Windows, sure, it's a pain otherwise.

Qbicz:
Sounds good!

What about attaching the Windows DLL to the next release? Do you think it would be a good idea? (I can provide MSVC-built library). Then having the artifacts is not essential.

z3ntu:
I also have https://github.com/signal11/hidapi/pull/324 on my CI branch which is needed for the Windows build. 

I'll work on the Windows dll artifacts tomorrow!

Youw:
I'd suggest switching to CMake instead of having MSVC2015/2017/2019/etc. projects.
Personally I'm using CMake for all my projects on all platforms (Windows/Linux/macOS/cross-aarch64).

z3ntu:
Windows DLL artifacts now work: https://ci.appveyor.com/project/z3ntu/hidapi/builds/25095767/job/q9378cp84do5qs49/artifacts

Youw:
I'd say having only a DLL as an artifact is not enough.
Perfect case, if we have .dll, .lib and .pdb - that's very common for Windows.

z3ntu:
[...]

PR #15, opened 2019-06-09
Add .cirrus.yml config for Cirrus-CI continuous integration

emaste:
Starting with Alpine and FreeBSD 11 and 12 tasks, others to be added.

Result from my fork: https://cirrus-ci.com/build/5686867601653760

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,33 @@
+alpine_task:
+        container:
+                image: alpine:latest
+        install_script: apk add autoconf automake g++ gcc libusb-dev libtool linux-headers eudev-dev make musl-dev
+        script:
+                - ./bootstrap
+                - ./configure || { cat config.log; exit 1; }
+                - make
+                - make install
+
+freebsd11_task:
+        freebsd_instance:
+                image: freebsd-11-2-release-amd64
+        install_script:
+                - pkg install -y
+                  autoconf automake libiconv libtool pkgconf
+        script:
+                - ./bootstrap
+                - ./configure || { cat config.log; exit 1; }
+                - make
+                - make install
+
+freebsd12_task:
+        freebsd_instance:
+                image: freebsd-12-1-release-amd64
+        install_script:
+                - pkg install -y
+                  autoconf automake libiconv libtool pkgconf
+        script:
+                - ./bootstrap
+                - ./configure || { cat config.log; exit 1; }
+                - make
+                - make install


z3ntu:
See also #6

Youw:
Even with #6, this does looks like an improvement. @z3ntu or do you suggest to merge the effort, or smth?

Also, in order to actually have this working, someone from libusb organisation has to give access to Cirrus-CI to this repository.

z3ntu:
I'd rather use the builds.sr.ht service as it's open-source and could  theoretically be self-hosted (and I doubt cirrus ci is open-source).

Qbicz:
Since we already have Alpine and FreeBSD, I would suggest to close this PR.

@z3ntu do you see value in having Cirrus configuration available? Or in testing multiple versions of FreeBSD as @emaste did?

z3ntu:
builds.sr.ht also has support for FreeBSD 11.x and FreeBSD CURRENT if we want that. Our freebsd builds use `freebsd/latest` which is FreeBSD 12 currently. So if you want that, I can add them.

But I think we're good with the current set of CIs (Alpine Linux with musl libc; Arch Linux with glibc, FreeBSD and three variants of Windows (mingw, cygwin & msbuild))

What could be done is to create "release candidate" CI configuration that makes a build for each OS supported on builds.sr.ht (Alpine Linux, Arch Linux, Debian, Fedora, FreeBSD, NixOS, OpenBSD, Ubuntu) and all versions supported there if someone has too much free time on their hands^^

emaste:
builds.sr.ht sounds like a good approach, although IMO there's not much downside of having a configuration for Cirrus-CI in the tree also, as with .travis.yml / .appveyor.yml; people doing development on their fork may find it convenient.

Qbicz:
@youw, @z3ntu, I support merging this if you agree. We don't have to have Cirrus set up for this repo, this could be for contributors convenience.

.cirrus.yml
```diff
@@ -0,0 +1,29 @@
+alpine_task:
+        container:
+                image: alpine:latest
+        install_script: apk add autoconf automake g++ gcc libusb-dev libtool linux-headers eudev-dev make musl-dev
+        script:
+                - ./bootstrap
+                - ./configure || { cat config.log; exit 1; }
```
z3ntu: Why does this configure statement have the statement behind but not the others?

emaste: Oh, I probably added that while trying to track down some problem building on alpine, and did not run into a problem on the other OSes that required more details. It's probably a good idea for all cases and I'll update the PR.

emaste:
Here's an Alpine, FreeBSD 11, FreeBSD 12 build: https://cirrus-ci.com/build/4977906178064384

Youw:
One last note before merge: can we move this file to a `.builds` subfolder?

emaste:
> One last note before merge: can we move this file to a `.builds` subfolder?

No, Cirrus-CI only looks at the root dir of the repository (as does Travis-CI etc.)

Youw:
I see, OK, let it be then.

PR #16, opened 2019-06-09
Fix documentation generation

z3ntu:
The first commit is auto-generated with `doxygen -u`

The second commit fixes most of the doc generation being wrong, because the "relative indentation" in hidapi.h comments was being intepreted as codeblocks. 

The third commit fixes a warning because in the hid_read_timeout documentation the parameter-name "device" was used but the function used "dev". As all other functions use "device", adjust the function.

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -195,7 +195,7 @@ extern "C" {
 			the Control Endpoint (Endpoint 0).
 
 			@ingroup API
-			@param device A device handle returned from hid_open().
+			@param dev A device handle returned from hid_open().
 			@param data The data to send, including the report number as
 				the first byte.
 			@param length The length in bytes of the data to send.
@@ -204,7 +204,7 @@ extern "C" {
 				This function returns the actual number of bytes written and
 				-1 on error.
 		*/
-		int  HID_API_EXPORT HID_API_CALL hid_write(hid_device *device, const unsigned char *data, size_t length);
+		int  HID_API_EXPORT HID_API_CALL hid_write(hid_device *dev, const unsigned char *data, size_t length);
 
 		/** @brief Read an Input report from a HID device with timeout.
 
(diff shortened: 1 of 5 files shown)


Youw:
The result looks good, but `doxygen -u` gives so unreadable diff.
Maybe manually change only what need to be updated (remove obsolete/add missing options)?

z3ntu:
That's why I made them separate commits - but future updates to the Doxyfile should be much smaller as this is a bump over several years of versions (if it's needed at all).

hidapi/hidapi.h
```diff
@@ -221,7 +221,7 @@ extern "C" {
 				-1 on error. If no packet was available to be read within
 				the timeout period, this function returns 0.
 		*/
-		int HID_API_EXPORT HID_API_CALL hid_read_timeout(hid_device *dev, unsigned char *data, size_t length, int milliseconds);
+		int HID_API_EXPORT HID_API_CALL hid_read_timeout(hid_device *device, unsigned char *data, size_t length, int milliseconds);
```
Qbicz: In all implementations (in .c files) there is `dev`. I would prefer to change doxygen text to `@param dev` instead of changing `dev` to `device` in function declaration.

Qbicz: And rename commit title accordingly

Youw: On the other hand - in all other places in this `.h` - it is `device`, so either we rename everywhere, or have `device` everywhere (I prefer more readable form for documentation).

Qbicz: `hid_device *dev` looks readable to me and `dev` is pretty standard abbreviation for device. But in the end, I would be okay with both options

Youw: sure
in any case - better if we keep consistency across entire file

z3ntu:
I've pushed an update that renames the `hid_device *device` parameters to `hid_device *dev` in the header file and in the `register_error` method.

Issue #11, opened 2019-06-06
Update README.txt/AUTHORS.txt


Youw:
- add notice about moving under @libusb/hidapi;
- update links if there are any;
- consider Markdown?

todbot:
+1 on Markdown

z3ntu:
Notes from #17 that should be fixed as a follow-up:

> * the text is now rather lengthy in comparison to original, which is more compressed. I feel the markdown version is overwhelming, what about having some table of contents in the beginning?
> * Title of library is broken down into 2 lines, with "X" from OS X in second line. What about using ## heading which would results in smaller letters?

Qbicz:
Fixed by #19

Youw:
AUTHORS.txt hasn't been updated yet, aren't we going to?

Qbicz:
@Youw if you have idea how AUTHORS should now look like, please go ahead with PR ;)

Youw:
I'll think of something when I'm back from a family trip. (Somewhere next week.)

PR #17, opened 2019-06-09
README: convert to markdown

z3ntu:
Ref #11 

This is a plain conversion without any changes to the text.

Preview: https://github.com/z3ntu/hidapi/blob/readme_md/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
-         HIDAPI library for Windows, Linux, FreeBSD and Mac OS X
-        =========================================================
+# HIDAPI library for Windows, Linux, FreeBSD and Mac OS X
 
-About
-======
+## About
 
 HIDAPI is a multi-platform library which allows an application to interface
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,6 @@
-         HIDAPI library for Windows, Linux, FreeBSD and Mac OS X
-        =========================================================
+# HIDAPI library for Windows, Linux, FreeBSD and Mac OS X
```
Youw:
```suggestion
# HIDAPI library for Windows, Linux, FreeBSD and macOS
```
this should shorten the title and even more accurate right now, according to Apple

z3ntu: > This is a plain conversion without any changes to the text.

:) Maybe let's do this as a follow-up change?

Youw: OK, agree - fix conflict and lets merge

Issue #12, opened 2019-06-06
CMake build system


Youw:
As a common CMake user I'd suggest it is very convenient to have CMake supported by the project off-the-box.

A platform like MSVC would gain a great benefit - won't need to have MSVC-2010/2015/2017/2019/etc. separate projects.

I'm ready to prepare the whole infrastructure if the team/community don't have anything against it.

z3ntu:
I'd rather use Meson as a build system than CMake

todbot:
I would suggest that any new build system not prevent the use of simple Makefiles for the many small *nixes out there that don't have access to Python or Internet access.

Youw:
I wouldn't suggest replacing build system. Not now.
NOTE1: Both Meson and CMake support generating Makefiles.
NOTE2: CMake doesn't require Python or Internet access.

> I'd rather use Meson

Never used it. But quick reading over its readme/some docs, and it seems there is no much benefit in using Meson over CMake.
I'd also say, that CMake is way more widely used and known - it has proved itself, specially after version 3.x.

z3ntu:
> NOTE1: Both Meson and CMake support generating Makefiles.

Meson actually generates files for the ninja build system :wink: 

In my opinion, Meson has a much better syntax than CMake and many useful features included (such as the pkg-config file generation, painless cross-compilation, etc)

There's also https://mesonbuild.com/Comparisons.html

I think it would be best to have manual Makefiles for those `many small *nixes` and a modern build system with all the fancy features you want for the rest (either cmake or meson)

Youw:
> ninja build system

I'd add it to `cons` for a project with a single `.c` file.

> manual Makefiles

What would be a benefit of that effort? `autotools` already does it very well for all originally supported platforms.

Also, as I pointed since the beginning - on of the main benefit with CMake is to have a single build system, instead of MSVC-version specific copy-paste, XCode support appears automatically too, etc.

Mason with ninja only support - wouldn't make any difference here.

> pkg-config file generation

Already done by current build system.
Trivially with CMake.

> painless cross-compilation

Trivially with CMake.

z3ntu:
Sure, we can argue back and forth between Meson and CMake and nothing would change :wink: I'm for Meson, you're for CMake. Maybe others can chime in their thoughts :)

> What would be a benefit of that effort? autotools already does it very well for all originally supported platforms.

[...]

PR #18, opened 2019-06-10
Add continuous integration config

z3ntu:
This PR is for the platforms listed at #6 except Windows (msbuild & cygwin) as that needs an upgrade of the VS project to a newer version; see #12 

See also #15 

This would need configuration of https://dispatch.sr.ht and https://travis-ci.com/ on this repo.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -1,0 +1,19 @@
+image: alpine/edge
+packages:
+- autoconf
+- automake
+- libtool
+- eudev-dev
+- libusb-dev
+- linux-headers
+sources:
+- https://github.com/libusb/hidapi
+tasks:
+- setup: |
+    cd hidapi
+    ./bootstrap
+    ./configure
+- build: |
+    cd hidapi
+    make
+    make DESTDIR=$PWD/root install
--- a/.builds/archlinux.yml
+++ b/.builds/archlinux.yml
@@ -1,0 +1,12 @@
+image: archlinux
+sources:
+- https://github.com/libusb/hidapi
+tasks:
+- setup: |
+    cd hidapi
+    ./bootstrap
+    ./configure
+- build: |
+    cd hidapi
+    make
+    make DESTDIR=$PWD/root install
--- a/.builds/fedora-mingw.yml
+++ b/.builds/fedora-mingw.yml
@@ -1,0 +1,18 @@
+image: fedora/latest
+packages:
+- autoconf
+- automake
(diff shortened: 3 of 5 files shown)


Youw:
I'd suggest to build an `install` target, to check by CI that it is correctly configured too.

z3ntu:
@Youw What should be checked there? I would think of passing autotools `--prefix=/myprefix`, then running `make DESTDIR=mydir install` and checking if `./mydir/myprefix/lib/libhidapi-hidraw.so` exists or something like that

Youw:
> --prefix=/myprefix

Yeap.

> What should be checked there?

As a sanity test - just to make sure that `make install` returns "success". Checking that `*.so` exists - may be a good check, but I wouldn't say it is mandatory. That is up to you.

Also, if we want to see artifacts at some point - artifacts from `install` target is what need to be gathered.

z3ntu:
Artifacts for *nix builds are unnecessary imo. I'll add the make install calls.

z3ntu:
Make install commands are now added to all CIs.

Qbicz:
@z3ntu should I integrate Travis and https://dispatch.sr.ht with this repo before we merge it?

z3ntu:
Yeah. And I think both CIs don't attempt to build something if the config files aren't present

Qbicz:
@LudovicRousseau can you check if you have received any email regarding Travis CI or Sourcehut (sr.ht) access to this repo? When I want authorize the CI, it says I cannot authorize it for the organization libusb and that email has been sent to the owner.

LudovicRousseau:
@Qbicz yes, I got an access request.

> Review third-party application access request
> Granting access will give this application the ability to request access to private data in the libusb organization. 

I don't know why Sourcehut would need to access the **private** data of the libusb organization.

What is Sourcehut supposed to do?

I granted access to travis-ci.com. travis-ci.org already had access.

z3ntu:
Comment from https://todo.sr.ht/~sircmpwn/dispatch.sr.ht/22:

> We need those permissions to install the webhook which notifies sr.ht of pull requests et al. There are no more conservative permissions for this.

But there's also a ticket open (but not in progress) for not granting those permissions and setting up the webhook manually: https://todo.sr.ht/~sircmpwn/dispatch.sr.ht/25

LudovicRousseau:
Approved now.

Qbicz:
Added hooks for dispatch.sr.ht and Travis.

PR #20, opened 2019-06-13
Update Makefile.am after README.md

rijandn:
The 'make' build on Linux/MacOS generates an error since README.txt was renamed to README.md. See #17 .


z3ntu:
Oops^^

PR #22, opened 2019-06-17
hidtest: don't use a C++ source file, since it's pure C

z3ntu:
This allows to build the test program with toolchains that don't have
C++ support.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>

Originally https://github.com/signal11/hidapi/pull/249


z3ntu:
The cpp file is still referenced in 
```
windows/hidtest.vcproj:                         RelativePath="..\hidtest\hidtest.cpp"
```
This should be updated as well I think?

z3ntu:
I can't 100% verify that the Windows version builds fine because of an additional commit needed for appveyor (vs2015); but in general it still builds fine: https://ci.appveyor.com/project/z3ntu/hidapi/builds/25520538/job/64xs7pjqx1mhas4q

Youw:
> additional commit needed for appveyor (vs2015)

we can fix any possible issues when it comes to it
so far looks good

Youw:
compiles on Windows with automatic conversion of `.sln` by MSVC2017

PR #23, opened 2019-06-17
hidraw: Implement hid_error()

z3ntu:
Here is my first try at implementing error messages for the Linux/hidraw implementation.

* I used a thread-local variable for non-device error messages.
* Non-device errors can be obtained by passing `NULL` into `hid_error` as I suggested in #123.
* I **belive** to have covered all system calls that could fail; please check this.

A few extra questions:

* I am not sure why in `utf8_to_wchar_t` the input is called `utf8`. As far as I know, the encoding used in `mbstowcs` depends on the current locale, if if I haven't set my locale to use UTF8, then the input is not UTF8.
* The `register_error` equivalent in `windows/hid.c` doesn't use its argument `const char *op`. Is that intended?
* In the Windows implementation, do functions like [`HidD_GetPreparsedData`](http://msdn.microsoft.com/en-us/library/windows/hardware/ff539679%28v=vs.85%29.aspx) actually set the error code for `GetLastError()`?

Originally https://github.com/signal11/hidapi/pull/125

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -151,6 +151,8 @@ extern "C" {
 			If @p serial_number is NULL, the first device with the
 			specified VID and PID is opened.
 
+			This function sets the return value of hid_error().
+
 			@ingroup API
 			@param vendor_id The Vendor ID (VID) of the device to open.
 			@param product_id The Product ID (PID) of the device to open.
@@ -168,6 +170,8 @@ extern "C" {
 			The path name be determined by calling hid_enumerate(), or a
 			platform-specific path name can be used (eg: /dev/hidraw0 on
 			Linux).
+
+			This function sets the return value of hid_error().
 
 			@ingroup API
 		    @param path The path name of the device to open
@@ -194,6 +198,8 @@ extern "C" {
 			one exists. If it does not, it will send the data through
 			the Control Endpoint (Endpoint 0).
 
+			This function sets the return value of hid_error().
+
 			@ingroup API
 			@param dev A device handle returned from hid_open().
 			@param data The data to send, including the report number as
@@ -211,6 +217,8 @@ extern "C" {
 			Input reports are returned
 			to the host through the INTERRUPT IN endpoint. The first byte will
 			contain the Report number if the device uses numbered reports.
+
+			This function sets the return value of hid_error().
 
 			@ingroup API
 			@param dev A device handle returned from hid_open().
@@ -232,6 +240,8 @@ extern "C" {
 			Input reports are returned
 		    to the host through the INTERRUPT IN endpoint. The first byte will
 			contain the Report number if the device uses numbered reports.
+
+			This function sets the return value of hid_error().
 
 			@ingroup API
 			@param dev A device handle returned from hid_open().
@@ -282,6 +292,8 @@ extern "C" {
 			report data (16 bytes). In this example, the length passed
 			in would be 17.
 
+			This function sets the return value of hid_error().
+
 			@ingroup API
 			@param dev A device handle returned from hid_open().
 			@param data The data to send, including the report number as
@@ -303,6 +315,8 @@ extern "C" {
 			still contain the Report ID, and the report data will
 			start in data[1].
 
+			This function sets the return value of hid_error().
+
 			@ingroup API
 			@param dev A device handle returned from hid_open().
 			@param data A buffer to put the read data into, including
@@ -321,6 +335,8 @@ extern "C" {
 		int HID_API_EXPORT HID_API_CALL hid_get_feature_report(hid_device *dev, unsigned char *data, size_t length);
 
 		/** @brief Close a HID device.
+
+			This function sets the return value of hid_error().
 
 			@ingroup API
 			@param dev A device handle returned from hid_open().
@@ -378,8 +394,18 @@ extern "C" {
 
 		/** @brief Get a string describing the last error which occurred.
 
+			Whether a function sets the last error is noted in its
+			documentation. These functions will reset the last error
+			to NULL before their execution.
+
+			Strings returned from hid_error() must not be freed by the user!
+
+			This function is thread-safe, and error messages are thread-local.
+
 			@ingroup API
-			@param dev A device handle returned from hid_open().
+			@param dev A device handle returned from hid_open(),
+			  or NULL to get the last non-device-specific error
+			  (e.g. for errors in hid_open() itself).
 
 			@returns
 				This function returns a string containing the last error
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -75,8 +75,12 @@ struct hid_device_ {
 	int device_handle;
 	int blocking;
 	int uses_numbered_reports;
+	wchar_t *last_error_str;
 };
 
+/* Global error message that is not specific to a device, e.g. for
+   hid_open(). It is thread-local like errno. */
+__thread wchar_t *last_global_error_str = NULL;
 
 static __u32 kernel_version = 0;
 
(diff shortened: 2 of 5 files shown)


Youw:
When a client code gets an "error string" as `char*` (or, in our case `wchar_t*`), it likely assumes, that that is a valid c-string, and for instance try to:
```cpp
printf("Last HIDAPI call status: %S", hide_error());
```
and that would immediately lead to crash, since `hide_error()` sometime returns `NULL`.

I'd suggest always return a valid c-string (non-NULL) instead.
If there is no error, we might return `""` or `"Success"`, or smth like that.

Youw:
my suggestion stands

@z3ntu if you accept - please update PR accordingly, otherwise give your note why you wouldn't do it within this PR; maybe we'll have a separate task for it

z3ntu:
Yes, will update it. I've just been travelling a lot in the past couple of days.

Do you think it's better that if the `register_global_error` function receives a NULL argument, that it sets the `last_global_error_str` variable to `Success` or that the `register_global_error` function should be called with parameter `Success`? I'm personally for the first option.

Youw:
- first looks OK
- `register_global_success` is an alternative
- `hid_error` could return "Success" if `last_global_error_str` is NULL

to be honest I'm more concerned of usage on the user side, and I'll support any reasonable good implementation, so it is up to you to decide
(and yes, I think passing "Success" in each place wouldn't look very nice)

z3ntu:
The last sounds the best imo

linux/hid.c
```diff
@@ -639,15 +690,19 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
 
 		/* Get Report Descriptor Size */
 		res = ioctl(dev->device_handle, HIDIOCGRDESCSIZE, &desc_size);
-		if (res < 0)
-			perror("HIDIOCGRDESCSIZE");
-
+		if (res < 0) {
+			char errormsg[100];
+			snprintf(errormsg, sizeof(errormsg), "ioctl (GRDESCSIZE): %s", strerror(errno));
+			register_device_error(dev, errormsg);
```
z3ntu: I'm not very fond of this construct being present in multiple places. Should I create a helper function for that? If yes, name suggestions?

linux/hid.c
```diff
@@ -639,15 +690,19 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
...
+		}
 
 		/* Get Report Descriptor */
 		rpt_desc.size = desc_size;
 		res = ioctl(dev->device_handle, HIDIOCGRDESC, &rpt_desc);
 		if (res < 0) {
-			perror("HIDIOCGRDESC");
+			char errormsg[100];
+			snprintf(errormsg, sizeof(errormsg), "ioctl (GRDESC): %s", strerror(errno));
```
z3ntu: I've taken the opportunity to also improve the messages here to match the ioctl messages further below :)

z3ntu:
Should be good for a review now

linux/hid.c
Youw: `register_device_ioctl_error`?

linux/hid.c
```diff
@@ -269,7 +303,7 @@ static int get_device_string(hid_device *dev, enum device_string_id key, wchar_t
 	/* Create the udev object */
 	udev = udev_new();
 	if (!udev) {
-		printf("Can't create udev\n");
+		register_global_error("Couldn't create udev context\n");
```
Youw:
```suggestion
		register_global_error("Couldn't create udev context");
```

linux/hid.c
```diff
@@ -395,7 +431,7 @@ struct hid_device_info  HID_API_EXPORT *hid_enumerate(unsigned short vendor_id,
...
```
Youw:
```suggestion
		register_global_error("Couldn't create udev context");
```

windows/hid.c
```diff
@@ -881,7 +881,11 @@ int HID_API_EXPORT_CALL HID_API_CALL hid_get_indexed_string(hid_device *dev, int
 
 HID_API_EXPORT const wchar_t * HID_API_CALL  hid_error(hid_device *dev)
 {
-	return (wchar_t*)dev->last_error_str;
+	if (dev)
+		return (wchar_t*)dev->last_error_str;
```
Youw: `L"Success"` if NULL here too

linux/hid.c
z3ntu: I've gone for `static void register_device_error_format(hid_device *dev, const char *format, ...)`

linux/hid.c
```diff
@@ -130,6 +135,48 @@ static wchar_t *utf8_to_wchar_t(const char *utf8)
...
+	va_list args;
+	va_start(args, format);
+
+	char msg[100];
+	vsnprintf(msg, sizeof(msg), format, args);
+
+	va_end(args);
+
+	register_device_error(dev, msg);
+}
```
z3ntu: Can someone confirm that this C code is written correctly and I didn't make some fundamental mistake with how the `char msg[100]` is handled? :D

Youw: LGTM

Youw:
@jdk, @todbot, @Qbicz your comments?

linux/hid.c
```diff
@@ -130,6 +135,48 @@ static wchar_t *utf8_to_wchar_t(const char *utf8)
...
+ * The given error message will be copied (and decoded according to the
+ * currently locale, so do not pass in string constants).
+ * The last stored global error message is freed.
+ * Use register_device_error(device, NULL) to indicate "no error". */
+static void register_device_error(hid_device *dev, const char *msg)
+{
+	if (dev->last_error_str)
+		free(dev->last_error_str);
+
+	dev->last_error_str = msg? utf8_to_wchar_t(msg): NULL;
```
Qbicz: `msg` will be evaluated to `false` when you call `register_device_error(NULL);` so I think you could change
```
dev->last_error_str = msg? utf8_to_wchar_t(msg): NULL;
```
to 
```
dev->last_error_str = msg;
```

Qbicz: Sorry, I forgot about need to call utf8_to_wchar_t. So the `? :` is needed.

linux/hid.c
Qbicz: You pass the pointer to local `msg[100]` which is on stack, but in `register_device_error()` it is copied to `dev->last_error_str` which is dynamically allocated and it's lifespan is the same as lifespan of a device. This is copied before the stack is freed. So it's fine ;)

z3ntu: Great! :)

linux/hid.c
z3ntu: Applied
```diff
diff --git a/linux/hid.c b/linux/hid.c
index 6ce2d9e..4b6cad3 100644
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -146,7 +146,7 @@ static void register_global_error(const char *msg)
        if (last_global_error_str)
                free(last_global_error_str);
 
-       last_global_error_str = msg? utf8_to_wchar_t(msg): NULL;
+       last_global_error_str = utf8_to_wchar_t(msg);
 }
 
 
@@ -160,7 +160,7 @@ static void register_device_error(hid_device *dev, const char *msg)
        if (dev->last_error_str)
                free(dev->last_error_str);
 
-       dev->last_error_str = msg? utf8_to_wchar_t(msg): NULL;
+       dev->last_error_str = utf8_to_wchar_t(msg);
 }
 
 /* See register_device_error, but you can pass a format string into this function. */
```

Youw: looks good

hidapi/hidapi.h
```diff
@@ -378,8 +394,18 @@ extern "C" {
 
 		/** @brief Get a string describing the last error which occurred.
 
+			Whether a function sets the last error is noted in its
+			documentation. These functions will reset the last error
+			to NULL before their execution.
+
+			Strings returned from hid_error() must not be freed by the user!
+
+			This function is thread-safe, and error messages are thread-local.
```
Youw: And it is not thread-safe, when you're using same device from different threads.

PR #25, opened 2019-06-17
hidraw: only print errors when DEBUG_PRINTF is defined

z3ntu:
Without this PR, the hidraw implementation prints messages to stderr, which the application using this library has no control over. This should be optional, just like it is in the libusb backend.

Originally https://github.com/signal11/hidapi/pull/351

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -54,6 +54,12 @@
 #define HIDIOCGFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x07, len)
 #endif
 
+#ifdef DEBUG_PRINTF
+#define DEBUG_PERROR(...) perror(__VA_ARGS__)
+#else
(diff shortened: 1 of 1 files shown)


Youw:
There're still `printf`(s)

z3ntu:
Those printf messages should be nicely handle-able with #23

z3ntu:
This kernel version check is necessary for one workaround for kernels < 2.6.34. As this release is over 9 years old, we can probably remove that workaround now...

And the `printf("Can't create udev\n");` calls should be replaceable with #23

Youw:
Sure, why not - will you update it?

Youw:
@z3ntu please give your comment

z3ntu:
Yeah. Let's keep this PR on ice for now until #23 is ready and then adjust this one for hid_error function

z3ntu:
This is obsoleted by #23 now. Closing.

PR #26, opened 2019-06-18
windows: Fix undeclared variable name in register_error()

Qbicz:
Signed-off-by: Kubicz, Filip <Filip.Kubicz@nordicsemi.no>


Qbicz:
In #16, we forgot to update variable names in `register_error()`.

@z3ntu can you review?

PR #31, opened 2019-06-20
WIP: Support for OpenBSD

z3ntu:
OpenBSD currently have a patchset on hidapi in their repos but it was pretty unusable for upstream.

This is proper support for OpenBSD and it compiles: https://builds.sr.ht/~z3ntu/job/70849 . Should be tested with an actual application as well.

An interesting note from the OpenBSD ports:

> rename hid_init() to hidapi_hid_init() to avoid collision with usbhid

What they mean is that a library in OpenBSD https://man.openbsd.org/usbhid.3 already has the name `hid_init`. No idea how that can be solved cleanly.

### DIFF
--- a/.builds/openbsd.yml
+++ b/.builds/openbsd.yml
@@ -1,0 +1,20 @@
+image: openbsd/6.5
+packages:
+- autoconf-2.69p2
+- automake-1.16.1
+- libiconv
+- libtool
+- libusb1
+sources:
+- https://github.com/libusb/hidapi
+tasks:
+- setup: |
+    cd hidapi
+    export AUTOCONF_VERSION=2.69
+    export AUTOMAKE_VERSION=1.16
+    ./bootstrap
+    ./configure
+- build: |
+    cd hidapi
+    make
+    make DESTDIR=$PWD/root install
--- a/Makefile.am
+++ b/Makefile.am
@@ -31,6 +31,10 @@ if OS_KFREEBSD
 SUBDIRS += libusb
 endif
 
+if OS_OPENBSD
+SUBDIRS += libusb
+endif
+
 if OS_WINDOWS
 SUBDIRS += windows
 endif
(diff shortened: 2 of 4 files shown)


Youw:
We could have an inline wrapper `hid_init` over `hidapi_hid_init`
That would make it source compatible with all projects
To keep binary compatibility - maybe enable `hidapi_hid_init` and its wrapper only, when built on BSD

This is the best what I could think of right now

mcuee:
To be honest, using libusb backend for HID device under BSDs are not that helpful as it is not easy to use the libusb to replace the HID driver. It will be good to use native HID api.

mcuee:
Take note this is different from FreeBSD case. Back in 2012 I did tested quite a bit of FreeBSD USB stuff including HIDAPI as the USB stack allows generic driver along with HID driver for generic HID device.

mcuee:
Reference:
http://www.apcupsd.com/manual/manual.html#bsd-usb-configuration

```
Users of OpenBSD, NetBSD, and some versions of FreeBSD will need to rebuild the kernel in order to enable the ugen driver and disable the uhid driver. uhid is not sufficient for apcupsd at this time and we need to prevent it from grabbing the UPS device. You should make the following changes to your kernel config file:

FreeBSD (v5.4 and below, v6.0)
(you will not lose use of USB keyboard and mouse)
Disable: uhid
Enable: ugen
FreeBSD (v5.5, v6.1 and above)
(you will not lose use of USB keyboard and mouse)
Disable: (nothing)
Enable: ugen
This is the default configuration for a GENERIC kernel on many platforms so you most likely will not need to recompile.

NetBSD (v3.x and below)
(you will lose use of USB keyboard and mouse)
Disable: uhidev, ums, wsmouse, ukbd, wskbd, uhid
Enable: ugen
NetBSD (v4.0 and above)
You can use apcupsd on single USB port without disabling the USB keyboard and mouse on other ports, though all other devices will be disabled on the port you pick for your UPS.

OpenBSD
(you will lose use of USB keyboard and mouse):
Disable: uhidev, ums, wsmouse, ukbd, wskbd, uhid
Enable: ugen
```

Youw:
Maybe update `openbsd.yml` to use CMake? And discard all other changes?

PR #38, opened 2019-07-03
Add ar-lib & compile to .gitignore

z3ntu:
This should result in no untracked files being present in the working directory when `./bootstrap` was run


Youw:
what creates those files?

z3ntu:
`ar-lib` from automake and `compile` from autoconf apparently

PR #39, opened 2019-07-03
linux: Drop workaround for kernel < 2.6.34

z3ntu:
Also drop HIDIOC* definitions as they are present since kernel 2.6.39.

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -45,15 +45,6 @@
 
 #include "hidapi.h"
 
-/* Definitions from linux/hidraw.h. Since these are new, some distros
-   may not have header files which contain them. */
-#ifndef HIDIOCSFEATURE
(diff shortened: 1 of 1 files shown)


Youw:
where do we define minimum supported kernel?

Youw:
I still think it worth adding at least README notice. Would you?

z3ntu:
If you're willing to test hidapi on e.g. kernel 2.6.39 then sure; but other than downloading the oldest supported CentOS version I wouldn't even know how to get a VM running kernel 2.6.39

Youw:
Hm, good point - don't want to check/make any assures on support.
What about:
"Don't try to compile with Kernel older than 2.6.39 - it won't".

As usual, there will be that one person, who will try and open a bug.

z3ntu:
> As usual, there will be that one person, who will try and open a bug.

I'd say, if someone opens a bug report for that, we can add it to the readme :wink:

PR #44, opened 2019-07-13
WIP: Support for NetBSD

z3ntu:
Patch by nia from pkgsrc.se (NetBSD ports)

Should work fine, but I'll test it in a VM.

builds.sr.ht mostly has NetBSD CI support finished, but it could still take a while until a NetBSD image is usable; but afterwards a NetBSD CI can be added.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -31,6 +31,10 @@ if OS_KFREEBSD
 SUBDIRS += libusb
 endif
 
+if OS_NETBSD
+SUBDIRS += libusb
+endif
+
 if OS_WINDOWS
 SUBDIRS += windows
 endif
--- a/configure.ac
+++ b/configure.ac
@@ -107,6 +107,19 @@ case $host in
 	AC_CHECK_LIB([usb], [libusb_init], [LIBS_LIBUSB_PRIVATE="${LIBS_LIBUSB_PRIVATE} -lusb"], [hidapi_lib_error libusb])
 	echo libs_priv: $LIBS_LIBUSB_PRIVATE
 	;;
+*-netbsd*)
+	AC_MSG_RESULT([ (NetBSD back-end)])
+	AC_DEFINE(OS_NETBSD, 1, [NetBSD implementation])
+	AC_SUBST(OS_NETBSD)
+	backend="libusb"
+	os="netbsd"
+	threads="pthreads"
+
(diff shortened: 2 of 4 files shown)


z3ntu:
Compiles fine in my NetBSD VM, but (same as with OpenBSD), the USB stack doesn't seem to work correctly with QEMU ("hub1: device problem, disabling port 2")

mcuee:
To be honest, using libusb backend for HID device under NetBSD/OpenBSD are not that helpful as it is not easy to use the libusb to replace the HID driver. It will be good to use native HID api.

Youw:
I think this the statement can be applied to Linux platform too.
Maybe we could have two separate backends for BSD too: libusb + native.

mcuee:
For Linux it is at least possible to detach kernel hid driver.

mcuee:
Take note this is different from FreeBSD case. Back in 2012 I did tested quite a bit of FreeBSD USB stuff including HIDAPI as the USB stack allows generic driver along with HID driver for generic HID device.

I tested quite a bit of FreeBSD HPS USB stack back in 2007/2008 time. Then I tested for Alan Ott for FreeBSD HIDAPI in 2012. 

But I am no longer be able to test under macOS, NetBSD/OpenBSD/HaiKu now that my Mac Mini 2011 was dead.

mcuee:
Reference:
http://www.apcupsd.com/manual/manual.html#bsd-usb-configuration

```
Users of OpenBSD, NetBSD, and some versions of FreeBSD will need to rebuild the kernel in
order to enable the ugen driver and disable the uhid driver. uhid is not sufficient for 
apcupsd at this time and we need to prevent it from grabbing the UPS device. You should
make the following changes to your kernel config file:

FreeBSD (v5.4 and below, v6.0)
(you will not lose use of USB keyboard and mouse)
Disable: uhid
Enable: ugen
FreeBSD (v5.5, v6.1 and above)
(you will not lose use of USB keyboard and mouse)
Disable: (nothing)
Enable: ugen
This is the default configuration for a GENERIC kernel on many platforms so you most likely 
will not need to recompile.

NetBSD (v3.x and below)
(you will lose use of USB keyboard and mouse)
Disable: uhidev, ums, wsmouse, ukbd, wskbd, uhid
Enable: ugen
NetBSD (v4.0 and above)
You can use apcupsd on single USB port without disabling the USB keyboard and mouse on 
other ports, though all other devices will be disabled on the port you pick for your UPS.

OpenBSD
(you will lose use of USB keyboard and mouse):
Disable: uhidev, ums, wsmouse, ukbd, wskbd, uhid
Enable: ugen
```

PR #51, opened 2019-07-19
testgui: fix compile warnings on macOS

z3ntu:
The first commit should be backwards compatible (< 10.12), as I've added a define for the old type.

The second commit I'm not sure about, from what I could fine it was an API change in NSApplication in macOS 10.10, so it might not compile on versions before that?

### DIFF
--- a/testgui/mac_support_cocoa.m
+++ b/testgui/mac_support_cocoa.m
@@ -8,6 +8,15 @@
 #include <fx.h>
 #import <Cocoa/Cocoa.h>
 
+#ifndef MAC_OS_X_VERSION_10_12
+#define MAC_OS_X_VERSION_10_12 101200
+#endif
+
+// macOS 10.12 deprecated NSAnyEventMask in favor of NSEventMaskAny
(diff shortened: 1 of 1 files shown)


testgui/mac_support_cocoa.m
```diff
@@ -8,10 +8,15 @@
 #include <fx.h>
 #import <Cocoa/Cocoa.h>
 
+// macOS 10.12 deprecated NSAnyEventMask in favor of NSEventMaskAny
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 101200
```
Youw: lets more friendly name for this constant, e.g.: MAC_OS_X_VERSION_10_12

z3ntu: This constant isn't available under macOS 10.12 so we'd have to add
```c
#ifndef MAC_OS_X_VERSION_10_12
#define MAC_OS_X_VERSION_10_12 101200
#endif
```
Should I do that?

Youw: what about [this solution](https://stackoverflow.com/a/41497822/3697939) ?
your suggestion is fine by me as well
anything is better than magic numbers

PR #52, opened 2019-07-20
WIP: freebsd: use AM_ICONV macro

z3ntu:
Kind of a revert of #42 but I found a patch in the FreeBSD ports that uses this macro which should also simplify further BSD support

Unfortunately this adds a build-time dependency on gettext **on all platforms** (not just FreeBSD) because it contains that macro.

Does anyone know how auto* can be convinced it doesn't have to have this macro on other platforms?

### DIFF
--- a/config.rpath
+++ b/config.rpath
@@ -1,0 +1,684 @@
+#! /bin/sh
+# Output a system dependent set of variables, describing how to set the
+# run time search path of shared libraries in an executable.
+#
+#   Copyright 1996-2020 Free Software Foundation, Inc.
+#   Taken from GNU libtool, 2001
+#   Originally by Gordon Matzigkeit <gord@gnu.ai.mit.edu>, 1996
+#
+#   This file is free software; the Free Software Foundation gives
+#   unlimited permission to copy and/or distribute it, with or without
+#   modifications, as long as this notice is preserved.
+#
+# The first argument passed to this file is the canonical host specification,
+#    CPU_TYPE-MANUFACTURER-OPERATING_SYSTEM
+# or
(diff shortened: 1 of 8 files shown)


z3ntu:
If nobody can/wants to fix this, I'll close the PR because I'm not willing to add a build dependency plus an extra build step (`gettextize` is needed to be run apparently) just for nicer macros

Youw:
my suggestions (mutually exclusive):
- define our own macro, and manually add a compile time flag (e.g. `-DHIDAPI_ICONV_CONST`) for manually specified platform(s)
-  just suppress the warning with `#pragma` - functionally it never breaks anything

z3ntu:
I've never written autotools macros and don't plan to do so, so someone else would have to do it :)

Youw:
I have another suggestion:
What if we ship our own copy of `iconv.m4`? E.g. from [here](http://tug.ctan.org/tex-archive/macros/texinfo/texinfo/gnulib/m4/iconv.m4).
We already have a [folder](https://github.com/libusb/hidapi/tree/master/m4) for this purpose.

@z3ntu what do you think?

z3ntu:
@Youw done now I guess. I'm also not sure about using this approach though - if you look at the diff.

Youw:
yeah, I didn't realize there is so much dependencies :\

z3ntu:
With an older version of the macros at least `m4/host-cpu-c-abi.m4` isn't required but the rest still are.

mcuee:
This may not be relevant now that CMake is the way to go.

z3ntu:
Will close, as noone else seems to be interested since 2019. Maybe CMake has a better way.

PR #59, opened 2019-08-01
Added support for Get Input Report to linux, mac, windows and libusb.

DanielVanNoord:
This was tested for all operating systems and each worked identically.

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -334,6 +334,31 @@ extern "C" {
 		*/
 		int HID_API_EXPORT HID_API_CALL hid_get_feature_report(hid_device *dev, unsigned char *data, size_t length);
 
+				/** @brief Get a input report from a HID device.
+
+			Set the first byte of @p data[] to the Report ID of the
+			report to be read.  Make sure to allow space for this
+			extra byte in @p data[]. Upon return, the first byte will
+			still contain the Report ID, and the report data will
+			start in data[1].
+
+			@ingroup API
+			@param device A device handle returned from hid_open().
+			@param data A buffer to put the read data into, including
+				the Report ID. Set the first byte of @p data[] to the
+				Report ID of the report to be read, or set it to zero
+				if your device does not use numbered reports.
+			@param length The number of bytes to read, including an
+				extra byte for the report ID. The buffer can be longer
+				than the actual report.
+
+			@returns
+				This function returns the number of bytes read plus
+				one for the report ID (which is still in the first
+				byte), or -1 on error.
+		*/
+		int HID_API_EXPORT HID_API_CALL hid_get_input_report(hid_device *dev, unsigned char *data, size_t length);
+
 		/** @brief Close a HID device.
 
 			This function sets the return value of hid_error().
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -334,7 +334,7 @@ extern "C" {
 		*/
 		int HID_API_EXPORT HID_API_CALL hid_get_feature_report(hid_device *dev, unsigned char *data, size_t length);
 
-				/** @brief Get a input report from a HID device.
+		/** @brief Get a input report from a HID device.
 
 			Set the first byte of @p data[] to the Report ID of the
 			report to be read.  Make sure to allow space for this
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -337,7 +337,7 @@ extern "C" {
 		/** @brief Get a input report from a HID device.
 
 			Set the first byte of @p data[] to the Report ID of the
-			report to be read.  Make sure to allow space for this
+			report to be read. Make sure to allow space for this
 			extra byte in @p data[]. Upon return, the first byte will
 			still contain the Report ID, and the report data will
 			start in data[1].
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -791,7 +791,7 @@ int HID_API_EXPORT hid_get_feature_report(hid_device *dev, unsigned char *data, 
 	return res;
 }
 
-//Not implemented for Linux HidRaw yet
+// Not supported by Linux HidRaw yet
 int HID_API_EXPORT HID_API_CALL hid_get_input_report(hid_device *dev, unsigned char *data, size_t length)
 {
 	return -1;
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -794,7 +794,6 @@ int HID_API_EXPORT hid_get_feature_report(hid_device *dev, unsigned char *data, 
 int HID_API_EXPORT HID_API_CALL hid_get_input_report(hid_device *dev, unsigned char *data, size_t length)
 {
 	int res;
-
 	unsigned char bfr[64] = {0}; //Maximum Hid Report Size is 64 bytes, added one extra. See https://www.usb.org/sites/default/files/documents/hid1_11.pdf section 5.6
 
 	if(length > 64){
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -38,7 +38,6 @@
 #include <poll.h>
 
 /* Linux */
-#include <linux/hiddev.h>
 #include <linux/hidraw.h>
 #include <linux/version.h>
 #include <linux/input.h>
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -797,7 +797,7 @@ int HID_API_EXPORT HID_API_CALL hid_get_input_report(hid_device *dev, unsigned c
 
 	res = ioctl(dev->device_handle, HIDIOCGREPORT, data);
 	if (res < 0)
-		register_device_error_format(dev, "ioctl (GINPUT): %s", strerror(errno));
+		register_device_error_format(dev, "ioctl (GREPORT): %s", strerror(errno));
 
 	return res;
 }
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -38,6 +38,7 @@
 #include <poll.h>
 
 /* Linux */
+#include <linux/hiddev.h>
 #include <linux/hidraw.h>
 #include <linux/version.h>
 #include <linux/input.h>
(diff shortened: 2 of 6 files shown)


hidapi/hidapi.h
```diff
@@ -334,6 +334,31 @@ extern "C" {
 		*/
 		int HID_API_EXPORT HID_API_CALL hid_get_feature_report(hid_device *dev, unsigned char *data, size_t length);
 
+				/** @brief Get a input report from a HID device.
```
Youw:
```suggestion
		/** @brief Get a input report from a HID device.
```

linux/hid.c
```diff
@@ -790,6 +791,16 @@ int HID_API_EXPORT hid_get_feature_report(hid_device *dev, unsigned char *data,
 	return res;
 }
 
+int HID_API_EXPORT HID_API_CALL hid_get_input_report(hid_device *dev, unsigned char *data, size_t length)
+{
+	int res;
+
+	res = ioctl(dev->device_handle, HIDIOCGREPORT, data);
+	if (res < 0)
+		perror("ioctl (GINPUT)");
```
Youw: use `register_device_error_format` same way as in `hid_get_feature_report` does

DanielVanNoord:
Windows implementation better be same as for hid_send_feature_report, to return actual length
(tested in my projects)
I converted it over to use this method.
 
mac implementation sends report ID as part of data, as it was in hid_get_feature_report before #3
See my open issue /issues/60, this works (for me) the other way does not. It is possible that my device is out of spec. But it seems really odd that all the other platforms work and mac, when implemented to spec, does not.

hidraw implementation does not respect length parameter, which may lead to memory corruption
Not sure how to fix this one. The linux api that I am using does not appear to have an option for length.

Youw:
> Not sure how to fix this one. The linux api that I am using does not appear to have an option for length.

pass pointer to stack-allocated data that is guaranteed to be large enough for the API, and then copy up to `length` bytes of that data to `data`

Youw:
regarding mac and #60 - lets wait for comments of other maintainers/users

linux/hid.c
```diff
@@ -790,6 +791,16 @@ int HID_API_EXPORT hid_get_feature_report(hid_device *dev, unsigned char *data,
 	return res;
 }
 
+int HID_API_EXPORT HID_API_CALL hid_get_input_report(hid_device *dev, unsigned char *data, size_t length)
+{
+	int res;
+
+	res = ioctl(dev->device_handle, HIDIOCGREPORT, data);
+	if (res < 0)
+		register_device_error_format(dev, "ioctl (GINPUT): %s", strerror(errno));
```
z3ntu: In the log message it should be `GREPORT` as that's the ioctl constant being used

Youw:
something doesn't adds up in hidraw implementation: if I understand how `ioctl` works, as per `HIDIOCGREPORT` definition, its argument can only be `struct hiddev_report_info`, not some user data

I'll spend more time investigating it a bit later

DanielVanNoord:
It is possible, I wrote (and studied) the windows, mac and libusb implementations. Another developer that I worked with wrote the Linux one (I saw your issue at signal11/hidapi and am trying to up-stream now). I simply verified that it worked with both test boards that I have.

windows/hid.c
```diff
@@ -818,6 +822,55 @@ int HID_API_EXPORT HID_API_CALL hid_get_feature_report(hid_device *dev, unsigned
 #endif
 }
 
+
+int HID_API_EXPORT HID_API_CALL hid_get_input_report(hid_device *dev, unsigned char *data, size_t length)
+{
+#if 0
```
Qbicz: Does it make sense to keep this disabled code? I know it is the same way in `hid_get_feature_report()`, but what purpose does this code section serve?

windows/hid.c
```diff
@@ -818,6 +822,55 @@ int HID_API_EXPORT HID_API_CALL hid_get_feature_report(hid_device *dev, unsigned
...
+	BOOL res;
+	res = HidD_GetInputReport(dev->device_handle, data, length);
+	if (!res) {
+		register_error(dev, "HidD_GetInputReport");
+		return -1;
+	}
+	return length
```
Qbicz:
```suggestion
	return length;
```

DavidCNelson:
> regarding mac and #60 - lets wait for comments of other maintainers/users
I am working on a new device with HID and also on the host side. This pull request is rather timely for me as the day it was created was the same day I was getting ready to implement the same functionality.
The issue seems to be the question if there is a report ID at the beginning of the data transfer. The [HID Spec v1.11](https://www.usb.org/sites/default/files/documents/hid1_11.pdf) in section 5.6 describes that if the HID device defines report IDs in the descriptor, then the report ID must be sent as the first byte of all data transfers. If the device does not define report IDs(appears as `report ID == 0`), then the non-existent report id is not included in the data transfer. 
Consider the code in `set_report` in mac/hid.c at lines 774-785. If the report id is 0, it does not send the report ID as part of the data. If the report ID is non-zero, then the whole buffer is sent. A similar thing needs to happen for `hid_get_feature_report` and `hid_get_input_report`. If the report ID is 0, then the received data will not have a report ID, so the data should be read into the buffer starting with the second byte of the buffer. If the report ID is non-zero, the data received should start with the report ID, so you can fill the buffer starting with the first byte.
Depending on the device one is using, #3 fixes an issue or, as #60 points out, it is a regression.

Youw:
HID Spec v1.11 in section 5.6 also specifies how to know if a device is using numbered reports: 

> A Report ID item tag assigns a 1-byte identification prefix to each report transfer. _If no Report ID item tags are present in the **Report descriptor**_, it can be assumed that only one Input, Output, and Feature report structure exists and together they represent all of the device’s data.

@DanielVanNoord, @DavidCNelson can you confirm, that your devices defines corresponding Report IDs tags in the Report descriptor?

DavidCNelson:
> HID Spec v1.11 in section 5.6 also specifies how to know if a device is using numbered reports:
> 
> > A Report ID item tag assigns a 1-byte identification prefix to each report transfer. _If no Report ID item tags are present in the **Report descriptor**_, it can be assumed that only one Input, Output, and Feature report structure exists and together they represent all of the device’s data.
> 
> @DanielVanNoord, @DavidCNelson can you confirm, that your devices defines corresponding Report IDs tags in the Report descriptor?

Yes, the device I'm working with does define report IDs in the descriptor, and so we do use report IDs for all transfers.

windows/hid.c
DanielVanNoord: hid_get_feature_report had it so I included it as well. The only benefit I can think of is that it shows that the other way is known and wrong, even though it is much shorter.

DanielVanNoord:
Same here, the device I am working on uses report IDs in the descriptors for all transfers.

linux/hid.c
```diff
@@ -790,6 +791,25 @@ int HID_API_EXPORT hid_get_feature_report(hid_device *dev, unsigned char *data,
...
+	int res;
+	unsigned char bfr[64] = {0}; //Maximum Hid Report Size is 64 bytes, added one extra. See https://www.usb.org/sites/default/files/documents/hid1_11.pdf section 5.6
+
+	if(length > 64){
+		register_device_error_format(dev, "ioctl (GREPORT): %s", strerror(EINVAL)); //22      /* Invalid argument */
+	}
+
+	memcpy(bfr, data, length); // copy the data onto the buffer	
+
+	res = ioctl(dev->device_handle, HIDIOCGREPORT, bfr);
```
micolous: `HIDIOCGREPORT` [expects a `hiddev_report_info`](https://elixir.bootlin.com/linux/v4.4/source/include/uapi/linux/hiddev.h#L160), which is [declared here](https://elixir.bootlin.com/linux/v4.4/source/include/uapi/linux/hiddev.h#L65).

This differs from `HIDIOCGFEATURE` which takes in the [first byte as a report ID](https://elixir.bootlin.com/linux/v4.4/source/include/uapi/linux/hidraw.h#L40).

I'm also not sure that this is a valid `ioctl` for `hidraw` -- this looks like  it is for `hiddev`!

Youw: also, I don't see support of Input Reports by hidraw at all

Youw: Nope, [it doesn't](https://elixir.bootlin.com/linux/v5.3-rc3/source/drivers/hid/hidraw.c#L424)

linux/hid.c
```diff
@@ -790,6 +791,11 @@ int HID_API_EXPORT hid_get_feature_report(hid_device *dev, unsigned char *data,
 	return res;
 }
 
+//Not implemented for Linux HidRaw yet
```
Youw:
```suggestion
// Not supported by Linux HidRaw yet
```

Youw:
@DanielVanNoord please resolve merge conflict, and:
- update mac implementation as per #70 
- update windows\ddk_build\hidapi.def

DanielVanNoord:
@Youw Should be taken care of now

linux/hid.c
```diff
@@ -38,6 +38,7 @@
 #include <poll.h>
 
 /* Linux */
+#include <linux/hiddev.h>
```
Youw: remove

windows/hid.c
```diff
@@ -109,6 +110,7 @@ extern "C" {
 	typedef BOOLEAN (__stdcall *HidD_GetProductString_)(HANDLE handle, PVOID buffer, ULONG buffer_len);
 	typedef BOOLEAN (__stdcall *HidD_SetFeature_)(HANDLE handle, PVOID data, ULONG length);
 	typedef BOOLEAN (__stdcall *HidD_GetFeature_)(HANDLE handle, PVOID data, ULONG length);
+	typedef BOOLEAN(__stdcall *HidD_GetInputReport_)(HANDLE handle, PVOID data, ULONG length);
```
Youw:
```suggestion
	typedef BOOLEAN (__stdcall *HidD_GetInputReport_)(HANDLE handle, PVOID data, ULONG length);
```

PR #65, opened 2019-08-21
support multipe usage/usage-pairs on MacOS

fengj2006:
This change supports one device has multiple usage/usage-page pairs on MacOS, which is already supported on Windows.

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -46,7 +46,7 @@ extern "C" {
 		typedef struct hid_device_ hid_device; /**< opaque hidapi structure */
 
 		/** hidapi info structure */
-		typedef struct hid_device_info {
+		struct hid_device_info {
 			/** Platform-specific device path */
 			char *path;
 			/** Device Vendor ID */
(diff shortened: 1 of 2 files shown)


hidapi/hidapi.h
```diff
@@ -80,7 +80,7 @@ extern "C" {
 
 			/** Pointer to the next device */
 			struct hid_device_info *next;
-		};
+		} HID_DEVICE_INFO;
```
Youw: this change is not related to what PR is really about - support usage pages on macOS
it also changes public API, and there is no real need for it, please undo

fengj2006: Fixed

Youw:
Fixes #116

mac/hid.c
```diff
@@ -392,6 +406,117 @@ static void process_pending_events(void) {
...
+			}
+			else {
+				cur->next = next;
+			}
+			cur = next;
+		}
+	}
+
+	if (root == NULL) {
+		/* error when generating or parsing usage pairs */
```
```suggestion
		/* error when generating or parsing usage pairs, or device/driver doesn't report usage pairs (correctly) */
```

Youw:
I don't think there is a way to submit your changes to this particular PR since you'd need to commit to author's PR, and you don't have write access to his copy of the repo.

But you can:
- clone original repo (origin);
- add git remote to fengj2006;
- checkout this PR's branch (this andbthevprevious just so you have )

PR #69, opened 2019-09-01
Prevent memory leak in hid_open_path() if hid_init() fails

Qix-:
On MacOS, a `calloc()` is made but isn't freed in the event the HID manager doesn't initialize properly.

### DIFF
--- a/mac/hid.c
+++ b/mac/hid.c
@@ -696,11 +696,11 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
 	io_registry_entry_t entry = MACH_PORT_NULL;
 	IOReturn ret = kIOReturnInvalid;
 
-	dev = new_hid_device();
-
 	/* Set up the HID Manager if it hasn't been done */
 	if (hid_init() < 0)
-		return NULL;
+		goto return_error;
+
+	dev = new_hid_device();
 
 	/* Get the IORegistry entry for the given path */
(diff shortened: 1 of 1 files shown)


mac/hid.c
```diff
@@ -696,12 +696,12 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
 	io_registry_entry_t entry = MACH_PORT_NULL;
 	IOReturn ret = kIOReturnInvalid;
 
-	dev = new_hid_device();
-
 	/* Set up the HID Manager if it hasn't been done */
 	if (hid_init() < 0)
 		return NULL;
```
Qbicz: Wouldn't it be better to use `goto return_error` instead of `return NULL`? It will solve the problem as well, but also will prevent any similar problems if anyone put any allocation before `if (hid_init() < 0)` line in the future.

Youw: PR author is not answering, but I think the patch is more than acceptable as is - it does better then before in any case

Qix-: @Qbicz I'm not well versed enough in the library to have an opinion, hence why I didn't answer. I figured someone else might have an opinion on it.

z3ntu: I agree with @Qbicz

```suggestion
		goto return_error;
```

Qix-:
Done!

Issue #60, opened 2019-08-01
Possible regression in mac get_feature_report


DanielVanNoord:
Prior to /pull/3 being merged get_feature_report was working identically for me on all platforms. Now the mac driver is failing as the returned buffer is different from the other platforms. My device does use report ids.
Is it possible that the behavior of OSX was changed from when this change was originally requested back in 2015?
Can anyone reproduce this issue?

Youw:
Can you describe how different is it?
And is it different from previous release shipped with homebrew (0.8.0)?

DanielVanNoord:
The message that it is sending on the mac does not appear to include the report ID, so my device reports an error. If I write the rid in both the 1st and 2nd byte it sends the rid as part of the report. However I get a response on the mac of rid, rid, bytes as the mac library is writing the response out starting at the 2nd byte rather then the 1st.  Prior to this change it worked correctly.

Other platforms include the rid in the (outbound) report and responded with the expected rid, bytes.

I wasn't using homebrew, I am building this myself. I was using Signal11/HidAPI (where it worked) but am trying to switch to yours as it is maintained. Once I reverted this pull everything worked as it did before.

Youw:
Documentation of [IOHIDDeviceGetReport](https://developer.apple.com/documentation/iokit/1588659-iohiddevicegetreport?language=objc) is very explicit, that `report` should contain only inbound report data, and that is what #3 about it.

Can you confirm, that rid is other than 0x0 ?

DanielVanNoord:
I agree that it is very explicit, that's why I asked if anyone else experience this issue. In this case the report id values that are not working are 0x14, 0x15 and 0x16. 

It is possible that there is something out of spec with the device I have been using for testing. However it works just fine on all other platforms and on mac without this change.

micolous:
What's happening at a protocol level ([see section 8.2 of HID 1.11 spec](https://www.usb.org/sites/default/files/documents/hid1_11.pdf)), is that the Report ID will be removed from a response when a device doesn't use Report IDs.

This leaves some gaps – it's now up to the kernel to figure out if there's a report ID in there or not.

HID has other platform differences:

* [Linux's `HIDIOCGFEATURE` ioctl documentation says](https://www.kernel.org/doc/Documentation/hid/hidraw.txt) (emphasis mine):

  > HIDIOCGFEATURE(len): Get a Feature Report
  > This ioctl will request a feature report from the device using the control
[...]

PR #70, opened 2019-09-02
macOS: send report id conditionally for hid_get_feature_report

Youw:
same as in hid_send_feature_report, and Windows/Linux

### DIFF
--- a/mac/hid.c
+++ b/mac/hid.c
@@ -763,38 +763,61 @@ return_error:
 
 static int set_report(hid_device *dev, IOHIDReportType type, const unsigned char *data, size_t length)
 {
-	const unsigned char *data_to_send;
-	size_t length_to_send;
+	const unsigned char *data_to_send = data;
+	CFIndex length_to_send = length;
 	IOReturn res;
+	const unsigned char report_id = data[0];
 
-	/* Return if the device has been disconnected. */
-	if (dev->disconnected)
-		return -1;
-
-	if (data[0] == 0x0) {
+	if (report_id == 0x0) {
 		/* Not using numbered Reports.
 		   Don't send the report number. */
 		data_to_send = data+1;
 		length_to_send = length-1;
 	}
-	else {
-		/* Using numbered Reports.
-		   Send the Report Number */
-		data_to_send = data;
-		length_to_send = length;
+
+	/* Avoid crash if the device has been unplugged. */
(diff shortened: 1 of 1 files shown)


mac/hid.c
```diff
@@ -763,38 +763,61 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
...
 	}
-	else {
-		/* Using numbered Reports.
-		   Send the Report Number */
-		data_to_send = data;
-		length_to_send = length;
-	}
 
+	/* Avoid crash if the device has been unplugged. */
 	if (!dev->disconnected) {
```
vitalys: Probably it has to be `dev->disconnected`

Youw:
This will likely reference https://github.com/libusb/hidapi/issues/60.

@DanielVanNoord, @micolous, please try if this works with your device(s).
It does with mine.

mac/hid.c
Youw: thanks)

Youw:
@DanielVanNoord, @micolous ping

hidapi team, your opinion/feedback is needed
with all my local tests, this implementation looks best and consistent with other systems

@z3ntu, you have originally propagated #3, so you feedback is appreciated too

z3ntu:
Apparently the devices I use hidapi with, don't use numbered reports. I'll test this PR tomorrow if it works with my application!

Youw:
Thanks.
The device I use, does use numbered reports, and #3 did broke my app (as I mentioned earlier, I didn't notice that, since I have my own fork with my own patches).

If this works for you, I think we will accept it this was.

z3ntu:
Works fine for my use case :+1:

Youw:
@Qbicz, if you don't have comments regarding the implementation, I'd like to merge this ASAP

DanielVanNoord:
@Youw I'll grab a Mac and test it with my devices that use report ids today. Will let you know the results as soon as I can.

Qbicz:
@youw it looks good to me. I don't have a Mac setup so I didn't notice the problem before.

Youw:
meanwhile found bug in my own implementation
nothing conceptually changes though

DanielVanNoord:
Ran some quick tests on a Mac with a device that uses feature report ids and everything seems to be working.

Youw:
great, we have a winner here

PR #102, opened 2019-09-27
Updated spec for v0.9.1 (merge it before releasing 0.9.1)

vitalys:


Youw:
we should also update the library version first
I might do it in one shot later, but thanks for reminding

Issue #105, opened 2019-10-17
04f2:1670 Chicony Electronics Co., Ltd keyboard is broken by hidapi example (hidapi-libusb)


amosbird:
Hello. I'm trying to test hidapi and possibly use it to set fn_lock of my keyboard. I grabbed the example code and tried to see if it runs.  https://github.com/libusb/hidapi#what-does-the-api-look-like 

Then suddenly my keyboard stops responding, `lsusb` doesn't list it anymore. The id was `0x17ef, 0x60b5` now becomes `0x04f2, 0x1670`  (I tested it while unplug and plug). 

Is there any way of restoring my keyboard? Thank you!

amosbird:
The device is very much broken

```
Bus 003 Device 012: ID 04f2:1670 Chicony Electronics Co., Ltd
Couldn't open device, some information will be missing
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0
  bDeviceSubClass         0
  bDeviceProtocol         0
  bMaxPacketSize0         8
  idVendor           0x04f2 Chicony Electronics Co., Ltd
  idProduct          0x1670
  bcdDevice            4.00
  iManufacturer           0
  iProduct                0
  iSerial                 0
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0029
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0
    bmAttributes         0xa0
      (Bus Powered)
      Remote Wakeup
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      0
      bInterfaceProtocol      0
      iInterface              0
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 Report
          wDescriptorLength      31
         Report Descriptors:
           ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
[...]

PR #106, opened 2019-10-18
README: Add warning about writing data to HID products

Qbicz:
In response to https://github.com/libusb/hidapi/issues/105

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,9 @@ int main(int argc, char* argv[])
 	for (i = 0; i < 4; i++)
 		printf("buf[%d]: %d\n", i, buf[i]);
 
+	// Close the device
+	hid_close(handle);
+
 	// Finalize the hidapi library
 	res = hid_exit();
 
--- a/README.md
+++ b/README.md
@@ -68,11 +68,14 @@ which HIDAPI supports.  Since it relies on a 3rd party library, building it
 is optional but recommended because it is so useful when debugging hardware.
 
 ## What Does the API Look Like?
-The API provides the the most commonly used HID functions including sending
-and receiving of input, output, and feature reports.  The sample program,
+The API provides the most commonly used HID functions including sending
+and receiving of input, output, and feature reports. The sample program,
 which communicates with a heavily hacked up version of the Microchip USB
 Generic HID sample looks like this (with error checking removed for
 simplicity):
+
+**Warning: Only run the code you understand, and only when it conforms to the
+device spec. Writing data at random to your HID devices can break them.**
 
 ```c
 #ifdef WIN32
@@ -138,6 +141,10 @@ int main(int argc, char* argv[])
 	return 0;
 }
 ```
+
+You can also use [hidtest/hidtest.c](https://github.com/libusb/hidapi/blob/master/hidtest/hidtest.c)
+as a starting point for your applications.
+
 
 ## License
 HIDAPI may be used by one of three licenses as outlined in [LICENSE.txt](LICENSE.txt).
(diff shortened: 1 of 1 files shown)


Qbicz:
@Youw @z3ntu @amosbird can you review?

README.md
```diff
@@ -139,6 +142,10 @@ int main(int argc, char* argv[])
 }
 ```
 
+You can also use [hidtest/hidtest.c](https://github.com/libusb/hidapi/blob/master/hidtest/hidtest.c)
+as a starting point for your programmes.
```
z3ntu: `programs` instead of `programmes`? I know it's a bit British vs American English but it's more fitting in this context.

README.md
```diff
@@ -68,12 +68,15 @@ which HIDAPI supports.  Since it relies on a 3rd party library, building it
...
-The API provides the the most commonly used HID functions including sending
-and receiving of input, output, and feature reports.  The sample program,
+The API provides the most commonly used HID functions including sending
+and receiving of input, output, and feature reports. The sample program,
 which communicates with a heavily hacked up version of the Microchip USB
 Generic HID sample looks like this (with error checking removed for
 simplicity):
 
+`Warning: Only run the code you understand, only against devices which firmware 
+you control. Writing data at random to your HID devices can break them.`
```
z3ntu: I would use *s instead of `s, so to make it bold because it looks a bit weird in this preformatted style.

README.md
Qbicz: `applications` ;)

README.md
Qbicz: done

Qbicz:
> This statement is a bit too strong - you may not nesessary control the FW, but safely write, e.g. if you have official spec, etc.

@Youw done.

> I was thinking, maybe add explicit check for dev/ven before hid_write in the example

I think that would be too much, since to write to the device in the sample you have to enumerate devices, check vid/pid and hardcode it to the sample. Or do you mean interactive check like with `gets()`?

Youw:
I was thinking something like:
```cpp
#define SAMPLE_VEN_ID ...
#define SAMPLE_DEV_ID ...
unsigned short ven_id = SAMPLE_VEN_ID;
unsigned short dev_id = SAMPLE_DEV_ID;

int main(/*...*/) {
  if (argc /* ... */ ) {
    // ...
    ven_id = // parse arguments, etc.
    // ...
  }

  handle = hid_open(ven_id, dev_id, NULL);
  // ...

  // a WARNING comment here, what we're know what we're doing, but only with a specific device
  if (ven_id == SAMPLE_VEN_ID && dev_id == SAMPLE_DEV_ID) {
    // ...
    res = hid_write(handle, buf, 65);
    // ...
  }
  // hid_close
  // hid_exit
}
```

So we can include #14 and a little more, like explicitly encourage to `hid_close` which is not in an example.

---
Maybe this is a little too large change, but anyway nice to have.

Qbicz:
@Youw thanks for your comments, I added `hid_close()` call to the sample as you suggested, so this is now explicitly visible.

README.md
```diff
@@ -132,13 +135,20 @@ int main(int argc, char* argv[])
...
+	hid_close(handle);
+
 	// Finalize the hidapi library
 	res = hid_exit();
 
 	return 0;
 }
 ```
 
+You can also use [hidtest/hidtest.c](https://github.com/libusb/hidapi/blob/master/hidtest/hidtest.c)
```
Youw:
```suggestion
You can also use [hidtest/test.c](hidtest/test.c)
```

PR #109, opened 2019-10-24
Fix for wrong text on Windows on unusual cases

JosepMaJAZ:
Fix for unusual cases where the hardware does not return the name and the string is not initialized at all.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -437,6 +437,7 @@ struct hid_device_info HID_API_EXPORT * HID_API_CALL hid_enumerate(unsigned shor
 				cur_dev->path = NULL;
 
 			/* Serial Number */
+			wstr[0]= 0x0000;
 			res = HidD_GetSerialNumberString(write_handle, wstr, sizeof(wstr));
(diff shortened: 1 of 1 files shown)


Youw:
Hi @JosepMaJAZ, thank you for your contribution!

Please help me understand the issue:

_Sometime Windows HID driver may not initialize the string, even if return `success`, and this patch simply to avoid returning gibberish as a string value._

Is last statement correct?

JosepMaJAZ:
Hi. This library is used on the open source project Mixxx and when we were still using the signal11 version of it, some users, myself included, saw some devices with random characters. See here https://mixxx.org/forums/viewtopic.php?f=10&t=11748&p=43098&hilit=Hidapi#p43098

When i debugged it back then, I saw that the calls to get the names returned with no error, but the string was not informed in any way, and since the initial array is not initialized, it contains random data.

This is the reason why i created this patch and tried that it was merged to signal11 version, but I've been informed recently that libusb version is the one tha is being maintained.

Youw:
OK, thanks.

> libusb version is the one tha is being maintained

yes, that's true.

Issue #108, opened 2019-10-22
hidtest-libusb and hidtest-hidraw do not compile


piciu67:
Original signal11 hidtest.cpp was renamed to hidtest.c and example makefiles "Makefile-manual" in hidapi/libusb and hidapi/linux stopped working.

Youw:
Thanks for reporting this!

Would you be kind to prepare a PR and provide a fix?

It looks like it is a matter of removing `CPPOBJS` variable in `Makefile.linux`.

piciu67:
Simplest way fix this problem is just to rename hidtest.c back to hidtest.cpp.

Youw:
Maybe it will fix this particular issue, but it creates a bigger one: _requirement to have a C++ compiler_, which is a big deal sometime, and not really required to actually compile and run the example (see #22).

z3ntu:
A start for the patch would be
```diff
diff --git a/linux/Makefile-manual b/linux/Makefile-manual
index 04ce1de..c8aab27 100644
--- a/linux/Makefile-manual
+++ b/linux/Makefile-manual
@@ -19,16 +19,15 @@ CXXFLAGS ?= -Wall -g -fpic
 LDFLAGS  ?= -Wall -g
 
 
-COBJS     = hid.o
-CPPOBJS   = ../hidtest/hidtest.o
-OBJS      = $(COBJS) $(CPPOBJS)
+COBJS     = hid.o ../hidtest/hidtest.o
+OBJS      = $(COBJS)
 LIBS_UDEV = `pkg-config libudev --libs` -lrt
 LIBS      = $(LIBS_UDEV)
 INCLUDES ?= -I../hidapi `pkg-config libusb-1.0 --cflags`
 
 
 # Console Test Program
-hidtest-hidraw: $(COBJS) $(CPPOBJS)
+hidtest-hidraw: $(COBJS)
 	$(CXX) $(LDFLAGS) $^ $(LIBS_UDEV) -o $@
 
 # Shared Libs
@@ -39,9 +38,6 @@ libhidapi-hidraw.so: $(COBJS)
 $(COBJS): %.o: %.c
 	$(CC) $(CFLAGS) -c $(INCLUDES) $< -o $@
 
-$(CPPOBJS): %.o: %.cpp
-	$(CXX) $(CXXFLAGS) -c $(INCLUDES) $< -o $@
-
 
 clean:
[...]

PR #115, opened 2019-11-03
Fix hidtest builds

### LINKED ISSUE #114: hidtest build fails on macOS: No rule to make target `hidtest.cpp'
On macOS (Mojave 10.4.6) I have the error:
```
± make
/Applications/Xcode.app/Contents/Developer/usr/bin/make  all-recursive
Making all in mac
make[2]: Nothing to be done for `all'.
Making all in hidtest
make[2]: *** No rule to make target `hidtest.cpp', needed by `hidtest.o'.  Stop.
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

This is a regression caused by d98c408ed3939647cd2eef255fbb24b463b24622

If I rename the source code to `hidtestx.c` and adapt the `Makefile.am` I fix the issue.

I don't know why it fails. Maybe make is confused because the directory, the source code and the target have the same name?

```
± make -v
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

LudovicRousseau:
Fixes #114 and #108


linux/Makefile-manual
```diff
@@ -44,6 +44,6 @@ $(CPPOBJS): %.o: %.cpp
 
 
 clean:
-	rm -f $(OBJS) hidtest-hidraw libhidapi-hidraw.so ../hidtest/hidtest.o
+	rm -f $(OBJS) hidtest-hidraw libhidapi-hidraw.so $(CPPOBJS)
```
z3ntu: CPPOBJS is empty, I think COBJS should be here

LudovicRousseau: Exact.
New version pushed.

Issue #119, opened 2019-11-17
macOS: location of IOKit.framework


0xDP:
Hi there,

according to

https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/OSX_Technology_Overview/SystemFrameworks/SystemFrameworks.html

IOKit.framework is located at "/System/Library/Frameworks/", but mac/hid.c:hidapi_IOHIDDeviceGetService() tries to find it at "/System/Library/".

```
--- a/hidapi/mac/hid.c
+++ b/hidapi/mac/hid.c
@@ -305,7 +305,7 @@ static io_service_t hidapi_IOHIDDeviceGe
 	 * and the fallback method will be used.
 	 */
 	if (iokit_framework == NULL) {
-		iokit_framework = dlopen("/System/Library/IOKit.framework/IOKit", RTLD_LAZY);
+		iokit_framework = dlopen("/System/Library/Frameworks/IOKit.framework/IOKit", RTLD_LAZY);
 
 		if (iokit_framework != NULL)
 			dynamic_IOHIDDeviceGetService = (dynamic_IOHIDDeviceGetService_t) dlsym(iokit_framework, "IOHIDDeviceGetService");
```

Regards,
Dennis

z3ntu:
Please copy https://github.com/signal11/hidapi/pull/426 to this repo.

PR #120, opened 2019-11-17
macOS: IOKit framework location corrected.

0xDP:
The referenced location of IOKit framework is wrong.

See https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/OSX_Technology_Overview/SystemFrameworks/SystemFrameworks.html for details.

### DIFF
--- a/mac/hid.c
+++ b/mac/hid.c
@@ -305,7 +305,7 @@ static io_service_t hidapi_IOHIDDeviceGetService(IOHIDDeviceRef device)
 	 * and the fallback method will be used.
 	 */
 	if (iokit_framework == NULL) {
-		iokit_framework = dlopen("/System/Library/IOKit.framework/IOKit", RTLD_LAZY);
+		iokit_framework = dlopen("/System/Library/Frameworks/IOKit.framework/IOKit", RTLD_LAZY);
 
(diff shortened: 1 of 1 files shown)


Youw:
thanks for your PR

seems reasonable, but we have to take into account the fact that this has been working for years, and I don't recall any issues related to it

best case scenario, if we could test it on all versions of macOS starting with 10.7 and check how exactly does this work (before/after the change), not to accidentally cause some regressions
Any volunteers?

I can check only latest macOS 10.15 (Catalina)

---
Secondly, I'm not sure we want to support macOS 10.6 and older (for the same reason we dropped support of Linux 2.x(something) earlier, and simpler would be just ling against `IOKit.framework` statically

Youw:
Fixes #119

conurus:
This fix actually fixed my problem in Mojave and Catalina. My app worked until put in a sandbox and signed with Hardened Runtime with the com.apple.security.device.usb entitlement. hid_enumerate put empty strings in the path field and I was not able to open anything. This was fixed when the path to dlopen was corrected. Just here to say thank you for this fix and to confirm that this fix has real world significance even though things seem to have been working for many many years.

PR #123, opened 2019-11-19
libusb: allow hidapi to be configured without using iconv

wet34t:


Youw:
additionally would be great to have auto-tools option to enable this

wet34t:
Basically, the idea is to allow users to configure hidapi without using iconv, as icvon may have dependencies that may not be supported

z3ntu:
What's the use case for this patch? As far as I know, iconv is implemented directly in glibc and for non-glibc libcs, there's GNU libiconv which also apparently doesn't have any dependencies other than the C library.

wet34t:
iconv depends on /usr/lib[64]/gconv, which usually is part of glibc, but may not always be the case.

Youw:
I see no issues accepting this as is

z3ntu:
I still haven't seen an exact reason like "On X distro iconv isn't available and as such hidapi must be built without it" or something like that.

To my knowledge, glibc has iconv built in, musl as well. For other libc's you can install GNU libiconv and use that instead of the built-in-to-the-C-library libiconv.

Youw:
@wet34t, did you hit one of the cases described by @z3ntu?

wet34t:
I'm experiencing similar issues as described here:
https://stackoverflow.com/questions/9560671/weird-iconv-open-behavior-in-linux

Youw:
I see no issues having this as is for community convenience

Issue #24, opened 2019-06-17
How to build for Android


z3ntu:
```
$ ~/Android/Sdk/ndk-bundle/ndk-build
Android NDK: APP_PLATFORM not set. Defaulting to minimum supported version android-16.    
/mnt/hdd/Android/Sdk/ndk-bundle/build/core/build-binary.mk:672: Android NDK: Module hidapi depends on undefined modules: usb1.0    
/mnt/hdd/Android/Sdk/ndk-bundle/build/core/build-binary.mk:685: *** Android NDK: Aborting (set APP_ALLOW_MISSING_DEPS=true to allow missing dependencies)    .  Stop.
```
```
$ env APP_ALLOW_MISSING_DEPS=true ~/Android/Sdk/ndk-bundle/ndk-build
Android NDK: APP_PLATFORM not set. Defaulting to minimum supported version android-16.    
/mnt/hdd/Android/Sdk/ndk-bundle/build/core/build-binary.mk:672: Android NDK: Module hidapi depends on undefined modules: usb1.0    
/mnt/hdd/Android/Sdk/ndk-bundle/build/core/build-binary.mk:672: Android NDK: Module hidapi depends on undefined modules: usb1.0    
/mnt/hdd/Android/Sdk/ndk-bundle/build/core/build-binary.mk:672: Android NDK: Module hidapi depends on undefined modules: usb1.0    
/mnt/hdd/Android/Sdk/ndk-bundle/build/core/build-binary.mk:672: Android NDK: Module hidapi depends on undefined modules: usb1.0    
[arm64-v8a] Compile        : hidapi <= hid.c
/home/luca/Dev/hidapi/android/jni/../../libusb/hid.c:47:10: fatal error: 'libusb.h' file not found
#include <libusb.h>
         ^~~~~~~~~~
1 error generated.
make: *** [/home/luca/Dev/hidapi/android/obj/local/arm64-v8a/objs/hidapi/__/__/libusb/hid.o] Error 1
```

With the following modifications to the Android.mk file, the build succeeds (using `~/Android/Sdk/ndk-bundle/ndk-build`):
```diff
diff --git a/android/jni/Android.mk b/android/jni/Android.mk
index 527b43f..32e3442 100644
--- a/android/jni/Android.mk
+++ b/android/jni/Android.mk
@@ -1,8 +1,16 @@
-LOCAL_PATH:= $(call my-dir)
+MY_LOCAL_PATH:= $(call my-dir)
+
[...]

PR #130, opened 2019-12-12
Fix build when targeting Android 7 or later

fornwall:
Support for pthread_barrier was introduced in Android 7, so trying to build a custom implementation of that clashes with the system headers.

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -51,7 +51,7 @@
 
 #include "hidapi.h"
 
-#ifdef __ANDROID__
+#if defined(__ANDROID__) && __ANDROID_API__ < __ANDROID_API_N__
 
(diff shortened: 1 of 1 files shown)


z3ntu:
Can confirm that compilation works fine with this patch for sdk 29 (`~/Android/Sdk/ndk/20.0.5594570/ndk-build APP_PLATFORM=android-29` and #24) and it fails without this patch.

@fornwall Do you have a better solution for #24 or how are you compiling hidapi for Android?

Youw:
Thanks @z3ntu for checking.

fornwall:
@z3ntu Thanks! Regarding #24, I'm not using ndk-build when building here, instead using the NDK as a traditional cross-compilation toolchain to build a package for [Termux](https://termux.com/).

PR #131, opened 2019-12-14
Fix manual Makefiles and add CI for them

### LINKED ISSUE #108: hidtest-libusb and hidtest-hidraw do not compile

z3ntu:
Fixes #108

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -2,6 +2,7 @@ image: freebsd/latest
 packages:
 - autoconf
 - automake
+- gmake
 - libiconv
 - libtool
 - pkgconf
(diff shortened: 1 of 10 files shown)


.builds/freebsd.yml
```diff
@@ -16,3 +17,7 @@ tasks:
     cd hidapi
     make
     make DESTDIR=$PWD/root install
+    make clean
+- build-manual: |
+    cd hidapi/libusb
+    gmake -f Makefile-manual
```
z3ntu: Unfortunately the freebsd manual makefiles aren't posix compliant, so gmake is needed there. I researched a bit but I didn't find a good reference how e.g. `ifeq` statements should look like in POSIX make, so let's just use gmake

Youw:
LGTM

z3ntu:
bump?

Youw:
I don't this system(s) to try, but the changes looks OK.
If there are any issues, we could fix them later.

Issue #142, opened 2020-02-01
Linux: Hanging inside hid_close()


vozhyk-com:
For catching this hanging need just open device by hid_open() or hid_open_path() and close it by hid_close(), see simple example:

#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include "hidapi.h"

int main(int argc, char* argv[])
{
     hid_device *handle;
     struct hid_device_info *devs, *cur_dev;
   
     if (hid_init())
          return -1; 
     int i;
     for (i =0 ; i < 10000; i++) { 
          handle = hid_open(0x22c9, 0x0901, NULL);
          if (!handle) {
               printf("unable to open device\n");
               return 1;
          }
          printf("iteration: %d, going to call hid_close() \n", i); 
          hid_close(handle);
          printf("iteration: %d, hid_close() done \n", i); 
     }   
     /* Free static HIDAPI objects. */
     hid_exit();
     return 0;
}

Youw:
I wasn't able to reproduce the original issue, presumably because I don't have a device that would allow me to do so.
I believe there is a race condition, that when hit, may lead to deadlock, when trying to close the device, in case if device submits transfers/interruptions frequently, and device is being closed exactly during successful handling of such transfer.

The fix suggested in #143 kind-of fixes the issue. Kind-of, because it constantly tries not to hit a deadlock, but a race condition is still there, it is a root cause, and we better fix a root cause, not symptoms.

@vozhyk-com would you be able to check a patch I prepared, with your device? https://github.com/libusb/hidapi/tree/libusb-hang

vozhyk-com:
I confirm that the patch from branch libusb-hang solves #142 issue. I can not reproduce it with the patch, but easy can reproduce without.

Youw:
That's awesome news. As I hoped, the root issue found and fixed.

PR #143, opened 2020-02-01
Fix: Hanging inside hid_close()

vozhyk-com:
#142 

Function hid_close() waits exit of read_thread().
Thread read_thread() at the end sets 
     libusb_cancel_transfer(dev->transfer);
and goes to 
     while (!dev->cancelled)
               libusb_handle_events_completed(usb_context, &dev->cancelled);

where libusb calls function read_callback() that should set dev->cancelled to 1.

The problem is: in normal transfer->status should be LIBUSB_TRANSFER_CANCELLED,
but time to time it is LIBUSB_TRANSFER_COMPLETED, I do not know why.
in that case dev->cancelled never set to 1 and we hang in a while().

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -844,8 +844,10 @@ static void *read_thread(void *param)
 	   if no transfers are pending, but that's OK. */
 	libusb_cancel_transfer(dev->transfer);
 
-	while (!dev->cancelled)
+	while (!dev->cancelled) {
+		libusb_cancel_transfer(dev->transfer);
 		libusb_handle_events_completed(usb_context, &dev->cancelled);
+	}
 
 	/* Now that the read thread is stopping, Wake any threads which are
 	   waiting on data (in hid_read_timeout()). Do this under a mutex to


Youw:
In case of `LIBUSB_TRANSFER_COMPLETED`, `libusb_handle_events_completed(usb_context, &dev->cancelled);` supposed to set `dev->cancelled` to `TRUE`, and it shouldn't be the case

Maybe there is a race condition which we do not encounter for yet, and we better find and fix it, as even if this patch fixes the behaviour, I think leaving hidden race condition is not a good thing to do in general

Youw:
@LudovicRousseau would you mind to take a look at this? Just because you know how libusb works (at least better than me) you might be able to save us some time
Thanks

LudovicRousseau:
I am not a HIDAPI expert. In fact I do not use this API any more since a few years now.

I don't see why adding a new call to `libusb_cancel_transfer(dev->transfer);` would solve the issue. The first `libusb_cancel_transfer()` should be enough.

Maybe the test in `while(!dev->cancelled)` should be updated to also handle the `LIBUSB_TRANSFER_COMPLETED` case?

vozhyk-com:
> I don't see why adding a new call to `libusb_cancel_transfer(dev->transfer);` would solve the issue. The first `libusb_cancel_transfer()` should be enough.
You are right, but time to time it is not work.
New call at the end sets status to LIBUSB_TRANSFER_CANCELLED.

The question is why first call 'libusb_cancel_transfer(dev->transfer);' does not do this?
 By the way it is not a first call 'libusb_cancel_transfer(dev->transfer);', first call do hid_close() before join to read_thread().
I do not see any other thread that use libusb and can change status to  LIBUSB_TRANSFER_COMPLETED.

vozhyk-com:
libusb_cancel_transfer() - function that asynchronously cancels a transfer. 
I did try to wait until status will be changed to LIBUSB_TRANSFER_CANCELLED,
but it is not work.
`
diff --git a/libusb/hid.c b/libusb/hid.c
index 92fb021..af93360 100644
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -844,8 +844,14 @@ static void *read_thread(void *param)
           if no transfers are pending, but that's OK. */
        libusb_cancel_transfer(dev->transfer);

 \+       printf("status[%d]\n", dev->transfer->status);
\+       while (dev->transfer->status != LIBUSB_TRANSFER_CANCELLED) {
\+               usleep(50);
\+               printf("status[%d] != LIBUSB_TRANSFER_CANCELLED... wait...\n", dev->transfer->status);
\+       }
\+
        while (!dev->cancelled)
                libusb_handle_events_completed(usb_context, &dev->cancelled);
`
Test output:

***
iteration: 209, hid_close() done 
iteration: 210, going to call hid_close() 
status[3]
iteration: 210, hid_close() done 
iteration: 211, going to call hid_close() 
status[0]
status[0] != LIBUSB_TRANSFER_CANCELLED... wait...
status[0] != LIBUSB_TRANSFER_CANCELLED... wait...
status[0] != LIBUSB_TRANSFER_CANCELLED... wait...
status[0] != LIBUSB_TRANSFER_CANCELLED... wait...

and so on infinitely...

Youw:
One more thing: @vozhyk-com what version/build of LibUSB are you using?

Since you're able to reproduce the original issue (I don't have my Linux machine available right now), could you try one more thing? - Enable libusb debug output (in your #142 example): `libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, LIBUSB_LOG_LEVEL_DEBUG);` (maybe inside `hid_init`, for a specific libusb context), and share the output when it hangs.

vozhyk-com:
[hidtest.log.tar.gz](https://github.com/libusb/hidapi/files/4144916/hidtest.log.tar.gz)

I use Debian 10:

Package: libusb-1.0-0
Version: 2:1.0.22-2
Priority: optional
Section: libs
Source: libusb-1.0
Maintainer: Aurelien Jarno <aurel32@debian.org>
Installed-Size: 136 kB
Depends: libc6 (>= 2.17), libudev1 (>= 183)
Homepage: http://www.libusb.info
Tag: role::shared-lib
Download-Size: 55.3 kB
APT-Manual-Installed: no
APT-Sources: http://deb.debian.org/debian stable/main amd64 Packages

Youw:
Closing in favor of #189

PR #148, opened 2020-02-22
Mac: hid_open_path in non exclusive mode

mattsains:
Based on https://github.com/signal11/hidapi/pull/453

### DIFF
--- a/mac/hid.c
+++ b/mac/hid.c
@@ -817,7 +817,7 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
 	}
 
 	/* Open the IOHIDDevice */
-	ret = IOHIDDeviceOpen(dev->device_handle, kIOHIDOptionsTypeSeizeDevice);
+	ret = IOHIDDeviceOpen(dev->device_handle, kIOHIDOptionsTypeNone);
 	if (ret == kIOReturnSuccess) {
 		char str[32];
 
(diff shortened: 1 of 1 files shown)


Youw:
The point is to be able to do it conditionally
The old behavior is might be required too

mattsains:
Can you do that with hidapi on Linux? My understand is that in Linux, hidapi does not use the device in exclusive mode. My intention with this PR is to make sure the behaviour on Mac is the same

Youw:
And you're not the first one :)
There're users, who are already using the library. They are relying on backward compatibility, when updating it.

mattsains:
Do you have examples of users who are expecting hidapi to work this way on Mac?

Youw:
I don't

Do you have proof there is none?

mattsains:
No, but I do have a bunch who didn’t expect it to work the way it does, especially since the behaviour is different for different operating systems, and I don’t think this difference is documented.

I don’t really have the space to implement a new feature to specify which mode to use, so if you’re uncomfortable making the change proposed here then feel free to close this pr and I’ll just use my fork in my project and this bug will have to remain until someone else has time to fix it

Youw:
"I don't have space/time" - is not a good excuse to fix something for some users, and potentially break something for others (we've been there)

I do believe we need to fix this, and yes, someone has to spend additional time to have _this_ fix, AND allow the old behavior, AND don't break API/ABI compatibility for those who don't care

That "someone" is likely going to be myself

Since you came here and raised the issue one more time, I'll move it higher in my TODO list

mattsains:
Although I disagree, I understand your position. Good luck with the project. It would be nice to see this fixed.

Youw:
Thanks, we'll do our best

PR #164, opened 2020-05-14
Add hid_get_path to API resolves #163

### LINKED ISSUE #163: Mapping from hid_device to hid_device_info
Assuming the manufacturer of a device did not use serial_numbers to make particular hid devices unique.  Is there a way to uniquely map from a given already open hid_device opaque pointer / handle back to a hid_device_info struct in the linked list provided by hid_enumerate(0,0)?

I initially thought that I could do something like the following:

````
hid_device_info match_device(hid_device *given_dev) {
  hid_device_info info = {};
  hid_device_info *first = hid_enumerate(0,0);
  for(hid_device_info *cur=first, cur; cur=cur->next) {
    hid_device *test_dev = hid_open_path(cur->path);
    if (test_dev == given_dev) {
      memcpy(&info, cur, sizeof(info));
      info->next = NULL;
    } else {
      hid_close(test_dev);
    }
  }
  hid_free_enumeration(first);
  return info;
}
````

Unfortunately, on windows the second hid_open_path function returns a handle that is not identical to the provided handle.  As such this code can not return a match.

An alternative method would be if there were a function to extract the path from a given hid_device.  In this case, the above code could be modified as following:

````
hid_device_info match_device(hid_device *given_dev) {
  hid_device_info info = {};
  /* hypothetical function to get path given an hid_device pointer */
  const char *path = hid_device_path(given_dev);
  hid_device_info *first = hid_enumerate(0,0);
  for(hid_device_info *cur=first, cur; cur=cur->next) {
     if (strcmp(path, cur->path) == 0) {
      memcpy(&info, cur, sizeof(info));
      info->next = NULL;
      break;
    } 
  }
  hid_free_enumeration(first);
  return info;
}
````

Is there a way to get the path from a hid_device pointer?

gotnone:
This resolves the ability to get the device path from hid_device pointer issue in #163 

Add a new hid_get_path function to the API.  This adds a new path member
to the opaque structure of each hid implementation.  The path is
dynamically allocated and populated in the respective hid_open_path
functions.  The function returns a const char pointer to this path
string.

- Add path string to each hid_device_ structure in hid.c files
- Add hid_get_path function to hidapi.h and hid.c files
- Add init and free code for dev->path to hid.c files
- Allocate and populate dev->path in hid_open_path in hid.c files
- Add example of hid_get_path usage in test.c

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -417,6 +417,16 @@ extern "C" {
 		*/
 		int HID_API_EXPORT_CALL hid_get_indexed_string(hid_device *dev, int string_index, wchar_t *string, size_t maxlen);
 
+		/** @brief Get the Path String from a HID device.
+
+			@ingroup API
+			@param dev A device handle returned from hid_open().
+
+			@returns
+				This function returns a pointer to path string or NULL on error.
+		*/
+		HID_API_EXPORT_CALL const char *hid_get_path(hid_device *dev);
+
 		/** @brief Get a string describing the last error which occurred.
 
 			Whether a function sets the last error is noted in its
--- a/hidtest/test.c
+++ b/hidtest/test.c
@@ -76,6 +76,9 @@ int main(int argc, char* argv[])
  		return 1;
 	}
 
+	// Read the Path to the Device
+	printf("Path: \"%s\"\n", hid_get_path(handle));
+
 	// Read the Manufacturer String
 	wstr[0] = 0x0000;
 	res = hid_get_manufacturer_string(handle, wstr, MAX_STR);
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -917,10 +917,7 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
 						}
 						good_open = 1;
 						/* Save path */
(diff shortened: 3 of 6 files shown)


gotnone:
This PR does have a significant caveat:

I don't have the ability to test this on anything other than windows at the moment.  The use of `char *` may not be appropriate on some implementations (I am thinking of linux with UTF-8 capable filesystems).  Please let me know if you can test this on other hidapi implementations.

gotnone:
After getting my hands on a linux machine.  I can say that the libusb and hidraw implementations also appear to work correctly.  The limitation on this may be the fact that this linux machine was set up to be en_US.UTF-8.

TODO: somehow test on a mac

Youw:
What's on to of my mind - replace `calloc` + `strcpy` with `strdup`
I'll check everything closely a bit later

Youw:
I'm probably going to spend some more time on this.
I think it is possible to postpone initialization of the "path" variable until it is first time called (at least for some backends).

Youw:
I haven't forgot about this one, but I remember I wanted to perform some kind of optimisation, at least for one of the platforms, before merging it.

PR #165, opened 2020-05-22
Add CMake to build HIDAPI

OgreTransporter:
As requested in #12 here a simple CMake script.

But I have only tested it completely under Windows.


Youw:
Thanks for your work!

A few general comments from my side:
> here a simple CMake
- I must say - it doesn't look simple at to me; E.g.: we could use [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) instead of explicitly making separate targets;

- doxygen is definitely out of scope of this PR;
- you've limited minimum version of CMake to 3.16 but haven't really used any feature of it (there is some from CMake 3.14; more then that - you're still using patterns from CMake v2 which are considered deprecated;
- this is more my personal preference, but so far I don't find any one who would completely disagree with me: I'd rather replace FOX libraries with something that is actively supported nowadays (e.g. Qt); that's why I wouldn't introduce support of FOX into a new build system (let's make a clean start);
- yeap, non-Windows targets require a bit more support;

---
But again, thanks for this!
Since someone really want to use CMake with hidapi, it is a good reason for me to finally finish something that I've started some time ago.

OgreTransporter:
Thanks for your feedback! I try to improve the script with your comments.

> I must say - it doesn't look simple at to me; E.g.: we could use BUILD_SHARED_LIBS instead of explicitly making separate targets;

Yeah, I can fix that real quick. It's a habit of mine, because I like to build static and shared versions of most libraries at once.

> you've limited minimum version of CMake to 3.16 but haven't really used any feature of it (there is some from CMake 3.14; more then that - you're still using patterns from CMake v2 which are considered deprecated;

I currently have CMake 3.16 installed, hence these minimum requirements. It is no problem to change them to 3.14. Which CMake-2 do I use? I'll remove them.

> this is more my personal preference, but so far I don't find any one who would completely disagree with me: I'd rather replace FOX libraries with something that is actively supported nowadays (e.g. Qt); that's why I wouldn't introduce support of FOX into a new build system (let's make a clean start);

I agree! Maybe I can change the test program to Qt 5.

OgreTransporter:
I think I renamed BUILD_TESTS to BUILD_TOOLS. To avoid confusion with BUILD_TESTING.

Youw:
> Yeah, I can fix that real quick. It's a habit of mine, because I like to build static and shared versions of most libraries at once.

I suggest we try to optimize the hot-path for most users. As a library user, you usually want to use a single (static or shared) version of the library at once.

> I currently have CMake 3.16 installed, hence these minimum requirements. It is no problem to change them to 3.14. Which CMake-2 do I use? I'll remove them.

As a user of Ubuntu 18.04.4 LTS (which is widely used nowadays), I'd suggest to stick at least with CMake 3.10.2. Maybe even lower, to cover more users off the box.
`include_directories` is from CMake 2 world. `target_include_directories` is a "modern way"

> Maybe I can change the test program to Qt 5.

Please at least let's make a separate PR. Let this be a small simple CMake build system.

Youw:
I'll have some time this weekend, I'll spend more time on hidapi project in general.

CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+install(TARGETS hidapi libhidapi)
+install(FILES hidapi/hidapi.h TYPE INCLUDE)
+
+option(BUILD_TESTS "Build test programms" ON)
+if(BUILD_TESTS)
+  find_package(FOX)
+  if(FOX_FOUND)
+    add_executable(hidapi-testgui ${HIDAPI_SRC} ${HIDAPI_HDR} testgui/test.cpp)
+    target_compile_definitions(hidapi-testgui PUBLIC -DHIDAPI_STATIC)
+	target_include_directories(hidapi-testgui PUBLIC ${FOX_INCLUDE_DIR})
```
z3ntu, line 61: wrong indentation

OgreTransporter:
Unfortunately I cannot create a new PR for the Qt program at the moment. There is a problem with Github somewhere. The code can be found in my repository at https://github.com/OgreTransporter/hidapi/tree/fox-to-qt/testgui. It is built with CMake.

Youw:
I'm not asking you to create a new PR right now, I'm just asking not to add those changes within this PR

bevanweiss:
Is there any progress on this?
Lack of built in CMake support means that for RPCS3 (Playstation3 emulator) we're currently having to use a fork of hidapi.
It'd be nicer to 'live on head' rather than being disconnected as we currently are.

We do have a very trivial static library linkage via CMake, so I'll submit a PR for that, but I doubt it would suit many others... (and it only targets a few platforms and not the full suite that hidapi does)

Youw:
I have a dev-in-progress branch: https://github.com/libusb/hidapi/tree/cmake-support
You can already try it out.

It is almost ready, except:
- I haven't completed testing all possible corner cases (like different versions of BSD);
- the documentation isn't ready;

I designed a CMake build system for 2 possible scenarios:
1)  Build/install as an external project, and use CMake's `find_package` functionality for it (or `ExternalProject_Add` for that matter); `hidapi::hidapi` is an imported target
2) `add_subdirectory`(\<hidapi src root\>/src), and use a regular CMake target `hidapi_hidapi`

Please try it out, if you desire so.

bevanweiss:
@Youw that sounds good, perhaps this PR should be closed (and I'll close my other PR) in preference to a PR from your cmake branch (into master)?
I think having partial CMake support for a subset of platforms is better than none at all :)
It would also open this implementation out to other versions of BSD etc

Youw:
Closing in favour of #220

PR #168, opened 2020-06-06
Consider merging changes to test.c back into libusb/hidapi

lenniea:
Review the (generic) changes I added to test.c (for testing arbitrary custom HID devices instead of hardcoding VID, PID, reports, etc.


Youw:
Thanks for your effort!

test.c is intended to be a _very simple_ example of the API usage, but with this change, it becomes too complicated I'd say. Most of the "example" becomes a int/hex/command-line converters which doesn't help its original intention.

Instead, I'd rather leave this test as is, and recommend to use https://github.com/todbot/hidapitester.

lenniea:
ok but I think its bad to have hardcoded vendor & product ids and
input/output reports in a test program like that.  I attempted to get
basically the same API calls as the original without hardcoding
everything.  The only "addition" I made was to support both 8-bit byte and
16-bit word formatting for a custom HID joystick interface.

-Lennie

Youw:
My position - readability and simplicity in the _example_ is the most value.

That hard-coded device was used by the original author years ago. I don't even know what that is, or does it exist at all.
I'd rather remove that part completely to make the example even simpler. When the bigger part of an example is a helper functions - it doesn't make it a _good example_.

PR #181, opened 2020-08-02
windows/hid.c: Silence some warnings.

UweBonnes:


Youw:
I imagine this would silence MinGW/Cygwin warnings, but I bet you haven't tried this with MSVC

UweBonnes:
Youw: What is the consequence of your remarks? Should the check be guarded by an GCC specific #ifdef? Or should the siliencing be avoided?

Youw:
1) If you want to silence a warning for one compiler - don't introduce a warning for another one. (There is more than one possible way to do so)

2) Yes, I believe so

Youw (approved):

PR #182, opened 2020-08-04
Basic CMake files from rpcs3

bevanweiss:
Here are the basic CMake files that were in use for RPCS3, they will allow for the generation of 'something'


Youw:
For this matter, it can be done exactly like this inside your project, when hidapi is a git submodule or smth.

bevanweiss:
Existing cmake branch more suitable

Issue #144, opened 2020-02-04
crash on device unplug on macOS Catalina (10.15)


Youw:
Typical usage scenario (simplified):
```c
hid_init();
dev = hid_open(...);
// use the device
// ...

// device physically disconnected somewhere here

// ...
// all further hid_* funcitons are gracefully failing

hid_close(dev); // ok here

hid_exit(); // <-- macOS Catalina - crash here
```

The cause (in `hid_close`):
https://github.com/libusb/hidapi/blob/65d22a9c3124d9fad02cc064d734fef84a575202/mac/hid.c#L1122-L1124
+
https://github.com/libusb/hidapi/blob/65d22a9c3124d9fad02cc064d734fef84a575202/mac/hid.c#L1097-L1104

The original comment from Alan:
https://github.com/libusb/hidapi/blob/65d22a9c3124d9fad02cc064d734fef84a575202/mac/hid.c#L1119-L1121

> it's been unplugged, then calling 
 	   IOHIDDeviceClose() will crash

Apparently that was the case for the version of macOS, where this was originally developed.
In Catalina (10.15) update - it is the opposite case: if we skip closing the device, the app crashes on `hid_exit()` (`IOHIDManagerClose`).

---
Ideally, of course, we should always close all handles, and avoid any resource leaks. Unfortunately, cannot do it unconditionally as some users on older systems might experience crashes as the original author.

---
Currently I have access only to macOS Catalina and cannot check the behaviour on other systems.
Would be great to check how it behaves on older systems, at least up(down) to 10.10.

vitalys:
@Youw I'd like to help you with this issue, but I can't reproduce it. 

What I tried:
- [Native Instruments Kontrol F1](https://www.native-instruments.com/en/products/traktor/dj-controllers/traktor-kontrol-f1/) device 
- Test app that writes to device (to highlight buttons/segment indicators)
- My test app doesn't read for the device (is this an issue?)
- I used hidapi from `master` branch

What my app does:
1) `hid_init`
2) `hid_open`
3) `hid_write`, until Ctrl-C is pressed (during this step I log if device is active or unplugged)
4) `hid_close`
5) `hid_exit`
6) Additional `printf` before exiting 

As result I can see final `printf` output, so my app doesn't crash
I tested using: 
- macOS 10.15.6 (Macbook Pro)
- macOS 10.15.1 (VirtualBox)
- macOS 11.0 Beta 5 (VirtualBox)

[...]

PR #186, opened 2020-09-20
Don't leak device handle in macOS 10.10 or newer

Youw:
In one of the early versions of macOS, when you try to close the device
with IOHIDDeviceClose() that is being physically disconnected.
Starting with some version of macOS, this crash bug was fixed,
and starting with macSO 10.15 the opposite effect took place:
in some environments crash happens if IOHIDDeviceClose() is _not_ called.

This patch is to keep a workaround for old versions of macOS,
and don't have a leak in new/tested environments.

Fixes: #144.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -79,7 +79,7 @@ case $host in
 	backend="mac"
 	os="darwin"
 	threads="pthreads"
-	LIBS="${LIBS} -framework IOKit -framework CoreFoundation"
+	LIBS="${LIBS} -framework IOKit -framework CoreFoundation -framework AppKit"
 	;;
 *-freebsd*)
 	AC_MSG_RESULT([ (FreeBSD back-end)])
--- a/mac/Makefile-manual
+++ b/mac/Makefile-manual
@@ -12,7 +12,7 @@ CC=gcc
 COBJS=hid.o ../hidtest/test.o
 OBJS=$(COBJS)
 CFLAGS+=-I../hidapi -Wall -g -c 
-LIBS=-framework IOKit -framework CoreFoundation
+LIBS=-framework IOKit -framework CoreFoundation -framework AppKit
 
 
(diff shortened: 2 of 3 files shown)


Youw:
@vitalys if you'd be able to check this patch in your environments (including macOS 10.11-10.15) - that would be fantastic.
I personally tested this on macOS 10.13-10.15 some while ago.
And one of colleagues claimed that it has to work for 10.11-10.12 too, but I never has a chance to check it myself.

vitalys:
@Youw I made first round of testing on macOS 10.15.6. It doesn't crash. I'm going to start testing on different macOS version tomorrow. It won't be fast, because I have to install different macOS versions to my test device, but I'm going to cover all latest macOS versions in 10.10-11.0 range

Youw:
Sounds awesome.
NOTE: My current patch only covers macOS 10.11.
If you'll be able to confirm that it works for 10.10 too (the condition has to be changed: `is_macos_10_10_or_greater = (NSAppKitVersionNumber >= `**`1343`**`); /* NSAppKitVersionNumber10_10 */`, I'll update the patch to include it too.

vitalys:
- 11.0 Beta 8 — ✅ 
- 10.15.7 — ✅ 
- 10.15.6 — ✅ 
- 10.15 — ✅ 
- 10.14.6 — ✅ 
- 10.13.6 — ✅ 
- 10.12.6 — ✅ 
- 10.11.6 — ✅ 
- 10.10.5 (using current version) — ✅
- 10.10.5 (using NSAppKitVersionNumber10_10 version) — ✅

@Youw Could you upgrade your PR and I'll double-check on 10.10.x?

Youw:
done

vitalys:
@Youw I've checked it on 10.10.5 and it works without issues

Youw:
awsome

PR #191, opened 2020-09-27
MSVC: fix /W4 /WX compilation

Youw:


Youw:
Better to check diff with `Hide whitespace changes` option

PR #194, opened 2020-10-02
MSVC: crash bugfix when trying to open device without a serial number

kteynor:
Currently trying to call `hid_open` on a device that doesn't have a serial number will result in a crash in the call to wcscmp here. Check that it's not null before calling wcscmp.


Youw (approved):
Thanks much!
The only thing that bothers me, is the fact that we threat this kind of issues (no serial number) differently on different platforms.
E.g.: on macOS serial_number may be an empty string "", but it never be a null pointer.
But the fix definitely doesn't make things any worse.

PR #205, opened 2020-10-26
Read podspec version from VERSION file

vitalys:
I've successfully published `hidapi.podspec` for v0.10.0 using this file. With this change we don't have to edit podspec file to release next versions of hidapi

### DIFF
--- a/dist/hidapi.podspec
+++ b/dist/hidapi.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |spec|
 
   spec.name         = "hidapi"
-  spec.version      = "<fill me up from VERSION file, before submit>"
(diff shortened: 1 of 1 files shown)


Youw (approved):

Youw:
BTW, 0.10 requires `AppKit`
don't see it within https://github.com/CocoaPods/Specs/blob/master/Specs/1/4/5/hidapi/0.10.0/hidapi.podspec.json

vitalys:
😱 
Let me check and fix this

vitalys:
pod spec lint works without issues, but I've added it to avoid issues

Youw (approved):

PR #207, opened 2020-11-12
hidapi.h: fix API prototype for gcc flag -Wstrict-prototypes

borneoa:
With gcc 10.2.0 and compile flag '-Wstrict-prototypes' the new
APIs in hidapi.h trigger a compile warning:

../hidapi/hidapi.h:481:32: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  481 |   HID_API_EXPORT const  struct hid_api_version* HID_API_CALL hid_version();
      |                                ^~~~~~~~~~~~~~~
../hidapi/hidapi.h:491:3: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  491 |   HID_API_EXPORT const char* HID_API_CALL hid_version_str();
      |   ^~~~~~~~~~~~~~

To replicate:
	./bootstrap
	CFLAG=-Wstrict-prototypes ./configure
	make

This is an issue for other projects, like OpenOCD, that use hidapi
and compile by default with '-Wstrict-prototypes -Werror'; the
warning above causes compile error while including hidapi.h.

Fix the prototypes by adding 'void' in the empty parameter list.

Signed-off-by: Antonio Borneo <borneo.antonio@gmail.com>

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -478,7 +478,7 @@ extern "C" {
 			@returns
 				Pointer to statically allocated struct, that contains version.
 		*/
-		HID_API_EXPORT const  struct hid_api_version* HID_API_CALL hid_version();
+		HID_API_EXPORT const  struct hid_api_version* HID_API_CALL hid_version(void);
 
 
 		/** @brief Get a runtime version string of the library.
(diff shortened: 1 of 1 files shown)


Youw (approved):

z3ntu:
Can we enable all these warning flags in CI as well that we're fixing so we don't regress?

borneoa:
Yes, it could be good to add more strict check during CI.
With OpenOCD we use
-Wall -Wstrict-prototypes -Wformat-security -Wshadow -Wextra -Wno-unused-parameter -Wbad-function-cast -Wcast-align -Wredundant-decls -Werror
not sure if this set covers all needs of hidapi, but tested after this patch does not show any further warning.

PR #215, opened 2020-11-26
remove superfluous whitespace at end of lines

prusnak:
This PR removes superfluous whitespace at end of lines

As can be seen here, only whitespace changes are included - https://github.com/libusb/hidapi/pull/215/commits/2f110a5028523a590da9885d7aeac43dc6799b3b?w=1

### DIFF
--- a/LICENSE-orig.txt
+++ b/LICENSE-orig.txt
@@ -3,7 +3,7 @@
 
  Copyright 2009, Alan Ott, Signal 11 Software.
  All Rights Reserved.
- 
+
  This software may be used by anyone for any reason so
  long as the copyright notice in the source files
  remains intact.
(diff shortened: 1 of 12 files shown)


Youw:
Do you think it is really so much needed?
There is not much development going on, but when one tries to investigate an issue/find out why certain things are implemented this way, and not the other - changes like this makes it a bit more difficult to investigate the history of changes.

prusnak:
I don't mind it that MUCH, but my editor shows whitespace and some files have ridiculous amount of extra whitespace, so it triggers me a bit :D

Youw:
Yeah, I know the filling :)

For the reason of "not clutter the history" I vote not to merge it.

Youw:
As much as I love the idea of having a clean code base, at this point lets preserve commit history.
Maybe we'll come back to it some time later.

PR #220, opened 2020-12-11
Add CMake build system

Youw:
This is still work-in-progress. A draft PR, so can be reviewed by community.


CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+add_subdirectory(src)
+
+set(BUILD_HIDTEST_DEFAULT OFF)
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(BUILD_HIDTEST_DEFAULT ON)
+endif()
+option(BUILD_HIDTEST "Build small console test application hidtest" ${BUILD_HIDTEST_DEFAULT})
+if(BUILD_HIDTEST)
+    add_subdirectory(hidtest)
```
Qix-, line 30:
```suggestion
    enable_testing()
    add_subdirectory(hidtest)
```

hidtest/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+project(hidtest C)
+
```
Qix-, line 2: Not necessary here unless `hidtest` is meant to be a root config. It might actually cause problems.
```suggestion
```

hidtest/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+set(HIDAPI_HIDTEST_TARGETS)
+if(NOT WIN32 AND NOT APPLE AND CMAKE_SYSTEM_NAME MATCHES "Linux")
```
Qix-, line 4:
```suggestion
if(UNIX AND NOT APPLE)
```

hidtest/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+    endif()
+else()
+    add_executable(hidtest test.c)
+    target_link_libraries(hidtest hidapi)
+    list(APPEND HIDAPI_HIDTEST_TARGETS hidtest)
+endif()
+
+install(TARGETS ${HIDAPI_HIDTEST_TARGETS}
+    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+)
```
Qix-, line 23: This isn't common - `install()` is meant for production targets, not tests. Better to rely on the CTest suite.

hidtest/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+project(hidtest C)
+
+set(HIDAPI_HIDTEST_TARGETS)
+if(NOT WIN32 AND NOT APPLE AND CMAKE_SYSTEM_NAME MATCHES "Linux")
+    if(TARGET hidapi_hidraw)
+        add_executable(hidtest_hidraw test.c)
+        target_link_libraries(hidtest_hidraw hidapi_hidraw)
+        list(APPEND HIDAPI_HIDTEST_TARGETS hidtest_hidraw)
```
Qix-, line 8:
```suggestion
        add_test(NAME hidapi_test_hidraw COMMAND $<TARGET_FILE:hidtest_hidraw>)
```

hidtest/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+    endif()
+    if(TARGET hidapi_libusb)
+        add_executable(hidtest_libusb test.c)
+        target_link_libraries(hidtest_libusb hidapi_libusb)
+        list(APPEND HIDAPI_HIDTEST_TARGETS hidtest_libusb)
```
Qix-, line 13:
```suggestion
        add_test(NAME hidapi_test_libusb COMMAND $<TARGET_FILE:hidtest_libusb>)
```

hidtest/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+    endif()
+else()
+    add_executable(hidtest test.c)
+    target_link_libraries(hidtest hidapi)
+    list(APPEND HIDAPI_HIDTEST_TARGETS hidtest)
```
Qix-, line 18:
```suggestion
    add_test(NAME hidapi_test COMMAND $<TARGET_FILE:hidtest>)
```

libusb/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+cmake_minimum_required(VERSION 3.6.3 FATAL_ERROR)
```
Qix-, line 1: This should be in the very root of the project, not in a subdirectory.

Qix- (changes requested):
Going to keep it at what I have now so it's not too overwhelming. Let me know if you'd like me to keep going or not :)

Qix-:
By the way, Travis removed support for open source projects as of this last November. You might want to consider moving entirely to sourcehut (since I see you're already using it) or to Github Actions/Circle/etc.

The travis jobs will probably never complete and thus stay yellow :)

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.1.3 FATAL_ERROR)
```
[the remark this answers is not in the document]
Youw: > nobody is on cmake <3 anymore

You will be surprised.

Although, define "legacy". It is not deprecated, and it is not recommended by CMake developers not to use.

Effectively, without it we only fail to give one good error message, in case if someone _accidentally_ tries to use a very old CMake.

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+project(hidapi LANGUAGES C)
+
+if(NOT DEFINED CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE STREQUAL "")
+    set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Debug;Release;MinSizeRel;RelWithDebInfo" FORCE)
+endif()
+set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+
```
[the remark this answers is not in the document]
Youw: so instead of type less, you suggest always type more (and possible type more errors)
I vote for user-friendly

> muscle memory for me personally

and it makes it harder to use for those, who don't have such muscle memory

---
even more: I use cmake-gui even more often then a command-line. It saves me much time. And it saves me time to pick a selection from a combo-box, instead of typing it manually each time.

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+if(APPLE)
+    option(CMAKE_FRAMEWORK "Build macOS/iOS Framework version of the library" OFF)
```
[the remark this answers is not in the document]
Youw: you're mistaken, this is not a HIDAPI option, this is a [standard CMake option](https://cmake.org/cmake/help/latest/variable/CMAKE_FRAMEWORK.html), and it is selected into an `option` for convenience of those, who like cmake-gui as much as I do

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+elseif(NOT WIN32)
+    if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+        option(HIDAPI_WITH_HIDRAW "Build HIDRAW-based implementation of HIDAPI" ON)
+        option(HIDAPI_WITH_LIBUSB "Build LIBUSB-based implementation of HIDAPI" ON)
+    endif()
+endif()
```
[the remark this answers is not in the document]
Youw: > Linux and no other *NIX distributions

I believe this is exactly what is originally written

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+option(BUILD_SHARED_LIBS "When set: build shared version of the libraries, otherwise - static" ON)
```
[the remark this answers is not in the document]
Youw: you're absolutely right
again, without it - cmake-gui doesn't give a nice hint about it

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+set(HIDAPI_INSTALL_TARGETS ON)
```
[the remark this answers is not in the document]
Youw: unfortunately you don't know the design that is hidden behind it

in short - the root CMakeLists.txt is not designed to be a subdirectory of a larger project `src/CMakeLists.txt` is designed for that
I'll surely explain that in the documentation/usage examples as soon as I can get my hands to it

hidtest/CMakeLists.txt
Youw: it is not the same, this is really means "Linux only, and no other UNIX system"

hidtest/CMakeLists.txt
Youw: this is not a (unit) test
this is a user application example, yes - a normal target

hidtest/CMakeLists.txt
Youw: again, this is not a test in terms of CTest

libusb/CMakeLists.txt
Youw: unfortunately, we have different minimum version requirements for different backends
my intention is to keep lowest minimum version possible, and still use modern CMake with minimum compromises

hidtest/CMakeLists.txt
Youw: > is meant to be a root config

a root config of a _project_
this _is_ a separate project called `hidtest`, that is an example of usage of the `hidapi` _library_ (which is a _separate project_ too)

both happened to be under the same build tree, and that is absolutely fine, CMake is absolutely well designed for it

> It might actually cause problems

I don't believe so, not without misuse of CMake

CMakeLists.txt
Youw: `hidtest` is not testing module, as I explained in my other comments

Youw:
> Travis removed support for open source projects as of this last November

Fortunately, that is not true.
They are just migrating all Open Source projects from travis-ci.org to travis-ci.com. Check out their [open letter](https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom). And of course, it remains [free for Open Source](https://travis-ci.com/plans) as it was before.

Youw:
> Let me know if you'd like me to keep going or not :)

I guess I'd rather finish with the documentation first, so we can avoid my answers like "this is by current design".
This whole implementation is designed to make the _usage simple_, not the implementation simple.

Youw:
> The travis jobs will probably never complete and thus stay yellow :)

This is a good one, thanks. I'll switch to travis-ci.com. I though I already did it.

CMakeLists.txt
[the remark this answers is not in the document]
Qix-: None of the other functions in the CMake file would work if they were on <3. That's why you don't see it in modern CMake configs anymore.

CMakeLists.txt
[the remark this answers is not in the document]
Qix-: > so instead of type less, you suggest always type more

In this case, yes. It's not about typing less/more. It's about not colliding with CMake builtins in the future.

Further, most people that will be using this will be either installing from a package manager or using it as a submodule - and only the latter would require someone typing it once to configure it.

> and it makes it harder to use for those, who don't have such muscle memory

The point wasn't that I'm better. The point is that it's a pretty standard way to use CMake.

> even more: I use cmake-gui even more often then a command-line

That is not the majority case.

CMakeLists.txt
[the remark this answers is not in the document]
Qix-: You're right, google didn't bring it up for some reason. Ignore.

CMakeLists.txt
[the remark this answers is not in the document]
Qix-: Fair enough.

CMakeLists.txt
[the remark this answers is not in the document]
Qix-: >  the root CMakeLists.txt is not designed to be a subdirectory of a larger project

It should be. That's how CMake libraries should be structured. The fact they can be installed system-wide is an extra thing, not the default.

You cut out users that want to statically link without polluting their system cache, or those who are cross compiling, etc.

CMakeLists.txt
Qix-: Ah okay.

hidtest/CMakeLists.txt
Qix-: It doesn't add anything, that's the point. Unless you're distributing these sources separately, then it doesn't add anything.

Qix-:
> They are just migrating

Huh, I read something entirely different on HN. I guess that's good news, then.

CMakeLists.txt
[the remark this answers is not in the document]
Youw: > None of the other functions in the CMake file would work if they were on <3

That is incorrect. Strictly speaking: "_some_ of the functions in this build tree will not work, _or will have a different outcome, that the script author intended_"

CMakeLists.txt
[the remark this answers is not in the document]
Youw: > or using it as a submodule - and only the latter would require someone typing it once to configure it.

And what problem does it make? Does it limit the users?

> The point wasn't that I'm better. 

Neither was mine.

> The point is that it's a pretty standard way to use CMake.

For all my "small" project I have a different muscle memory: `cmake .. && make -j`. Pretty standard to me (and I didn't invented it).

CMakeLists.txt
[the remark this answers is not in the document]
Youw: > It should be. That's how CMake libraries should be structured.

I don't suppose that statement is based on large number of different real-life CMake-based projects/libraries.

---
I had a pleasure integrating maybe a dozen CMake-based libraries into a larger CMake-based solution/build tree (as a submodule). And none of those projects could be integrated by simple `add_subdirectory(<subproject root>)`, and I don't believe that is because those projects or their build systems are badly implemented, not at all.
Although, a solution/build tree I mentioned above has lots of demands, some of them: it has to compile/deploy on at least 4 platforms: Windows, macOS, Linux, Android, including host x86_64 builds and ARM64 cross-compilations. All these and other requirements automatically applies to all dependencies.

Per my observations, each projects's build system consist of of two logical parts: _build infrastructure_ and _build definitions_. (It is not always possible to tell exactly where the infrastructure ends, and the definitions begins, but still.)
CMake itself is an infrastructure. It is designed to suit as much different possible needs as possible. That also means that it will never work of-the-box for everyone. CMake has many [variables](https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html) and [properties](https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html) to configure the build system, but it doesn't mean, that any combination of those values/properties can by applied to any project successfully. So each project (its author(s)) choose a subset of its variables/properties, maybe its values, into an actual CMake options that are intended to be configurable/changed. That all together makes a build infrastructure, that answers to question "how?": how to build a target, which compiler options, where to install, what kind of the binary should be, etc.
At this point we didn't even start developing our project, we don't have build definitions yet, we don't have our "what". Q: _what_ do we want to? - Make a library. So we add a build definition (a target): `add_library(...)`. There are certain properties of each target, that are absolutely required to successfully build a target, e.g.: macOS backend of hidapi has to be linked against `IOKit framework`, otherwise of won't build at all. That is essential, that has to be part of build definition. It cannot be optional or configurable.
On the other hand, we can build our source with `-O2` or `-g -O0` compiler options, but that doesn't change the essence/intention of the project. Maybe performance, at most, maybe convenience for the developer (e.g. ability to debug), and that is a part of the project's infrastructure.

My point is that root CMakeLists.txt usually contains a lot of infrastructure for convenience of the project's developer. Because project's developer keeps its focus on the ability to integrate the project into a package manager, install into a system (and make the library available for other projects), deploy it to production, ability to develop it fast and efficient.
hidapi is not an exception here. All the options are designed for _maximum convenience of the user/developer_, who tries to build hidapi as a _standalone project/library_. Install project or not - that is a question to the user: call/build `install` target or not, it shouldn't be a question to the build system as "add `install` target or not?".
I tried to cover all possible cases I could thing of. Most CMake-based projects I've seen so far, are actually doing the same in one degree or another. Without a good infrastructure, a usage of the project's build system is not impossible, but less convenient/harder for sure.

On the other hand, when someone wan'ts to use a hidapi as a subdirectory of the existing CMake build tree, it likely means they _already have its own build infrastructure_. CMake options that _make perfect sense for hidapi_ might have absolutely _no sense to the other project_ that want's to use hidapi (e.g. as exactly a single static library). So the only thing other project needs from hidapi - its _build definitions_, and minimum dependencies.
By current design, root CMakeLists.txt of hidapi contains an _infrastructure_. `src/CMakeLists.txt` and below - that's definitions. As an author of the build system, I've tried to design it that way, and only because I've thought of all that even before I've started the development.
Unfortunately not all of the infrastructure can be removed from definitions (at least, because some of the infrastructure is actually what makes definitions possible), so all of the infrastructure parts that are left are designed to be configurable as much as possible, to cover the majority of possible usages.

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.1.3 FATAL_ERROR)
+project(hidapi LANGUAGES C)
+
+if(NOT DEFINED CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE STREQUAL "")
+    set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Debug;Release;MinSizeRel;RelWithDebInfo" FORCE)
+endif()
+set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
```
Be-ing, line 7: What is the point of this? Why don't you leave it to CMake to set the default?

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+if(APPLE)
+    option(CMAKE_FRAMEWORK "Build macOS/iOS Framework version of the library" OFF)
+elseif(NOT WIN32)
+    if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+        option(HIDAPI_WITH_HIDRAW "Build HIDRAW-based implementation of HIDAPI" ON)
+        option(HIDAPI_WITH_LIBUSB "Build LIBUSB-based implementation of HIDAPI" ON)
```
Be-ing, line 14: Is the libusb backend usable on Windows?

CMakeLists.txt
Be-ing, line 18:
```suggestion
option(BUILD_SHARED_LIBS "Build shared version of the libraries, otherwise build statically" ON)
```

Be-ing:
What is left to do here?

Be-ing:
We're considering using vcpkg for our dependencies for Mixxx but the package of hidapi in vcpkg relies on the VS project file. It would be great to have this merged so hidapi could be installed easily on vcpkg on Windows, macOS, and Linux.

CMakeLists.txt
Youw: > What is the point of this? 

It is for convenience of those, who are building hidapi as a standalone package/library. Try cmake-gui and you'll see.

> Why don't you leave it to CMake to set the default?

Because by default CMake doesn't set it at all. Maybe it does makes sence somtime, but I haven't seen a single real project where it would.

---
As for the convenience part: see my other comment https://github.com/libusb/hidapi/pull/220#discussion_r541672007

CMakeLists.txt
Youw: Theoretically speaking - it _could be_ usable.
Practically: its implementation relies on POSIX build environment.

I believe it is possible to replace all unix/linux-specific dependencies and replace it with C11, but that is not done yet.

Youw:
> What is left to do here?

There're a few things, e.g. https://github.com/libusb/hidapi/pull/227 (or rather its variance).
There is also other [issue reported](https://github.com/libusb/hidapi/issues/12#issuecomment-749001253).

And, the most important - the documentation... I can't get my hands to it.

Since [vcpkg](https://github.com/microsoft/vcpkg/pull/15657) depends on it more and more, I'll try to finish up things sooner.

CMakeLists.txt
Be-ing: I still don't get the point of overriding CMake's default behavior, but I guess it doesn't hurt? :shrug: It'll just be a bit of confusing code for people reading this in the future.

CMakeLists.txt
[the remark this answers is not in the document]
Qix-: The point here is that this is not common. Another user now has asked about it; you generally don't set these things, because debug/release isn't binary. I've rarely, hardly ever actually, seen a config that does this. It's the general case yo specify the config type you want to build for.

CMakeLists.txt
Youw: > but I guess it doesn't hurt?

That's my point.

> CMake's default behavior

Humor me. Can you tell what's CMake's default behavior if you don't specify CMAKE_BUILD_TYPE at all?

jm4R: @Youw 
There are multi-config generators (like MSVC) that ignores `CMAKE_BUILD_TYPE`. The value of this variable is also not limited to the ones you've listed - for example I use also `TSan` and `ASan` build type for sanitizers. I agree this part of code isn't necessary here but I don't think it will break something on the other hand.

Youw: > ignores `CMAKE_BUILD_TYPE`

The problem that it is not exactly ignored. It is still available in the CMake scripts, and what bothers me the most: CMake doesn't default it to empty value when MSVC generator is used, but rather to "Debug" (gcc + make - it is defaulted to empy value).
I find such behavior way more confusing, but I guess CMake's developers must have had their reasons for that.

> I use also `TSan` and `ASan` build type for sanitizers

I must admit - I haven't seen any custom CMAKE_BUILD_TYPE usage/implementation, so I guess that is a good point too.
Can you point to me a reference how your/typical usage look like?

As I mentioned earlier, a standalone build likely not to be used with a non-standard build types but I agree that prohibiting it wouldn't be a good thing, so there is [that](https://github.com/libusb/hidapi/commit/8ae8ffb2cf0ef6596baa55a9c8db3fc2dfc31e9f).

.github/workflows/macos.yml
```diff
@@ -24,3 +24,28 @@ jobs:
     - name: Build Manual makefile
       working-directory: mac
       run: make -f Makefile-manual
+
+  build-cmake:
+
+    runs-on: macos-latest
```
Be-ing, line 30: Use GitHub Actions' matrix feature to avoid having separate jobs for each OS. For example: https://github.com/Be-ing/portaudio/blob/3b2bf31b447c74919ded00fd0552e566dcada171/.github/workflows/c-cpp.yml

Youw: thanks, I'll look into it

Youw: btw, does the matrix variances run in parallel?

Be-ing:
Use `set_target_properties(hidapi PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS TRUE)` and you can get rid of the `__declspec(dllexport)` hack for MSVC.

.github/workflows/macos.yml
Be-ing: Yup! And GitHub Actions is very generous with computing resources. You get 20 concurrent jobs for free (max 5 macOS jobs).

Youw:
> WINDOWS_EXPORT_ALL_SYMBOLS 

This property is supported starting with CMake 3.4, and Windows build is working just fine with 3.1.13.

In any case, CMake is not first nor only build system for HIDAPI. autotools build is there since the very beginning and it has to work as well.

---
Also, I personally find it to be not the best thing to do. In all my projects I use `CMAKE_C_VISIBILITY_PRESET hidden` and always explicitly specifying what to export with `__declspec(dllexport)`, `__declspec(dllimport)` and `__attribute__((visibility("default")))`.
This is more common practice and more portable across different build systems.

Be-ing:
> In any case, CMake is not first nor only build system for HIDAPI. autotools build is there since the very beginning and it has to work as well.

I strongly recommend removing all build systems besides CMake. There is no reason to maintain systems that CMake can generate.

Be-ing:
Another reason to remove other build systems is that downstream projects can't depend on the autogenerated CMake config files for `find_package` if some package managers choose to use a different build system.

Be-ing:
> I strongly recommend removing all build systems besides CMake. There is no reason to maintain systems that CMake can generate. Otherwise, you will get bug reports that are specific to one build system or another.

Case in point: https://github.com/libsndfile/libsamplerate/issues/140

Be-ing:
Ping. It would be great to get this finished to get hidapi support on macOS and Linux in vcpkg.

src/CMakeLists.txt
```diff
@@ -0,0 +1,148 @@
+get_filename_component(PROJECT_ROOT "${CMAKE_CURRENT_LIST_DIR}/.." ABSOLUTE)
+
+# Read version from file
+file(READ "${PROJECT_ROOT}/VERSION" VERSION_STR)
+string(REGEX MATCH "^([0-9]+\\.[0-9]+\\.[0-9]+)(.*)" VERSION_STR "${VERSION_STR}")
```
madebr, line 5: Is the regex guaranteed to succeed?
Maybe
```cmake
if(NOT VERSION_STR)
    message(FATAL_ERROR "Failed to extract version from ${PROJECT_ROOT}/VERSION")
endif()
```

Or just read the file using `file(READ)` + `string(STRIP)`?

src/CMakeLists.txt
```diff
@@ -0,0 +1,148 @@
...
+            if(NOT TARGET usb-1.0)
+                set(HIDAPI_NEED_EXPORT_LIBUSB TRUE)
+            endif()
+        endif()
+    elseif(NOT TARGET hidapi_hidraw)
+        message(FATAL_ERROR "Select at least one option to build: HIDAPI_WITH_LIBUSB or HIDAPI_WITH_HIDRAW")
+    endif()
+endif()
+
+add_library(hidapi ALIAS hidapi_${EXPORT_ALIAS})
```
madebr, line 125: Making available the target using the same name (`hidapi::hidapi`) as `hidapi-config.cmake`, makes it possible to use `hidapi` as a subproject or as installed project interchangeably. 
```suggestion
add_library(hidapi ALIAS hidapi_${EXPORT_ALIAS})
add_library(hidapi::hidapi ALIAS hidapi_${EXPORT_ALIAS})
```

This enabled:

```cmake
if(USE_INSTALLED_HIDAPI)
    find_package(hidapi REQUIRED)
else()
    add_subdirectory(vendored/hidapi)
endif()
target_link_libraries(my_exe PRIVATE hidapi::hidapi)
```

windows/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+add_library(hidapi_winapi
+    ${HIDAPI_PUBLIC_HEADERS}
+    hid.c
+)
+target_link_libraries(hidapi_winapi
+    PUBLIC hidapi_include
+    PRIVATE Setupapi
```
madebr, line 7: mingw on Linux is case sensitive.
```suggestion
    PRIVATE setupapi
```

src/CMakeLists.txt
Youw: good idea, let me test this out

windows/CMakeLists.txt
Youw: and this is the caseness those use? The "correct" one (as per WindowsSDK) is even different:
image

Youw:
Keeping the heads-up: I got my hands to the documentation for CMake (finally) over the weekend. I'd call it 80% done. Soon I'll be confident-enough to make this PR ready.

windows/CMakeLists.txt
madebr: This line indicates `libsetupapi.a` is generated:
https://github.com/mirror/mingw-w64/blob/c6e13e0c105eab7797c2373819b49fff6b05566c/mingw-w64-crt/lib64/Makefile.am#L539

Youw: as much as I personally don't like the idea of tuning a "Windows" backends for non-windows build environments, I guess an ideology itself is not a very good reason for keeping Linux/mingw off-the-board

madebr: This change will benefit all cross building.
The problem would appear as well when building using msys2 on Windows.

CMakeLists.txt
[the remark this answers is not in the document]
madebr: You can detect whether you're the main project or included as a subproject as follows:
```cmake
cmake_minimum_required(VERSION 3.0)
if(PROJECT_SOURCE_DIR)
    message(STATUS "Included as a subproject")
    set(HIDAPI_MAIN_PROJECT OFF)
else()
    message(STATUS "I'm the main project")
    set(HIDAPI_MAIN_PROJECT ON)
endif()
project(libusb)
```

That way, there's no need for two `CMakeLists.txt` files.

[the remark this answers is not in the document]
Youw: The two `CMakeLists.txt` files are still there, just because the defaults/behavior will be a bit different, when built as a sub_directory and as a standalone project, but yes, this idea is already implemented: https://github.com/libusb/hidapi/commit/81291b205b52013019c2498620b6d27f2ee3bcc9

And technically, checking for `PROJECT_SOURCE_DIR` isn't the same as checking for `CMAKE_/CURRENT_/SOURCE_DIR`, and isn't strictly correct. No one prohibits you to do something like:
```cmake
cmake_minimum_required(VERSION 3.0)

add_subdirectory(hidapi)

project(my_awsome_project)

#...
```

sleiner:
I just tested this in two setups and it works great 👍🏻 Is anything holding back the merge? Is there anything I could do to help? 😊

mcuee:
I like this CMake improvement as well. Thanks.

mcuee:
It seems to work fine.

```
hidapi_cmake/build on  cmake-support ❯ cmake ..
-- The C compiler identification is AppleClang 12.0.5.12050022
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Looking for pthread.h
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/mcuee/build/hidapi/hidapi_cmake/build

hidapi_cmake/build on  cmake-support [?] via △ v3.20.3 ❯ vim CMakeCache.txt 
(to enable hidtest and framwork build, proper way is to define in the command line).

hidapi_cmake/build on  cmake-support [?] via △ v3.20.3 took 1m27s ❯ cmake ..
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/mcuee/build/hidapi/hidapi_cmake/build

hidapi_cmake/build on  cmake-support [?] via △ v3.20.3 ❯ make
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/mcuee/build/hidapi/hidapi_cmake/build
[ 25%] Building C object src/mac/CMakeFiles/hidapi_darwin.dir/hid.c.o
[ 50%] Linking C shared library hidapi.framework/hidapi
Copying OS X content src/mac/hidapi.framework/Versions/0/Headers/hidapi.h
[ 50%] Built target hidapi_darwin
[ 75%] Building C object hidtest/CMakeFiles/hidtest.dir/test.c.o
[100%] Linking C executable hidtest
[100%] Built target hidtest
```

src/CMakeLists.txt
Youw: as it appears, in case of broken VERSION file the configuration would still fail on later steps (project() command would fail) but an explicit check at least makes the the error message more readable

src/CMakeLists.txt
Youw: scripts updated

Be-ing:
What is left to do here?

Youw:
https://github.com/libusb/hidapi/pull/220/files#diff-0bc78e3e3203f9abc8d4db2b1e58170a5948f03e807325064aead14f8b310cc2R8

Working on it right now. Will finish today/tomorrow.
When I do - I'll squash all changes into two commits (one for CMake and one for Documentation update) and will open a new PR. Simply not to break this branch if someone is already using it.

Youw:
> It seems to work fine under my Mac Mini M1 (macOS Big Sur ARM64)

Thanks for testing it. I don't have M1 mac to check it myself. Even though there is nothing platform/architecture-specific that could suggest that the build could have failed, having a confirmation is always a good thing.

BUILD.cmake.md
```diff
@@ -0,0 +1,203 @@
...
+```sh
+cd <build dir>
+# configure the build
+cmake -GNinja <HIDAPI source dir>
+# build it!
+ninja
+# by default installs into /usr/local/
+ninja install
+# NOTE: you need to run `ninja install` command as root, to be able to install into /usr/local/
+```
```
Be-ing, line 38: This assumes the user has Ninja installed. I recommend removing the reference to any particular generator and using the default generator and generic instructions:
```
cmake -S . -B build
cmake --build build
cmake --install build
```

BUILD.cmake.md
```diff
@@ -0,0 +1,203 @@
...
+- [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) - standard possible values: `Debug`, `Release`, `RelWithDebInfo`, `MinSizeRel`; Defaults to `Release` for HIDAPI, if not specified;
+- [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) - when set to TRUE, HIDAPI is built as a shared library, otherwise build statically; Defaults to `TRUE` for HIDAPI, if not specified;
+
+<details>
+  <summary>macOS-specific variables</summary>
+
+  - [`CMAKE_FRAMEWORK`](https://cmake.org/cmake/help/latest/variable/CMAKE_FRAMEWORK.html) - (since CMake 3.15) when set to TRUE, HIDAPI is built as a framework library, otherwise build as a regular static/shared library; Defaults to `FALSE` for HIDAPI, if not specified;
+  - [`CMAKE_OSX_DEPLOYMENT_TARGET`](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html) - minimum version of the target platform (e.g. macOS or iOS) on which the target binaries are to be deployed; defaults to a maximum supported target platform by currently used XCode/Toolchain;
```
Be-ing, line 60: Shouldn't this be the minimum supported target platform?

BUILD.cmake.md
```diff
@@ -0,0 +1,203 @@
...
+
+Available CMake targets after successful `find_package(hidapi)`:
+- `hidapi::hidapi` - indented to be used in most cases;
+- `hidapi::include` - if you need only to include `<hidapi.h>` but not link against the library;
+- `hidapi::winapi` - same as `hidapi::hidapi` on Windows; available only on Windows;
+- `hidapi::darwin` - same as `hidapi::hidapi` on macOS; available only on macOS;
+- `hidapi::libusb` - available when libusb backend is used/available;
+- `hidapi::hidraw` - available when hidraw backend is used/available on Linux;
+
+**NOTE**: on Linux often both `hidapi::libusb` and `hidapi::hidraw` backends are available; in that case `hidapi::hidapi` is an alias for **`hidapi::libusb`**. The motivation is that hidraw backend doesn't have an implementation of all of the HIDAPI functions due to lack of Linux kernel support. If libusb backend isn't built at all (`hidapi::hidraw` is the only target) - `hidapi::hidapi` is an alias for `hidapi::hidraw`.
```
Be-ing, line 125: What is not implemented by the hidraw backend? Where is that documented? I am doubting libusb should be the default backend because it can't work with Bluetooth HID devices.

CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+cmake_minimum_required(VERSION 3.1.3 FATAL_ERROR)
+
+if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+    add_subdirectory(src)
+    # compatinility with find_package() vs add_subdirectory
+    set(hidapi_VERSION "${hidapi_VERSION}" PARENT_SCOPE)
+    return()
+endif()
+# All of the below in this file is meant for a standalone build.
+# When building as a subdirectory of a larger project, most of the options may not make sence for it,
```
Be-ing, line 10:
```suggestion
# When building as a subdirectory of a larger project, most of the options may not make sense for it,
```

BUILD.cmake.md
Youw: 1) This is a default behavior of CMake, not something invented by HIDAPI. To be honest, it is not documented by CMake, but it is true for all latest toolchains by my observation (I've been chacking this since probably macOS 10.11 or so);

2) Think about it: minimum supported platform by XCode is something like OS X 10.2. It misses half the API used by most applications today; not the most convenient default =)

BUILD.cmake.md
Youw: Good point. But I still want to leave Ninja example as a _second_ example.

BUILD.cmake.md
Youw: > What is not implemented by the hidraw backend?

https://github.com/libusb/hidapi/blob/7620bc7faac7613ff8ea9efbbd54b1a0d719cf76/linux/hid.c#L1069
https://github.com/libusb/hidapi/blob/7620bc7faac7613ff8ea9efbbd54b1a0d719cf76/linux/hid.c#L1108

> I am doubting libusb should be the default backend 

I had a hard time making that decision too. I guess I'm still open for discussion on this matter.

BUILD.cmake.md
Be-ing: okay

BUILD.cmake.md
Be-ing: Is it still true that the hidraw API does have the features needed for hidapi to implement these APIs for the hidraw backend? Maybe they have been implemented since the hidraw backend was written.

mcuee: #259 is one issue tracker to implement hid_get_input_report() for new kernel versions. Not so sure about hid_get_indexed_string().

mcuee: Still I will vote to set hidraw backend as the default for Linux as it is the way to go for most use cases down the road.

Youw: updated

BUILD.cmake.md
```diff
@@ -28,27 +28,46 @@ On macOS CMake may be installed by Homebrew/MacPorts or using the installer from
...
 ```sh
 cd <build dir>
 # configure the build
-cmake -GNinja <HIDAPI source dir>
+cmake <HIDAPI source dir>
 # build it!
+cmake --build .
+# install library; by default installs into /usr/local/
+cmake --build . --target install
+# NOTE: you need to run install command as root, to be able to install into /usr/local/
```
Be-ing: This would use the source tree as the build directory.
```
cmake -S . -B build
```
creates the directory `build` if it does not exist

```
cmake --build build
cmake --install build
```

Youw: > This would use the source tree as the build directory.

Of course it wouldn't. It will create a build tree under _current_ directory, which by `cd <build dir>` is any directory on the filesystem.

> cmake -S . -B build

This creates a build directory _inside the source tree_, which by my experience is almost as bad as using source tree for build tree.<br>
I'm using a software that scans all files under the source tree to index symbols (automatically) and if the build directory is under source tree - it scans it too, during the build as the build goes... So I prefer (and always recommend) to move build directorry somewhere outside the source tree completely.

PR #224, opened 2020-12-13
Windows: ensure buffer passed to HidD_SetFeature is large enough

Mornix:
`HidD_SetFeature()` requires a feature report be at least a certain size similar to the requirement for `WriteFile()` in the implementation of `hid_write()`.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -146,6 +146,7 @@ struct hid_device_ {
 		USHORT output_report_length;
 		size_t input_report_length;
 		USHORT feature_report_length;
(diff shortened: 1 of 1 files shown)


Youw:
Thanks, this is good.
Do you mind making a small change?
Cache the buffer as a device member, and allocate it only on first use.
An optimisation, so we don't do dynamica allocations on each report.

I wanted to apply same optimisation for hid_write to, just didn't have a chance to do so yet.

Youw (approved):
Nice!
This is exactly how I'd do it.

PR #227, opened 2020-12-29
Don't install targets/build tests when build as non-master project

jm4R:
@Youw this is the common solution for using CMake as both master and non-master projects. I've noticed your comment to change this

```
add_subdirectory(${hidapi_SOURCE_DIR} ${hidapi_BINARY_DIR})
```

into this:

```
add_subdirectory(${hidapi_SOURCE_DIR}/src ${hidapi_BINARY_DIR})
```

when using the hidapi with cmake's FetchContent, but this change may reduce user's confusion so you may accept it if you want.


Youw:
> is the common solution for using CMake as both master and non-master projects

I like the idea, even though I've never seen it used by any project so far (maybe you could point me to some examples?)

I think I'll make it even simpler:

```cmake
if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
    add_subdirectory(src)
    return()
endif()
```

jm4R:
@Youw 
> maybe you could point me to some examples?

1) {fmt}
https://github.com/fmtlib/fmt/blob/master/CMakeLists.txt#L10

2) Catch2
https://github.com/catchorg/Catch2/blob/devel/CMakeLists.txt#L6

3) GSL
https://github.com/microsoft/GSL/blob/master/CMakeLists.txt#L23

Youw:
Resolved by: https://github.com/libusb/hidapi/commit/81291b205b52013019c2498620b6d27f2ee3bcc9

Be-ing:
I suggest leaving issues open until the pull request that fixes them is actually merged. GitHub will automatically close the issue when a commit containing the `Fixes: #issuenumber` in the commit description is merged.

Youw:
> until the pull request that fixes them is actually merged

That's what I did. 81291b2 is commited into the same branch this PR is intended to be merged to.

> GitHub will automatically close

Apparently, not in this particular case.

Be-ing:
GitHub closes the issue automatically when the fix is merged. Simply adding a commit to a PR will not automatically close an issue.

Youw:
Right, so I closed it manually.
This PR is about an issue in `cmake-support` branch. The issue is fixed by a direct commit into that branch. The original issue nor this PR is no longer stands.

Issue #234, opened 2021-01-20
Windows version doesn't enumerate devices when exactly one is connected


jm4R:
I don't know if this is coincidence, but this is my case:

I plug UPS HID device (CyberPower or APC) as an only HID device in the whole system, and `hid_enumerate` returns `nullptr`.

When I plug USB keyboard as a second device, `hid_enumerate` enumerates both - keyboard and UPS.

On Linux and exactly the same machine it works fine.

Youw:
I assume you're using build 0.10.1.

Can you step-trace a "single device" scenario with a debugger and tell what do you get from [SetupDiGetClassDevsA](https://github.com/libusb/hidapi/blob/master/windows/hid.c#L330)/[SetupDiEnumDeviceInterfaces](https://github.com/libusb/hidapi/blob/master/windows/hid.c#L339) (or after those)?

jm4R:
driver_name is "Battery" which is weird as in windows device manager the device class shows "HIDClass" as expected.

Youw:
And when you have two devices connected?

todbot:
Yes, I believe on Windows, with the HID Battery driver, that removes it from the list of devices hidapi can access.  (as what happens to gamepads) http://batcmd.com/windows/10/services/hidbatt/

todbot:
I just tested an APC Back-UPS ES 550 on Windows 10 Pro. It appears as both a HID device and as a "Battery" to device driver (see screenshot below).  And it appears to hidapi's `hid_enumerate` just fine (see `hidapitester` output below).  Perhaps you have an additional, non-Microsoft driver on the device?

<img width="1042" alt="Screen Shot 2021-01-20 at 2 21 44p" src="https://user-images.githubusercontent.com/274093/105248936-22795e00-5b2c-11eb-82f5-98b3fba24aba.png">
<img width="877" alt="Screen Shot 2021-01-20 at 2 33 22p" src="https://user-images.githubusercontent.com/274093/105249128-74ba7f00-5b2c-11eb-9c08-3b99c2f5cbdc.png">

Youw:
I think the problem comes if you'd have only this UPS connected, and no other HID devices in the system.

todbot:
Yes I understand that better now. :)  Trying to figure out how to test this since RDP just decided to stop working with a recent Win10 update...

jm4R:
[...]

PR #235, opened 2021-01-21
Windows: Fix devices enumeration

jm4R:
Based on
https://github.com/microsoft/Windows-driver-samples/blob/master/hid/hclient/pnp.c#L32

Fixes wrong detection of the class of devices.

Fixes: #234

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -309,14 +309,16 @@ struct hid_device_info HID_API_EXPORT * HID_API_CALL hid_enumerate(unsigned shor
 	struct hid_device_info *root = NULL; /* return object */
 	struct hid_device_info *cur_dev = NULL;
 
-	/* Windows objects for interacting with the driver. */
+	/* Hard-coded GUID retreived by HidD_GetHidGuid */
 	GUID InterfaceClassGuid = {0x4d1e55b2, 0xf16f, 0x11cf, {0x88, 0xcb, 0x00, 0x11, 0x11, 0x00, 0x00, 0x30} };
+
+	/* Windows objects for interacting with the driver. */
 	SP_DEVINFO_DATA devinfo_data;
 	SP_DEVICE_INTERFACE_DATA device_interface_data;
 	SP_DEVICE_INTERFACE_DETAIL_DATA_A *device_interface_detail_data = NULL;
 	HDEVINFO device_info_set = INVALID_HANDLE_VALUE;
+	char driver_name[256];
 	int device_index = 0;
-	int i;
 
 	if (hid_init() < 0)
 		return NULL;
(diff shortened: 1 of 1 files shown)


windows/hid.c
```diff
@@ -378,32 +381,18 @@ struct hid_device_info HID_API_EXPORT * HID_API_CALL hid_enumerate(unsigned shor
...
-				(strcmp(driver_name, "Keyboard") == 0)) {
-				/* See if there's a driver bound. */
-				res = SetupDiGetDeviceRegistryPropertyA(device_info_set, &devinfo_data,
-				           SPDRP_DRIVER, NULL, (PBYTE)driver_name, sizeof(driver_name), NULL);
-				if (res)
-					break;
-			}
-		}
+		/* Populate devinfo_data. This function will return failure
+		   when there are no more interfaces left. */
```
Youw, line 385: please update or remove this comment, it is outdated

windows/hid.c
```diff
@@ -309,12 +309,15 @@ struct hid_device_info HID_API_EXPORT * HID_API_CALL hid_enumerate(unsigned shor
...
 	GUID InterfaceClassGuid = {0x4d1e55b2, 0xf16f, 0x11cf, {0x88, 0xcb, 0x00, 0x11, 0x11, 0x00, 0x00, 0x30} };
+
+	/* Windows objects for interacting with the driver. */
 	SP_DEVINFO_DATA devinfo_data;
 	SP_DEVICE_INTERFACE_DATA device_interface_data;
 	SP_DEVICE_INTERFACE_DETAIL_DATA_A *device_interface_detail_data = NULL;
 	HDEVINFO device_info_set = INVALID_HANDLE_VALUE;
+	char driver_name[256];
 	int device_index = 0;
 	int i;
```
Youw, line 322: remove it completely?

Youw (approved):
what are the chances we will fail to filter out a non-hid devices?
I understand that HID class GUID should do the trick, but still, this implementation been here like forever, and I'like to give it some thinking, just in case.

windows/hid.c
jm4R: What about "Make sure this device has a driver bound to it" section? This buffer is used there. I am not sure why we check it - HID device without any driver should still work, right?

Youw: yeap, only `i`

jm4R:
> what are the chances we will fail to filter out a non-hid devices?
> I understand that HID class GUID should do the trick, but still, this implementation been here like forever, and I'like to give it some thinking, just in case.

I am not WinApi expert but because this PR is based on the code straight from Microsoft, I assume it should be correct. I also read the documentation of each changed function and indicated that it should work. If the bug I fixed by this PR was there "forever" too (and I am 100% sure it is a bug), I am confused nobody discovered it yet. But maybe some people did - there are a number of open issues in the old repository, that could be caused by this. For example:
https://github.com/signal11/hidapi/issues/290
https://github.com/signal11/hidapi/issues/221

In my use case the old code didn't work (described in #234), but it is my use case :)

PR #260, opened 2021-04-08
Update hid.c

MonkeybreadSoftware:
Put back in HIDIOCSFEATURE define as it's missing for us.

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -45,8 +45,13 @@
 
 #include "hidapi.h"
 
-/* Definitions from linux/hidraw.h. Since these are new, some distros
-   may not have header files which contain them. */
+#ifdef HIDAPI_ALLOW_BUILD_WORKAROUND_KERNEL_2_6_39
+/* This definitions first appeared in Linux Kernel 2.6.39 in linux/hidraw.h.
+    hidapi doesn't support kernels older than that,
+    so we don't define macros below explicitly, to fail builds on old kernels.
+    For those who really need this as a workaround (e.g. to be able to build on old build machines),
+    can workaround by defining the macro above.
+*/
 #ifndef HIDIOCSFEATURE
 #define HIDIOCSFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x06, len)
 #endif
@@ -54,6 +59,7 @@
 #define HIDIOCGFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x07, len)
 #endif
 
+#endif
 
 
 /* USB HID device property names */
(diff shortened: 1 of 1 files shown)


MonkeybreadSoftware:
We build hidapi here for 11 platforms. Lot's of combinations of compilers and platforms, but not all have the HIDIOCSFEATURE define yet.

z3ntu:
You're building for kernels older than 2.6.39? We don't have support for those anymore in hidapi, so even if it builds it probably doesn't work at runtime.

MonkeybreadSoftware:
The build machine is old to avoid some libc dependencies.
The actual machine running the code is usually newer.

And yes, I know we should update it to something newer.

linux/hid.c
```diff
@@ -45,6 +45,16 @@
...
 
+/* Definitions from linux/hidraw.h. Since these are new, some distros
+   may not have header files which contain them. */
+#ifndef HIDIOCSFEATURE
+#define HIDIOCSFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x06, len)
+#endif
+#ifndef HIDIOCGFEATURE
+#define HIDIOCGFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x07, len)
+#endif
+
```
Youw, line 56: Since we dropped support of the old `2.6.39` kernel, I'd rather prefer if hidapi build would fail by default on those.

@MonkeybreadSoftware would this compromise work for you?

```suggestion
#ifdef HIDAPI_ALLOW_BUILD_WORKAROUND_KERNEL_2_6_39
/* This definitions first appeared in Linux Kernel 2.6.39 in linux/hidraw.h.
    hidapi doesn't support kernels older than that,
    so we don't define macros below explicitly, to fail builds on old kernels.
    For those who really need this as a workaround (e.g. to be able to build on old build machines),
    can workaround by defining the macro above.
*/
#ifndef HIDIOCSFEATURE
#define HIDIOCSFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x06, len)
#endif
#ifndef HIDIOCGFEATURE
#define HIDIOCGFEATURE(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x07, len)
#endif

#endif
```

Youw: Of course this would require tweaking you build infrastructure tiny bit.

MonkeybreadSoftware: Sure, it's fine with me. Thank you.

Youw (approved):

MonkeybreadSoftware (commented):
looks good

PR #261, opened 2021-04-09
Use Conditional PlatformToolset projects for VS

jeremyouellet:
This change allows the code to compile without error when used in Visual Studio 2019.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-os: Visual Studio 2015
+os: Visual Studio 2019
 
 environment:
   matrix:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-os: Visual Studio 2019
+os: Visual Studio 2015
 
 environment:
   matrix:
--- a/windows/hidapi.vcxproj
+++ b/windows/hidapi.vcxproj
@@ -26,24 +26,24 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
     <WholeProgramOptimization>true</WholeProgramOptimization>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
     <WholeProgramOptimization>true</WholeProgramOptimization>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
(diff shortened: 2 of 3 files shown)


Youw:
This is the reason why I against supporting .vcproj build for hidapi at all.
While this fixes the build for those who want to build hidapi using v142 toolset with Visual Studio, it breaks the build for those who already using toolset v140 and has it integrated as part of their build system etc.

Strictly speaking, to be able to use Visual Studio 2019, you don't nesessary have to use toolset v142 - you can install toolset v140 for MSVC 2019 and use it in the same way (check Visual Studio Installer -> individual compotents).

---
And generally speaking, I suggest generating MSVC project for your specific Visual Studio/toolset version, using [CMake build](https://github.com/libusb/hidapi/tree/cmake-support).

jeremyouellet:
I understand, that makes sense. Would you like me to close my PR?

Youw:
Unless someone has strong arguments/motivation to vote to accept it, I recommend keeping existing .vcproj files for MSVC2015 forever, and use CMake for all other environments.
Yes, that means this PR likely won't be accepted/can be closed.

Youw:
Thanks for your effort anyway.
I always apreciate clean minical change requests.

jeremyouellet:
What could be done is to have multiple vcxjproj files. One for 2015 and one for 2019. It wouldn't break current setup and would allow people to use VS2019. I've seen this used in other projects before.  I use this as part of a c# project and there is no simple way to integrate cmak into my build process (in a simple manner.) The v140 toolchain is 3.65GB and i find that it seems like a big request that could be solved by adding a couple files.

Though I will respect your choice if you opt to not choose this.

Youw:
> One for 2015 and one for 2019

And in no time we'll have: \<2010, 2012, 2015, 2017, 2019\> \* \<static, shared\> combinations of projects that are essentially identical.
Here is [an example](https://github.com/vslavik/winsparkle/tree/18f63f6b4c51d25a063e79cd8de958eadd5c028d) of what I'd really would like to avoid.
I personally [been there](https://github.com/vslavik/winsparkle/commit/9d3cfdb2f030f4c5dacfa0d2f5f33d044d53919c), when you need to update (and test) all of those, in case of a change, e.g. adding a new file. Super inconvenient.

> I use this as part of a c# project and there is no simple way to integrate cmak into my build process

The idea is that you run a small configuration script _before opening_ your (C#) master solution. That script would generate MSVC solution for hidapi (but won't build it), which could be easily included by your solution.
Yes, this is an extra manual step, which is not really convenient, but nevertheless.

---
Now let's quit lyrics and talk real solutions:
1) Making your own project to build hidapi is as close to trivial as possible. You can always add your own copy of hidapi.vcproj as part of your source tree. It would only include a single source file and a single include path (and a few pretty much standard for Windows link libraries).

2) I did some research on my own, and it appears that .vcproj supports conditional toolset versions, depending on Visual Studio version used.
I must say that today is the first time [I've seen](https://github.com/vslavik/winsparkle/commit/0e86481a919af8f5f70cfc1d9ec09f342f45762e) this is possible. If that is really a working solution, I think it worth a shot.

jeremyouellet:
Damn, I didn't know that, that seems really cool. I made the change, hopefully it passes the test. I don't know if tests need to be changed.

jeremyouellet:
Do you think this is good enough @Youw ?

Youw:
Yes, I just hoped someone else from the team would look as well.
The only thing that bothers me, is that we used to use `v140` and now we use `v140_xp`. I didn't test if that has any impact on the existing projects, where hidapi is used as nested project.

Unless someone explicitly needs it and can properly test, I wouldn't introduce a support of Windows XP. Who develops for it nowadays anyway?

jeremyouellet:
I'll change it, Give me 5 minutes

Youw (approved):

PR #276, opened 2021-05-17
improve documentation for udev rule files on Linux

Be-ing:
Refer to the commit messages for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,11 @@ HIDAPI has five back-ends:
 On Linux, either the hidraw or the libusb back-end can be used. There are
 tradeoffs, and the functionality supported is slightly different.
 
+Note that you will need to install a udev rule file with your application
+for unprivileged users to be able to access HID devices with hidapi. Refer
+to the [69-hid-udev.rules](udev/69-hid-udev.rules) file in the `udev` directory
+for an example.
+
 __Linux/hidraw__ (`linux/hid.c`):
 
 This back-end uses the hidraw interface in the Linux kernel, and supports
--- a/udev/69-hid.rules
+++ b/udev/69-hid.rules
@@ -4,9 +4,7 @@
 
 # If you are using the libusb implementation of hidapi (libusb/hid.c), then
 # use something like the following line, substituting the VID and PID with
-# those of your device. Note that for kernels before 2.6.24, you will need
-# to substitute "usb" with "usb_device". It shouldn't hurt to use two lines
-# (one each way) for compatibility with older systems.
+# those of your device.
 
 # HIDAPI/libusb
 SUBSYSTEM=="usb", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uaccess"
--- a/udev/69-hid.rules
+++ b/udev/69-hid.rules
@@ -17,8 +17,10 @@ SUBSYSTEM=="usb", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uacce
 # HIDAPI/hidraw
 KERNEL=="hidraw*", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uaccess"
 
-# Once done, optionally rename this file for your device, and drop it into
-# /etc/udev/rules.d. Then, replug your device or run:
+# Once done, optionally rename this file for your application, and drop it into
+# /etc/udev/rules.d. Note that these rules must have priorty before 70-uaccess.rules
+# for example, name the file /etc/udev/rules.d/69-my-application-hid.rules.
+# Then, replug your device or run:
 # sudo udevadm control --reload-rules && sudo udevadm trigger
 
 # Note that the hexadecimal values for VID and PID are case sensitive and
--- a/udev/99-hid.rules
+++ b/udev/99-hid.rules
@@ -18,8 +18,8 @@ SUBSYSTEM=="usb", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uacce
 KERNEL=="hidraw*", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uaccess"
 
 # Once done, optionally rename this file for your device, and drop it into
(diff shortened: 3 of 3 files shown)


69-hid-udev.rules
```diff
@@ -0,0 +1,28 @@
...
+# HID devices.
+
+# If you are using the libusb implementation of hidapi (libusb/hid.c), then
+# use something like the following line, substituting the VID and PID with
+# those of your device. Note that for kernels before 2.6.24, you will need
+# to substitute "usb" with "usb_device". It shouldn't hurt to use two lines
+# (one each way) for compatibility with older systems.
+
+# HIDAPI/libusb
+SUBSYSTEM=="usb", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uaccess"
```
Youw, line 12: since this is an example, lets have an alternative example, where GROUP + MODE is used instead of 'uaccess', which is commonly known among community

69-hid-udev.rules
```diff
@@ -0,0 +1,28 @@
...
+
+# If you are using the hidraw implementation (linux/hid.c), then do something
+# like the following, substituting the VID and PID with your device. Busnum 1
+# is USB.
```
Youw, line 16: don't see where `Busnum` is referenced/used

README.md
```diff
@@ -47,6 +47,10 @@ HIDAPI has five back-ends:
 On Linux, either the hidraw or the libusb back-end can be used. There are
 tradeoffs, and the functionality supported is slightly different.
 
+Note that you will need to install a udev rule file with your application
+for unprivileged users to be able to access HID devices with hidapi. Refer
+to the 69-hid-udev.rules file in the root of this repository for an example.
```
Youw, line 52:
```suggestion
to the [69-hid-udev.rules](69-hid-udev.rules) file for an example.
```

Youw (commented):
Lets keep .rules sample unser `udev/` where it was.

69-hid-udev.rules
Be-ing: That is not a good solution because it requires explaining to (human) users that they need to add their user account to a specially named group which may vary depending on the application or distribution. This is an annoying user support burden. I know from experience. `TAG+="uaccess"` does not have this problem.

69-hid-udev.rules
Be-ing: oh, I removed it, but not from the comment. Good catch

69-hid-udev.rules
Youw: I'm not saying this is a bad solution. For some scenarios it is a great solution. But other than that there is no other reasons to limit users from available options.

Apparently you don't need to run unit-tests for your software from CI over SSH, using a real device connected to the build machine, where access to the device is expected exclusively from a remote session.
Having access setup for a specific USER/GROUP is a great solution for that case too.

Be-ing: Does anyone actually do that?

Youw: At least one suck user for your service.

Be-ing: Okay, I added a comment to the end of the file.

Youw (approved):

udev/69-hid.rules
```diff
@@ -0,0 +1,33 @@
...
+SUBSYSTEM=="usb", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uaccess"
+
+# If you are using the hidraw implementation (linux/hid.c), then do something
+# like the following, substituting the VID and PID with your device.
+
+# HIDAPI/hidraw
+KERNEL=="hidraw*", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uaccess"
+
+# Once done, optionally rename this file for your application, and drop it into
+# /etc/udev/rules.d. Note that these rules must have priorty before 70-uaccess.rules
```
tresf, line 19: @Be-ing I hope you do not mind the tag.  Can you help explain the following:

> that these rules must have priorty before `70-uaccess.rules`

Be-ing: I don't recall the reason for this. I'm not sure we ever dug deep to figure out exactly what the issue was, but IIRC we stumbled on this shipping the udev rule file for Mixxx.

Be-ing: If you figure this out, it would be great if you open a pull request to clarify the documentation in this example udev rule file.

tresf: I've been having problems using the `hidraw` in general so I came here more as a "what am I doing wrong" step.  I'll surely improve if I get further. 🍺

Be-ing: Is this for LMMS?

tresf: It's actually for another project (QZ Tray), which uses hidapi through hid4java.

Youw: > Can you help explain the following:
>>that these rules must have priorty before `70-uaccess.rules`

I'm pretty sure it is bacause of the use `TAG+="uaccess"` - this tag is handled by the `70-uaccess.rules` file, and all the rule files are processed in strict order, and if the tag is added _after_ `70-uaccess.rules` is processed - the device will be processed as if this tag is not present at all.

tresf: @Youw thanks for the tip... I stumbled upon this bug report which seems to state something similar. 
 https://github.com/systemd/systemd/issues/4288#issuecomment-348166161

Youw: Aha, so I haven't actually looked up the documentation for `70-uaccess.rules`, I just assumed its purpose based on the documentation as suggested by this PR.
Thanks for pointing out to that comment. It appears my assumption is absolutely correct, but regarding a different file: `73-seat-late.rules`.

I'll update a documentation once again.

Youw: On that note: it looks like [`70-uaccess.rules`](https://cgit.freedesktop.org/systemd/systemd/tree/src/login/70-uaccess.rules) just automatically adds the `uaccess` tag to a few known devices/device types.

Be-ing:
Tangentially related: I requested a means for [sandboxed applications to access HID devices](https://github.com/systemd/systemd/issues/20533) without installing a udev rule file.

PR #277, opened 2021-05-17
README: document how to choose between libusb & hidraw backends

Be-ing:


README.md
```diff
@@ -45,7 +45,9 @@ HIDAPI has five back-ends:
 * Mac (using IOHidManager)
 
 On Linux, either the hidraw or the libusb back-end can be used. There are
-tradeoffs, and the functionality supported is slightly different.
+tradeoffs, and the functionality supported is slightly different. Both are
+built. It is up to the application linking to hidapi to choose the backend
```
Youw, line 49:
```suggestion
built by default. It is up to the application linking to hidapi to choose the backend
```

Youw (approved):

README.md
Be-ing: `./configure --help` does not document any way to choose which backends are built.

Youw: CMake build is on the way, it will be easily configurable there.
And one can always manually patch [configuration file](https://github.com/libusb/hidapi/blob/master/Makefile.am#L19) before the build.
So the alternative is possible, but _by default_ you have booth.

Be-ing:
Thanks for merging.

Issue #278, opened 2021-05-19
MacOS panic when hid_write called with a zero length buffer


JnyJny:
### Software Versions
```console
$ sw_vers
ProductName:	macOS
ProductVersion:	11.3.1
BuildVersion:	20E241

$ python --version
Python 3.9.4

# https://github.com/trezor/cython-hidapi
$ poetry show | grep hid
hidapi               0.10.1    A Cython interface to the hidapi from https://github.com/libusb/hidapi
```

### Reproducer
```python
#!/usr/bin/env python3

import hid

path = hid.enumerate()[0]["path"]
device = hid.device()
device.open_path(path)
device.write(0)
# panic
```

### Discussion

I know this is a case of "don't do that" but I was completely surprised that writing a zero caused a panic (several panics since I did it **a bunch** to confirm the cause of the panic).

I realize this problem begins in `python` but I'm reasonably confident the panic happens in `hidapi`.

Starting at [hid.pyx:device.write](https://github.com/trezor/cython-hidapi/blob/b94c8a09582681aff6edd145a3bfd3948ff7e5ca/hid.pyx#L136), a zero is transliterated into a `bytes` object with zero length and calls [hid_write](https://github.com/libusb/hidapi/blob/c7a2339b0530aa63814aee824c04872e89613a02/mac/hid.c#L961), a wrapper around [set_report](https://github.com/libusb/hidapi/blob/c7a2339b0530aa63814aee824c04872e89613a02/mac/hid.c#L896), which then dereferences (a possibly NULL) buffer pointer. Presuming the buffer pointer is not null,  the buffer pointer and length are then passed to `IOHIDDeviceGetReport`. The panic report (see below) doesn't seem to give much clue whether it panic'ed in `set_report` or `IOHIDDeviceGetReport`.  My _guess_ is `IOHIDDeviceGetReport`. 

### Desired Outcome 
I would prefer to be gently chided with a return value of `-1 ` and `errno=EBADPROGRAMMER` when I call `hid_write` with a zero length buffer. 

### Panic Report 
[Full panic report](https://gist.github.com/JnyJny/7ad2bed77183555b8d1db065ef6b7ad8) 

#### Extracted panicString
```
panic(cpu 0 caller 0xfffffff026ebf448): macOS panic detected
Debugger message: panic
Memory ID: 0x6
OS release type: User
OS version: 18P4556
macOS version: 20E241
[...]

PR #279, opened 2021-05-19
Guard against zero length buffers in hid_write

JnyJny:
This is my proposed solution for avoiding a macOS panic in `set_report` when the user supplies a zero length buffer (indicated by either a NULL data or a length of 0 or less).

Looking back at the source, I see that `length` is of type `size_t` which my foggy brain remembers being unsigned so the test for `less than or equal to zero` should likely be `equal zero`. 

Fixes: #278

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1047,7 +1047,6 @@ int HID_API_EXPORT hid_write(hid_device *dev, const unsigned char *data, size_t 
 	int skipped_report_id = 0;
 
 	if (!data || (length ==0)) {
-		errno = EINVAL;
 		return -1;
 	}
 
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1043,8 +1043,15 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
 int HID_API_EXPORT hid_write(hid_device *dev, const unsigned char *data, size_t length)
 {
 	int res;
-	int report_number = data[0];
+	int report_number;
 	int skipped_report_id = 0;
+
+	if (!data || (length ==0)) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	report_number = data[0];
 
 	if (report_number == 0x0) {
 		data++;
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -963,6 +963,12 @@ int HID_API_EXPORT hid_write(hid_device *dev, const unsigned char *data, size_t 
 {
 	int bytes_written;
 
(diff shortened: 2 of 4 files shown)


Youw (approved):
Since we're adding a check for macOS, I think it is best to add it for all backends in a row.

JnyJny:
Absolutely a good idea. Should I add to this PR or open one for each backend?

Youw:
A single PR is fine.

libusb/hid.c
```diff
@@ -1043,9 +1043,16 @@ hid_device * HID_API_EXPORT hid_open_path(const char *path)
 int HID_API_EXPORT hid_write(hid_device *dev, const unsigned char *data, size_t length)
 {
 	int res;
-	int report_number = data[0];
+	int report_number;
 	int skipped_report_id = 0;
 
+	if (!data || (length ==0)) {
+		errno = EINVAL;
```
Youw, line 1050: hidapi doen't use errno for error reporting, I would't introduce it randomly
Maybe introduce it once and for all functions properly, maybe as a separate PR

JnyJny: I was afraid of that, not a problem.

linux/hid.c
```diff
@@ -963,6 +963,12 @@ int HID_API_EXPORT hid_write(hid_device *dev, const unsigned char *data, size_t
 {
 	int bytes_written;
 
+	if (!data || (length == 0)) {
+		errno = EINVAL;
```
Youw, line 967: I guess hidraw is using system calls a lot and if any of those fails, errno will be set, so having it here, even if "randomly" should be fine

Youw (approved):

linux/hid.c
JnyJny: I was trying to "blend in" with the rest of the source using `register_device_error` and this looked out of place:

```c
        register_device_error(dev, strerror(EINVAL));
```

I'm ok with whatever is most in the style of the source, not trying cause more work or make waves :)

Youw: Thanks for your help

JnyJny: Glad to help, I've been a consumer of this project for some time and I'm happy to contribute back to it in some small way.

JnyJny:
Just checking in to make sure I haven't forgotten anything that needs doing for this PR.

Youw:
Looks good as far as I can tell.
Out team agreed to keep PRs for a few days just so everyone can catch up before the merge.

mcuee:
Nice one.

PR #296, opened 2021-07-03
CI: more CI for CMake

Youw:
- Linux CMake build;
- Windows MSVC build;
- Windows NMake build;
- Windows MinGW build;


Be-ing:
This could be simplified using GH Actions' matrix feature. For example: https://github.com/mixxxdj/libkeyfinder/blob/main/.github/workflows/build-and-test.yml

Youw:
I've looked into it, and I don't find it "simpler" than it is right now.
Each of the configurations has different parameters list or even different stages. Maybe it will be _less lines_ of code in total, but definitely _not simpler_.
Right now one can read (and use it as an example) how to build HIDAPI on different platforms/with different toolchains.
Having a lot of `if`s on each step makes it way harder to read.

PR #280, opened 2021-05-20
add libusb-specific hid_libusb_wrap_sys_device

Youw:
Rationale: on Android one must use UsbManager, to access any
USB device. As a result, libraries like libusb can only use file descriptors
that are provided by UsbManager.
libusb has an API to use such file descriptors: hid_libusb_wrap_sys_device.
Having hid_libusb_wrap_sys_device currently is the only way to make hidapi
work on Android without root access and without custom Android builds.

Relevant info: https://github.com/libusb/libusb/pull/830/files

PR #297, opened 2021-07-04
Remove redundant build systems

Be-ing:
There is no need for these anymore. CMake can autogenerate Visual Studio project files if desired, but that's not needed anymore either since Visual Studio now supports CMake.


Be-ing:
This is required to ensure that the CMake target files are installed consistently by packagers.

Youw:
I've heard your opinion, thank you.

There're project who can't use CMake simply because they aren't using it right now (maybe depending on a vcproj as a dependent project), and introducing the dependency for them achieves nothing.

There is no need to force everyone rewrite their build scripts just because "oh look, there is a better way".

Be-ing:
Well, there is. CMake projects can't rely on `find_package` working if Linux distro packagers choose to use autotools. Can we at least remove autotools?

Be-ing:
> introducing the dependency

Visual Studio now includes CMake, so this is not an issue on Windows.

Youw:
> Visual Studio now includes CMake

Yeah, tell that to MSVC 2015 users. And don't ask "why can't they just upgrade?".

> if Linux distro packagers choose to use autotools

Right now they _all_ use autotools. If we remove autotools right now, it will mean one of two outcomes:

1) Everyone will drop everything they do right now, and will reqrite their build scripts for HIDAPI to use CMake instead of Autotools;
2) Each distro will schedule, postone or delay the upgrade of HIDAPI library; for some time they will still use `0.10.1`, until someone will be generous-unough to upgrade the build script to use CMake;

Can you estimate the chances of the two?

Be-ing:
Actually, yes, I can from experience. Packagers don't have much a problem switching to CMake. @dvzrv updated the Arch Linux package for Mixxx for the recent release switching to CMake in just two days after the release.

However, downstream software will have a problem if packagers can keep using autotools because packagers will likely follow the path of least resistance and not install the CMake target files.

dvzrv:
@Youw Thanks for looking into the ins and outs of this pull request and @Be-ing thanks for mentioning.

From a distribution packaging perspective it is a bit of a "depends" situation I guess whether packagers choose one over the other build system.
While some people might have extensive experience with autotools and stick with it because of that, it is unfortunately also a fact, that autotools had a [development hiatus of about eight years](https://git.savannah.gnu.org/gitweb/?p=autoconf.git) and only recently has gotten back to even merging outstanding fixes again upstream. For me personally, the latter is reason enough to rather stick with cmake (if quality permits) in cases where both build systems are offered.

I fully understand, that you want to maintain backwards compatibility with e.g. MSVC 2015.
A good way to do so (and you can of course take that advice or leave it ;-) ) is to e.g. add a list of supported environments (the current [build instructions for Windows does not offer this](https://github.com/libusb/hidapi/blob/master/BUILD.md#windows)), as [MSVC seems to have clear EOL targets](https://en.wikipedia.org/wiki/Microsoft_Visual_Studio#History).
That way it is easier to define what is supported and also becomes more easy to calculate e.g. EOL for autotools, in case you choose to not want to maintain two build systems in the future anymore.

Youw:
> rather stick with cmake (if quality permits) in cases where both build systems are offered

I totally support such approach.

---
My current plan is to keep autotools for at least few next releases, maybe even until v1.0.0.

I keep hearing "support both build systems", but it is not like I (or anyone else from the community) needs to spend extra efforts all the time to keep it working or up-to-date with the source tree.
Yes, there're been a few patches from the community to fix something here and there, but I don't see any burden of keeping it as is for those who choose to use it, specially those who are already using it.

Also, having Autotools around doesn't automatically means, that it will be a 100% up-to-date with all the new features, and there is a few differences:
1) CMake doesn't support building FOX-based UI test app (and it won't be); even though it is deprecated, people still may make an attempt to use it, and Autotools is the only option to do so;
2) I have future plans to make a small Qt-based UI test app (instead of FOX-based) and that one will only be supported by CMake;
3) Even [this](https://github.com/libusb/hidapi/pull/280) change (and all further similar features, which I'm sure will be introduced) is going to be introduced only for CMake build;

---
And we can still work with distribution/package maintainers to have them switched to CMake for HIDAPI even if Autotools is still available.
I'll make an explicit note in the release notes for v0.11.0, that people should consider CMake over Autotools if they are able to.

Be-ing:
>  I'll make an explicit note in the release notes for v0.11.0, that people should consider CMake over Autotools if they are able to.

Can we call the old build systems official deprecated and recommend using CMake instead?

Youw:
I'll deprecate it in master once 0.11.0 is released.

Be-ing:
That seems like a good plan.

Issue #282, opened 2021-05-28
Need new backend for BLE under Windows


LeeChunHei:
I am using the python wrapper of this library, [pyhidapi](https://github.com/apmorton/pyhidapi)
when I check the product_string of the Bluetooth HID device, it returns None.
I noticed that in windows this library is using [HidD_GetProductString](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/hidsdi/nf-hidsdi-hidd_getproductstring) to get the name of the HID device.
So I asked in Microsoft community and they [replied](https://docs.microsoft.com/en-us/answers/questions/401236/hidd-getproductstring-with-ble-hid-device.html) this function will not work on BLE device but will work on the classic Bluetooth.
However, it still not working after I change to use Bluetooth classic HID device.
Therefore, I would like to ask if anyone successfully gets the Bluetooth classic HID Device product string with Windows10 before.
Or I just messed up with some other stuff.

Youw:
I think this may be relevant to you:
https://stackoverflow.com/questions/65970214/where-does-bluetooth-and-other-devices-get-its-information-from#comment116656875_65970214

The other thing that comes to my mind: is your device a Mouse-/Keyboard-like device? In other words - does hid_open_path falls-back to no read/write access (`open_device(path, FALSE)`) ?

LeeChunHei:
> The other thing that comes to my mind: is your device a Mouse-/Keyboard-like device? In other words - does hid_open_path falls-back to no read/write access (`open_device(path, FALSE)`) ?

My device is not the mouse-/keyboard-like device, I set the report descriptor to a custom report, and when I try to read write the device, the python program did not warn me with no read/write access.

> I think this may be relevant to you:
> https://stackoverflow.com/questions/65970214/where-does-bluetooth-and-other-devices-get-its-information-from#comment116656875_65970214

This one is interesting, did the question mean that the product name can be seen like this
![image](https://user-images.githubusercontent.com/24974375/119983209-6dcbdd80-bff2-11eb-9355-9de367b50a73.png)
But showing generic name in hid device like this
![image](https://user-images.githubusercontent.com/24974375/119983307-8d630600-bff2-11eb-9c9b-18ba5916e1da.png)
Then HidD_GetProductString is provided in Win32API instead of WinRT?
And it is possible that in that question, a BLE device is used which the Microsoft guy replied HidD_GetProductString not supported.

[...]

PR #301, opened 2021-07-07
Fix possible garbage in device strings returned from hid_enumerate()

DJm00n:
In some cases HidD_GetSerialNumberString/HidD_GetManufacturerString/HidD_GetProductString can fail.
For example if we have Bluetooth LE gamepad on Windows.
In this case we are currently returning not initialized strings in `hid_device_info`.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -207,8 +207,8 @@ static void register_error(hid_device *dev, const char *op)
 	   end of the message. Thanks Microsoft! */
 	ptr = msg;
 	while (*ptr) {
-		if (*ptr == '\r') {
-			*ptr = 0x0000;
+		if (*ptr == L'\r') {
+			*ptr = L'\0';
 			break;
 		}
 		ptr++;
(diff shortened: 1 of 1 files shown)


DJm00n:
Related to #282

Youw (commented):
I know this is very debatable, but I'm not a big fan of `NULL` for C-strings values.
There is a lot of code like [this](https://github.com/libusb/hidapi/blob/f6d0073fcddbdda24549199445e844971d3c9cef/hidtest/test.c#L61) where values are passed to functions like `printf` (without a check) which requires a valid value. (Even C++ `std::string` doesn't accept NULL C-string).

So instead of fallback to `NULL` I'd suggest a fallback to `_wcsdup(L"")`

windows/hid.c
```diff
@@ -463,28 +463,22 @@ struct hid_device_info HID_API_EXPORT * HID_API_CALL hid_enumerate(unsigned shor
...
 			/* Serial Number */
-			wstr[0]= 0x0000;
+			wstr[0]= L'\0';
 			res = HidD_GetSerialNumberString(write_handle, wstr, sizeof(wstr));
-			wstr[WSTR_LEN-1] = 0x0000;
-			if (res) {
-				cur_dev->serial_number = _wcsdup(wstr);
-			}
+			wstr[WSTR_LEN-1] = L'\0';
+			cur_dev->serial_number = _wcsdup(wstr);
```
Youw, line 469: I don't think I follow the reasoning for this change.
So now, if `HidD_GetSerialNumberString` we would get a NULL-terminated but still garbage, aren't we?

DJm00n: `HidD_GetSerialNumberString` doesn't touch buffer if there was error.
So we'll receive null terminated empty string.

Youw (approved):

DJm00n:
Hi @Youw!
Could you please run build system workflow?

PR #302, opened 2021-07-07
Whitespace cleanup in Windows implementation

DJm00n:
For better diffs in future.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -8,7 +8,7 @@
  8/22/2009
 
  Copyright 2009, All Rights Reserved.
- 
+
  At the discretion of the user of this library,
  this software may be licensed under the terms of the
  GNU General Public License v3, a BSD-Style license, or the
@@ -153,7 +153,7 @@ struct hid_device_ {
 		BOOL read_pending;
 		char *read_buf;
 		OVERLAPPED ol;
-		OVERLAPPED write_ol;			  
+		OVERLAPPED write_ol;
 };
 
 static hid_device *new_hid_device()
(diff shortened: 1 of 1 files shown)


hidapi/hidapi.h
```diff
@@ -30,11 +30,11 @@
 #include <wchar.h>
 
 #ifdef _WIN32
-      #define HID_API_EXPORT __declspec(dllexport)
-      #define HID_API_CALL
+	  #define HID_API_EXPORT __declspec(dllexport)
```
Youw, line 33: you've replaced some of the spaces with tabs, but not all of them

consistency (even across small block of code) is better than "some general rules"

windows/hid.c
```diff
@@ -417,7 +417,7 @@ struct hid_device_info HID_API_EXPORT * HID_API_CALL hid_enumerate(unsigned shor
 		/* Check the VID/PID to see if we should add this
 		   device to the enumeration list. */
 		if ((vendor_id == 0x0 || attrib.VendorID == vendor_id) &&
-		    (product_id == 0x0 || attrib.ProductID == product_id)) {
+			(product_id == 0x0 || attrib.ProductID == product_id)) {
```
Youw, line 420: IMHO "before" is better then "after"
1) Now part the condition is aligned with the code, which a con from readability POV
2) Before, the start of the condition in each line was aligned which is proc from readability POV

windows/hid.c
```diff
@@ -552,12 +552,12 @@ HID_API_EXPORT hid_device * HID_API_CALL hid_open(unsigned short vendor_id, unsi
...
 	const char *path_to_open = NULL;
 	hid_device *handle = NULL;
-	
+
 	devs = hid_enumerate(vendor_id, product_id);
 	cur_dev = devs;
 	while (cur_dev) {
 		if (cur_dev->vendor_id == vendor_id &&
-		    cur_dev->product_id == product_id) {
+			cur_dev->product_id == product_id) {
```
Youw, line 560: ditto

Youw (commented):
Please see related discussion:
https://github.com/libusb/hidapi/pull/215

windows/hid.c
DJm00n: Alright. Anyways current state when we have mixed tabs and spaces is a mess.

Can we stick to spaces-only (I prefer that) instead?

Youw: historically HIDAPI is a tabs-only (except a few exceptions)

DJm00n:
@Youw I have removed spaces-to-tabs conversion. Now removing extra spaces only.

Youw (approved):

Issue #289, opened 2021-06-24
MSYS2 MinGW-w64 compiler warning about strncpy


mcuee:
There is a warning message generated under MSYS2.

```
C:\work\hid\hidapi\windows\hid.c: In function 'hid_enumerate':
C:\work\hid\hidapi\windows\hid.c:459:5: warning: 'strncpy' specified bound depends on the length of the source argument [-Wstringop-overflow=]
  459 |     strncpy(cur_dev->path, str, len+1);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\work\hid\hidapi\windows\hid.c:457:11: note: length computed here
  457 |     len = strlen(str);
      |           ^~~~~~~~~~~

$ gcc -v
Using built-in specs.
COLLECT_GCC=C:\msys64\mingw64\bin\gcc.exe
COLLECT_LTO_WRAPPER=C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/lto-wrapper.exe
Target: x86_64-w64-mingw32
[...]

PR #311, opened 2021-07-25
Update hid.c to fix MSYS2 MinGW-w64 compiler warning

mcuee:
Fix issue #289 that there is a compiler warning when using MSYS2 MinGW-w64 compiler. 
```
[ 25%] Building C object src/windows/CMakeFiles/hidapi_winapi.dir/hid.c.obj
In function 'hid_get_device_info',
    inlined from 'hid_enumerate' at C:\work\hid\hidapi\windows\hid.c:517:34:
C:\work\hid\hidapi\windows\hid.c:333:3: warning: 'strncpy' specified bound depends on the length of the source argument [-Wstringop-overflow=]
  333 |   strncpy(dev->path, path, len + 1);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\work\hid\hidapi\windows\hid.c: In function 'hid_enumerate':
C:\work\hid\hidapi\windows\hid.c:331:16: note: length computed here
  331 |   size_t len = strlen(path);
      |                ^~~~~~~~~~~~

$ gcc -v
Using built-in specs.
COLLECT_GCC=C:\msys64\mingw64\bin\gcc.exe
COLLECT_LTO_WRAPPER=C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/lto-wrapper.exe
Target: x86_64-w64-mingw32
Configured with: ../gcc-10.3.0/configure --prefix=/mingw64 --with-local-prefix=/mingw64/local --build=x86_64-w64-mingw32 --host=x86_64-w64-mingw32 --target=x86_64-w64-mingw32 --with-native-system-header-dir=/mingw64/x86_64-w64-mingw32/include --libexecdir=/mingw64/lib --enable-bootstrap --enable-checking=release --with-arch=x86-64 --with-tune=generic --enable-languages=c,lto,c++,fortran,ada,objc,obj-c++,jit --enable-shared --enable-static --enable-libatomic --enable-threads=posix --enable-graphite --enable-fully-dynamic-string --enable-libstdcxx-filesystem-ts=yes --enable-libstdcxx-time=yes --disable-libstdcxx-pch --disable-libstdcxx-debug --enable-lto --enable-libgomp --disable-multilib --disable-rpath --disable-win32-registry --disable-nls --disable-werror --disable-symvers --with-libiconv --with-system-zlib --with-gmp=/mingw64 --with-mpfr=/mingw64 --with-mpc=/mingw64 --with-isl=/mingw64 --with-pkgversion='Rev1, Built by MSYS2 project' --with-bugurl=https://github.com/msys2/MINGW-packages/issues --with-gnu-as --with-gnu-ld --with-boot-ldflags='-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high -Wl,--disable-dynamicbase -static-libstdc++ -static-libgcc' 'LDFLAGS_FOR_TARGET=-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high' --enable-linker-plugin-flags='LDFLAGS=-static-libstdc++\ -static-libgcc\ -pipe\ -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high\ -Wl,--stack,12582912'
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 10.3.0 (Rev1, Built by MSYS2 project)
```

Signed-off-by: Xiaofan Chen xiaofanc@gmail.com

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -63,17 +63,12 @@ extern "C" {
 
 #include <stdio.h>
 #include <stdlib.h>
-
+#include <string.h>
 
 #include "hidapi.h"
 
 #undef MIN
 #define MIN(x,y) ((x) < (y)? (x): (y))
-
-#ifdef _MSC_VER
-	/* Thanks Microsoft, but I know how to use strncpy(). */
-	#pragma warning(disable:4996)
-#endif
 
 #ifdef __cplusplus
 extern "C" {
@@ -330,8 +325,7 @@ static struct hid_device_info *hid_get_device_info(const char *path, HANDLE hand
 	if (path) {
 		size_t len = strlen(path);
 		dev->path = (char*)calloc(len + 1, sizeof(char));
-		strncpy(dev->path, path, len + 1);
-		dev->path[len] = '\0';
+		memcpy(dev->path, path, len + 1);
 	}
 	else
 		dev->path = NULL;


Youw (approved):

windows/hid.c
```diff
@@ -330,7 +330,7 @@ static struct hid_device_info *hid_get_device_info(const char *path, HANDLE hand
 	if (path) {
 		size_t len = strlen(path);
 		dev->path = (char*)calloc(len + 1, sizeof(char));
-		strncpy(dev->path, path, len + 1);
+		strcpy(dev->path, path);
 		dev->path[len] = '\0';
```
Youw, line 334: an alternative:
```suggestion
		memcpy(dev->path, path, len + 1);
```
but I don't believe it makes any practical difference

Youw (approved):

windows/hid.c
mcuee: Actually this is better. It will also remove the MSVC warning 4996 against strcpy or strncpy.

```
warning C4996: 'strcpy': This function or variable may be unsafe. Consider using strcpy_s instead. To disable deprecation, 
use _CRT_SECURE_NO_WARNINGS. See online help for details.
```

mcuee: Just pushed new commits which change to memcpy and remove the MSVC pragma to disable warning 4996.

Youw:
By C standard, `memcpy` is defined in `string.h`
Lets include it explicitly

mcuee:
> By C standard, `memcpy` is defined in `string.h`
> Lets include it explicitly

Done.

mcuee:
I probably messed up when I try to squash 4 commits into one (done correctly but then messed up). Please just look at here.
https://github.com/mcuee/hidapi/commit/904853d0d4ce539e133360388622d4636b6b0ccc

Youw (approved):

tormodvolden:
Maybe the compiler was right, since you are not checking the return from calloc(). Isn't strdup() what you want here?

Youw:
The compiler wasn't complaining about the return value of calloc(), but rather about 3rd argument of strncpy.
And you're right, that we better check for calloc() return.
And you're right - it is basically strdup().
The only reason why I didn't suggest using strdup() in a first place - I only suggest using a standard C functions, when you aren't 100% sure what compiler are you going to use.

mcuee:
> As a result of a successful squash you'd get a single commit in this PR.
> Apparently, since there is 6 commits - something gone sideways.
> 
> NOTE: if you had a PR, and then you've desided to squash all commits, you're basically rewriting your branch, it becomes incompatible with the version you've already pushed. In such case, in order to update the PR, you'd need to `push --force` your local squashed branch, that is the only way.
> 
> Also, it is not required most of the time.
> Github itself allows "Squash and merge" a PR, so the master history will get a single commit.

Thanks. This is very helpful. I will need to learn more about git. Previously I only use very simple functions like clone, pull, commit.

mcuee:
@tormodvolden It is not so good to use strdup() which will trigger the warning like the original strncpy().
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strdup-wcsdup?view=msvc-160

>The Microsoft-implemented POSIX function names strdup and wcsdup are deprecated aliases for the _strdup and _wcsdup functions. By default, they generate Compiler warning (level 3) C4996. The names are deprecated because they don't follow the Standard C rules for implementation-specific names. However, the functions are still supported.
>
>We recommend you use _strdup and _wcsdup instead. Or, you can continue to use these function names, and disable the warning. For more information, see Turn off the warning and POSIX function names.

tormodvolden:
OK, I agree. Better not introduce posixisms if they were not already in use.

But why not use strcpy()? Would the compiler also complain then?

Youw:
What would be the practival difference?

mcuee:
@tormodvolden Both strcpy and strncpy will trigger the warning 4996. Using strcpy was actually my initial fix (after using google). But after @Youw mentioned that memcpy can be an alternative (even though he mentioned that both were okay), I found out that we can remove the following MSVC lines. I think it is always to remove this kind of compiler specific stuff whenever we can.

```
#ifdef _MSC_VER
	/* Thanks Microsoft, but I know how to use strncpy(). */
	#pragma warning(disable:4996)
#endif

```

mcuee:
> And you're right, that we better check for calloc() return.

Ah yes that is a good point. I did not notice this, just trying to use this as a chance to learn how to create a pull request (still very bad at git). I am not a programmer and my focus is more on the testing and support side for the projects I am involved in.

Issue #310, opened 2021-07-22
Add support for errors in hid_error on MacOS


hardkrash:
The `IOreturn` value can be passed to the c library function `mach_error_string` to provide the error string for the error that was raised.

`mach_error_string` is defined in [mach_error.h](https://opensource.apple.com/source/xnu/xnu-2422.1.72/libsyscall/mach/mach/mach_error.h.auto.html)

`char *mach_error_string(mach_error_t error_value);`  and a `IOReturn` is a version of `mach_error_t`

mcuee:
@hardkrash Please help to create a pull request. Thanks.

Youw:
Oh, this is done as part of https://github.com/libusb/hidapi/pull/314.
Available in the latest release.

PR #314, opened 2021-07-31
Add error reporting to the MacOS port.

hardkrash:
Initial implementation of error messages for the macOS port.

Interface mostly copied from the linux port.
Messages probably need some love for the other errors being reported.


hardkrash:
This is to resolve issue #310

mac/hid.c
```diff
@@ -193,13 +196,84 @@ static struct hid_api_version api_version = {
 static	IOHIDManagerRef hid_mgr = 0x0;
 static	int is_macos_10_10_or_greater = 0;
 
+/* Global error message that is not specific to a device, e.g. for
+   hid_open(). It is thread-local like errno. */
+__thread wchar_t *last_global_error_str = NULL;
```
Youw, line 201: this `__thread` actually bugs me very much:
we claim that it is [thread safe](https://github.com/libusb/hidapi/blob/8741697359b8dd0d73702a69c54e64ffc552d3cf/hidapi/hidapi.h#L461), but imagine next scenario:

thread1: `hid_init()` - success
thread1: `hid_enumerat()` - error happened and registered
thread2: `hid_enumerat()` - error happened and registered
thread1: `hid_exit()` - `last_global_error_str` from thread1 is freed, but `last_global_error_str` from thread2 is leaked

I'm sure the very same issue is present with `hidraw` implementation of the global error

---
your intention with this PR is great, but I'd like to hear some ideas of how can we fix the global issues all together

hardkrash: I am not a threading expert, and only copied code from the other implementation from this code base.
Why is thread 2 allowed to run `hid_enumerate()` without calling `hid_init()`

How do we add `register_global_error(NULL);` to a `pthread_cleanup_push`?
But it looks like that's the solution.  In thread we need the cleanup_pop to have ability to clear the hid thread error.

It also looks like I might have missed adding the freeing of the device pointers on close of device

Youw: > Why is thread 2 allowed to run `hid_enumerate()` without calling `hid_init()`

That is the design of the API. `hid_init()` - is a _global_ initialisation of the library.

> How do we add `register_global_error(NULL)`; to a `pthread_cleanup_push`?
But it looks like that's the solution. In thread we need the cleanup_pop to have ability to clear the hid thread error.

`pthread_cleanup_push` might not be enough, as it clears the resources at thread exit only, and we should come up with something at `hid_exit()`.

I think we need to use [`pthread_key`](https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_key_create.html) here, unless someone has other ideas.

> It also looks like I might have missed adding the freeing of the device pointers on close of device

Self-code-review is always a good thing.

Youw: After going thru the git history again, and trying to find a good solution,
I think the best approach would be to discard the faulty behavior introduced not so long ago (https://github.com/libusb/hidapi/pull/23) and don't make any thread-safety assumptions for `hid_error`, i.e.: make `last_global_error_str` a simple `static` variable.

Youw:
Hi @hardkrash,

Are you still up to finish this PR?
I've started some changes for `hidraw` backend: https://github.com/libusb/hidapi/pull/428,
those introduce a safe strategy for `hid_error` and allows to proceed with the implementation for other (this one) backends.

mcuee:
#428 has been merged.

Youw (approved):
I took the liberty of updating the implementation:
- rebased to master/fixed conflicts;
- updated error messages/error handling a bit;

I don't have a chance to test this well as of yet.
Additional core review is welcome.

Youw:
I guess if any objections/improvements - we'll be fixing them in master.

PR #326, opened 2021-08-23
udev.rules documentation improvement

Youw:
As recently discovered (thanks @tresf), the `TAG+="uaccess"` has to be used
before `73-seat-late.rules` and not before `70-uaccess.rules`.

Technically it doesn't harm, to have a more restrictive recommenration,
but it was not technically currect.

Refference on `uaccess` TAG consumption:
[1](https://github.com/systemd/systemd/blob/3a1220eedfb65355c387c29584380742a21917aa/src/login/73-seat-late.rules.in), [2](https://www.apt-browse.org/browse/debian/wheezy/main/amd64/systemd/44-11+deb7u4/file/lib/udev/rules.d/73-seat-late.rules).

Relates to: #276, systemd/systemd/issues/4288.

### DIFF
--- a/udev/69-hid.rules
+++ b/udev/69-hid.rules
@@ -16,7 +16,7 @@ SUBSYSTEM=="usb", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uacce
 KERNEL=="hidraw*", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uaccess"
 
 # Once done, optionally rename this file for your application, and drop it into
-# /etc/udev/rules.d.
+# /etc/udev/rules.d/.
 # NOTE: these rules must have priorty before 73-seat-late.rules.
 # (Small discussion/explanation in systemd repo:
 #  https://github.com/systemd/systemd/issues/4288#issuecomment-348166161)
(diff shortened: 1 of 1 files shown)


udev/69-hid.rules
```diff
@@ -16,17 +16,20 @@ SUBSYSTEM=="usb", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uacce
 KERNEL=="hidraw*", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="003f", TAG+="uaccess"
 
 # Once done, optionally rename this file for your application, and drop it into
-# /etc/udev/rules.d. Note that these rules must have priorty before 70-uaccess.rules
-# for example, name the file /etc/udev/rules.d/69-my-application-hid.rules.
+# /etc/udev/rules.d/.
+# NOTE: these rules must have priorty before 73-seat-late.rules.
+# (Small discussion/explanation in systemd repo:
+#  https://github.com/systemd/systemd/issues/4288#issuecomment-348166161)
+# for example, name the file /etc/udev/rules.d/70-my-application-hid.rules.
```
Be-ing, line 23:
```suggestion
# for example, name the file /etc/udev/rules.d/72-my-application-hid.rules.
```

Be-ing:
How about renaming the file to `72-hid.rules`?

Be-ing:
I tested changing the priority of Mixxx's udev rule file and confirmed this is correct.

udev/69-hid.rules
Youw: > How about renaming the file to 72-hid.rules?

So there is a standard set of rules that adds a `uaccess` TAG (`70-uaccess.rules`) with priority `70`.
I'm thinking maybe having and example with the same priority would be nit.

Qbicz (approved):
Looks good to me!

Issue #49, opened 2019-07-19
Non-free files? And to remove DDK build scripts.


z3ntu:
Debian had code to remove `windows/ddk_build/makefile` and `testgui/mac_support.cpp` from the tarball in their `debian/rules` file

`windows/ddk_build/makefile`:
```
 Copyright (C) Microsoft Corporation 1995, 1996
       All Rights Reserved.
```

`testgui/mac_support.cpp`:
```
Some of this code is from Apple Documentation, most notably
 http://developer.apple.com/legacy/mac/library/documentation/AppleScript/Conceptual/AppleEvents/AppleEvents.pdf 
```
and the pdf (archived at http://web.archive.org/web/20141226100754/http://developer.apple.com/legacy/library/documentation/AppleScript/Conceptual/AppleEvents/AppleEvents.pdf) has a footer `Copyright © 2005, 2007 Apple Inc. All Rights Reserved.`

z3ntu:
I've tried removing `testgui/mac_support.cpp` and building the test gui on macOS and it still works fine. So I don't think this file is actually used. After looking around a bit, it seems that this file is unused since https://github.com/libusb/hidapi/commit/bab22b7ac1f2d97290209910fcb6f2cea68ea972
I'll send a PR.

Qbicz:
Good catch, thank you!

Qbicz:
@z3ntu did we merge it already? Is it https://github.com/libusb/hidapi/commit/a1d49824ea6f8d5fcc90734b11bd934067eb730a? If yes, we can close the issue.

z3ntu:
The `windows/ddk_build/makefile` part is still relevant I think

mcuee:
Actualy the whole ddk_build directory can be removed now. MinGW and Visual Studio are enough.

FYI: libusb has already removed the ddk_build legacy files.

mcuee:
With Windows 7 out of support by Microsoft, I think it does not make sense to keep the ddk build files which is only valid for Windows 7 DDK. New Windows WDK are totally different (based on msbuild).

mcuee:
With CMake added, I think we should remove this legacy Windows ddk build scripts.

mcuee:
It does not make sense to keep the Windows DDK build scripts now. In fact, it does not work any way.

```
c:\WinDDK\7600.16385.1\bin>setenv.bat c:\WinDDK\7600.16385.1 free x64 WIN7
[...]

PR #333, opened 2021-09-18
Remove DDK support files.

Youw:
Deleted DDK build scripts, support files and documentation for it,
as it wasn't working of-the-box with recent tool-sets from MS anyway.

Still keeping `HIDAPI_USE_DDK` macro in the code,
as it still _can_ be used if the build configuration is setup properly.

Resolves: #49


mcuee (approved):

mcuee:
Looks good to me.

Issue #328, opened 2021-08-30
Missing one byte feature report read under Windows for devices without report id


mcuee:
Reference here:
https://github.com/libusb/hidapi/issues/285

Now I start to question the commit https://github.com/libusb/hidapi/commit/6fcb0bb2282dbb505561a53689eba0c0536570d9 which I proposed in #232 with regard to feature report. It seems to be correct when the report id is not zero. However, it seems to report one less byte when there is no report id. It seems to me for Windows, we may need to report one more byte for reading of feature report when the report id is zero.

mcuee:
https://github.com/libusb/hidapi/issues/285#issuecomment-907570047

Just tried hidapitester with hidapi-0.10.1 release and hidapi git head, with regard to input report, the result is the same since it is using hid_read_timeout(). So #232 does not matter in that case.
https://github.com/todbot/hidapitester/blob/master/hidapitester.c#L379

With regard to feature report, then the result will be different as it uses hid_get_feature_report(). So commit https://github.com/libusb/hidapi/commit/6fcb0bb2282dbb505561a53689eba0c0536570d9 does matter for getting the feature report. In this particular case without report id (report id zero), it seems to me the commit is not correct.
https://github.com/todbot/hidapitester/blob/master/hidapitester.c#L403

```
$ ./hidapitester_release0101.exe --vidpid=0925:7001 --open --send-feature 0,0x40,0x41 --read-feature 0
Opening device, vid/pid: 0x0925/0x7001
Writing 64-byte feature report...wrote 64 bytes:
 00 40 41 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Reading 64-byte feature report, report_id 0...read 3 bytes:
 00 40 41 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Closing device

$ ./hidapitester_git4d63a50.exe --vidpid=0925:7001 --open --send-feature 0,0x40,0x41 --read-feature 0
Opening device, vid/pid: 0x0925/0x7001
[...]

Issue #285, opened 2021-06-15
RFC: to include hidapitester as it is a very useful tool for debugging


mcuee:
hidtest is using a test device based on modified Microchip Simple HID demo. It can be kept as an example but not that useful as a test tool for the project.

hidapitester is an excellent tool for testing of HID devices and HIDAPI. It will be good to include it in HIDAPI.
https://github.com/todbot/hidapitester

mcuee:
Only issue to include hidapitester is probably the GPL 3 license, which may or may not be a problem. I do not see it as a problem. But if some people see it as a problem, we can ask @todbot to see if he is willing to change the license.

mcuee:
@todbot How do you like the idea of including hidapitester in HIDAPI project?

todbot:
Sounds great to me

todbot:
I can change the license to whatever is most useful

mcuee:
HIDAPI license:
https://github.com/libusb/hidapi/blob/master/LICENSE.txt

hidapitester is already licensed under GPL3, just need to add the other two licenses.

mcuee:
Just FYI on the original hidtest, it is based on Microchip Demo FW (part of Microchip USB Stack) but I do not know if Alan Ott published the modified FW or not.

With the original FW (04d8:003f), here is the output.
```
Device Found
  type: 04d8 003f
  path: \\?\hid#vid_04d8&pid_003f#e&24ba9f4d&0&0000#{4d1e55b2-f16f-11cf-88cb-001111000030}
  serial_number: (null)
  Manufacturer: Microchip Technology Inc.
  Product:      Simple HID Device Demo
  Release:      2
  Interface:    -1
  Usage (page): 0x1 (0xff00)

Manufacturer String: Microchip Technology Inc.
Product String: Simple HID Device Demo
Unable to read serial number string
Serial Number String: (0)
Indexed String 1: Microchip Technology Inc.
Unable to send a feature report.
Unable to get a feature report.
Incorrect function.waiting...
Data read:
   81 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Press any key to continue . . .
```

mcuee:
[...]

PR #334, opened 2021-09-22
Windows: hid_get_feature/input_report return size fix

Youw:
- It appears when the numbered reports aren't used,
WinAPI returns size of the report excluding the data[0] which contains 0,
as an indication that numbered reports aren't used;
Explicity count that byte in the result.

Fixes: #328

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,8 @@ Makefile
 stamp-h1
 libtool
 
+# macOS
 .DS_Store
+
+# Qt Creator
+CMakeLists.txt.user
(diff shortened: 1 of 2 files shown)


mcuee (approved):

mcuee:
As mentioned in #328, the fixes for the feature report should be correct.

mcuee:
However, hidapitester can not be used to test the fix to the input report as it uses hid_read_timeout().
https://github.com/libusb/hidapi/issues/285#issuecomment-907570047

PR #337, opened 2021-09-24
Update 69-hid.rules

stolk:
SUBSYSTEMS instead of SUBSYSTEM

I tested this on my Ubuntu 21.04 system, and when using the SUBSYSTEM keyword, I would not get a match. But when I switched to SUBSYSTEMS (prompted by a Stackoverflow comment) it immediately started working.


Youw:
Thanks!

PR #348, opened 2021-11-10
windows: add missing include for mingw32

selsta:
Seems to be introduced in https://github.com/libusb/hidapi/pull/309/

```
2021-11-10T00:31:43.7352733Z hid.c:164:83: error: unknown type name ‘DEVPROPKEY’
2021-11-10T00:31:43.7355003Z   typedef CONFIGRET(__stdcall* CM_Get_DevNode_PropertyW_)(DEVINST dnDevInst, CONST DEVPROPKEY* PropertyKey, DEVPROPTYPE* PropertyType, PBYTE PropertyBuffer, PULONG PropertyBufferSize, ULONG ulFlags);
```

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -33,6 +33,7 @@ typedef LONG NTSTATUS;
 #endif
 
 #ifdef __MINGW32__
+#include <devpropdef.h>
 #include <ntdef.h>
 #include <winbase.h>
 #endif


Youw:
Could you confirm that the header name as provided by MinGW is exactlty `devpropdef.h`? And not `Devpropdef.h` (First letter caseness, it is upper-case in Windows SDK)?
Otherwise it will fail during cross-compilation from Linux.

selsta:
```
lrwxrwxrwx  1 root root    42 Nov 15  2019 devpropdef.h -> ../../share/mingw-w64/include/devpropdef.h
```

Seems to be `devpropdef.h`, though I have seen the capitalized spelling here: https://docs.microsoft.com/en-us/previous-versions/ff543546(v=vs.85)

> Otherwise , if it doesn't match, it will fail during cross-compilation from Linux.

Cross compiling from Linux is how I triggered this issue in the first place (https://github.com/monero-project/monero/pull/8055), with the patch applied it compiled fine.

Youw (approved):
> Seems to be devpropdef.h, though I have seen the capitalized spelling here

This is exactly why I've asked to double-check.

Thanks.

Issue #259, opened 2021-04-07
hidraw: hid_get_input_report implementation for Kernels that support it


Youw:
As pointed out by the community, Linux kernel now supports get_input_report operation for hidraw backend (#258).
Reference: https://github.com/torvalds/linux/commit/f43d3870cafa2a0f3854c1819c8385733db8f9ae

A little effort is required to use that by hidapi.

PR #351, opened 2021-11-14
implement hid_get_input_report for Linux hidraw backend

Be-ing:
Fixes https://github.com/libusb/hidapi/issues/259

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -62,6 +62,13 @@
 #endif
 
 
+// HIDIOCGINPUT is not defined in Linux kernel headers < 5.11.
+// This definition is from hidraw.h in Linux >= 5.11.
+// https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f43d3870cafa2a0f3854c1819c8385733db8f9ae
+#ifndef HIDIOCGINPUT
+#define HIDIOCGINPUT(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x0A, len)
+#endif
+
 /* USB HID device property names */
 const char *device_string_names[] = {
 	"manufacturer",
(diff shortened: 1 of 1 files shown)


Be-ing:
I don't know if this actually works with Linux < 5.11, but at least it should build.

Be-ing:
Looking closer at the [commit that implemented this in Linux](https://github.com/torvalds/linux/commit/f43d3870cafa2a0f3854c1819c8385733db8f9ae#), simply defining the macro won't do anything in Linux < 5.11 without the other changes in that commit. So I changed this to keep the old dummy code that just returns an error for old versions of Linux.

JoergAtGithub:
The ifdef checks the Kernel at compile time and returns the error if it was compiled on a too old Kernel.
But what happens if the executable is compiled on a new kernel, but executed on an old one?

Be-ing:
I have not tested that.

Be-ing:
I suppose that could actually happen with Flatpak... I'll test what happens with a VM running an old kernel.

Youw:
Lets have the functionality compile regardless of the kernel version (e.g. if not defined HIDIOCGINPUT - define it);
As fo running on the older kernel - I'm pretty sure the call will simply fail, since the correspondind ioctl isn't supported.

Be-ing:
Okay I switched it back to the original implementation copying the definition from hidraw.h.

linux/hid.c
```diff
@@ -1065,13 +1065,21 @@ int HID_API_EXPORT hid_get_feature_report(hid_device *dev, unsigned char *data,
 	return res;
 }
 
-// Not supported by Linux HidRaw yet
+// HIDIOCGINPUT is not defined in Linux kernel headers < 5.11.
+// This definition is from hidraw.h in Linux >= 5.11.
+// https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f43d3870cafa2a0f3854c1819c8385733db8f9ae
+#ifndef HIDIOCGINPUT
+#define HIDIOCGINPUT(len)    _IOC(_IOC_WRITE|_IOC_READ, 'H', 0x0A, len)
+#endif
```
Youw, line 1073: move it to line 64 of this file

Youw (approved):
This looks good.
I'd like to check it locally with my device(s) befor having it merged.

Be-ing:
Have you had a chance to test this yet?

Youw (approved):
I finally had.
Everything looks perfect. Thanks!

Be-ing:
Thanks. Do you have a plan when you will make the next release?

Youw:
The major (minor actually) release is not going to happen soon - nothing really done for that.
But I'll try to make a patch release soon.

Be-ing:
A patch release soon with this fix would be good.

Be-ing:
Any further updates about a patch release?

Be-ing:
Thanks! I made a [PR to update vcpkg](https://github.com/microsoft/vcpkg/pull/22304).

Be-ing:
I also made a [PR to update the Rust bindings](https://github.com/ruabmbua/hidapi-rs/pull/75).

PR #358, opened 2021-12-18
HIDRAW hid_get_indexed_string implementation

JoergAtGithub:
This is untested code, coded offline without a physical Linux system.

The easiest way to test it, should be to read the first strings, where every HID device stores Manufacturer and ProductID.

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -39,6 +39,7 @@
 
 /* Linux */
 #include <linux/hidraw.h>
+#include <linux/hiddev.h>
 #include <linux/version.h>
 #include <linux/input.h>
 #include <libudev.h>
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -1114,13 +1114,21 @@ int HID_API_EXPORT_CALL hid_get_serial_number_string(hid_device *dev, wchar_t *s
 	return get_device_string(dev, DEVICE_STRING_SERIAL, string, maxlen);
 }
 
-int HID_API_EXPORT_CALL hid_get_indexed_string(hid_device *dev, int string_index, wchar_t *string, size_t maxlen)
+int HID_API_EXPORT_CALL hid_get_indexed_string(hid_device * dev, int string_index, wchar_t * string, size_t maxlen)
 {
-	(void)dev;
-	(void)string_index;
-	(void)string;
-	(void)maxlen;
-	return -1;
+
+	int res;
+	struct hiddev_string_descriptor str_desc;
+
+	str_desc.index = string_index;
+	res = ioctl(dev->device_handle, HIDIOCGSTRING, str_desc);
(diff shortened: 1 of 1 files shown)


linux/hid.c
```diff
@@ -1114,13 +1115,21 @@ int HID_API_EXPORT_CALL hid_get_serial_number_string(hid_device *dev, wchar_t *s
 	return get_device_string(dev, DEVICE_STRING_SERIAL, string, maxlen);
 }
 
-int HID_API_EXPORT_CALL hid_get_indexed_string(hid_device *dev, int string_index, wchar_t *string, size_t maxlen)
+int HID_API_EXPORT_CALL hid_get_indexed_string(hid_device * dev, int string_index, wchar_t * string, size_t maxlen)
```
Youw, line 1118: I understand that we don't have a strict code-style document,
and we won't debate about the "west-const vs east-const" holy-war here.

But a **consistency** across implementation is a must-have - please revert this line.

Youw (commented):
I will be able to check this one.

But cannot promice when though. As usual =(

Youw:
This cannot possibly work by definition.

Tha handle we have (`dev->device_handle`) is a [`hidraw`](https://dri.freedesktop.org/docs/drm/hid/hidraw.html) device, which is handled by a hidraw kernel module. The devices are reported as `/dev/hidraw[0-31?]` (not sure about the upper limit, but that's not the point).

[`hiddev`](https://dri.freedesktop.org/docs/drm/hid/hiddev.html#the-hiddev-api) on the other hand - is entirely different subsystem. The devices are the `/dev/usb/hiddev[0..31?]`.

Because different subsystems are handling those devices, you can literally find different function to handle ioctl requests for those, i.e.:
- hidraw: [hidraw_ioctl](https://elixir.bootlin.com/linux/latest/source/drivers/hid/hidraw.c#L363);
- hiddev: [hiddev_ioctl](https://elixir.bootlin.com/linux/latest/source/drivers/hid/usbhid/hiddev.c#L594);

The complete list of supported IOCTLs for hidraw subsystem can be found, e.g. here: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/hidraw.h

---
`hid_get_indexed_string` cannot possibly be implemented for `hidraw` backend simply because it is not supported by kernel.

We only recently [added implementation](https://github.com/libusb/hidapi/commit/d67b5c9fec5bd16da7e88f337386ec1ef48ac4f2) for `hid_get_input_report` for hidraw backend because the support for input reports by the kernel was added [not so long ago](https://github.com/torvalds/linux/commit/f43d3870cafa2a0f3854c1819c8385733db8f9ae).

JoergAtGithub:
Thanks for investigating and explaining!

PR #360, opened 2021-12-20
Parse USB interface number from device Hardware IDs.

DJm00n:
Strickly speaking we cannot parse interface string at all and only allowed to parse Hardware ID string.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -412,10 +412,35 @@ static void hid_internal_get_ble_info(struct hid_device_info* dev, DEVINST dev_n
 	}
 }
 
+/* USB Device Interface Number.
+   It can be parsed out of the Hardware ID if a USB device is has multiple interfaces (composite device).
+   See https://docs.microsoft.com/windows-hardware/drivers/hid/hidclass-hardware-ids-for-top-level-collections
+   and https://docs.microsoft.com/windows-hardware/drivers/install/standard-usb-identifiers
+
+   hardware_id is always expected to be uppercase.
+*/
+static int hid_internal_get_interface_number(const wchar_t* hardware_id)
+{
+	int interface_number;
+	wchar_t *startptr, *endptr;
+	const wchar_t *interface_token = L"&MI_";
+
+	startptr = wcsstr(hardware_id, interface_token);
+	if (!startptr)
+		return -1;
+
+	startptr += wcslen(interface_token);
+	interface_number = wcstol(startptr, &endptr, 16);
+	if (endptr == startptr)
+		return -1;
+
+	return interface_number;
+}
+
 static void hid_internal_get_info(struct hid_device_info* dev)
 {
 	const char *tmp = NULL;
-	wchar_t *interface_path = NULL, *device_id = NULL, *compatible_ids = NULL;
+	wchar_t *interface_path = NULL, *device_id = NULL, *compatible_ids = NULL, *hardware_ids = NULL;
 	mbstate_t state;
 	ULONG len;
 	CONFIGRET cr;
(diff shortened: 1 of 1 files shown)


windows/hid.c
```diff
@@ -455,6 +479,27 @@ static void hid_internal_get_info(struct hid_device_info* dev)
...
 
+	/* Get the hardware ids from devnode */
+	len = 0;
+	cr = CM_Get_DevNode_PropertyW(dev_node, &DEVPKEY_Device_HardwareIds, &property_type, NULL, &len, 0);
+	if (cr == CR_BUFFER_SMALL && property_type == DEVPROP_TYPE_STRING_LIST) {
+		hardware_ids = (wchar_t*)calloc(len, sizeof(BYTE));
+		cr = CM_Get_DevNode_PropertyW(dev_node, &DEVPKEY_Device_HardwareIds, &property_type, (PBYTE)hardware_ids, &len, 0);
+	}
+	if (cr != CR_SUCCESS)
+		goto end;
```
Youw, line 490: are we sure this is the best decision?
in case if we fail to read `HardwareId` why don't we just skip reading the `interface_number` and get to the next property

windows/hid.c
```diff
@@ -412,17 +412,41 @@ static void hid_internal_get_ble_info(struct hid_device_info* dev, DEVINST dev_n
 	}
 }
 
+/* USB Device Interface Number.
+   It can be parsed out of the Hardware ID if a USB device is has multiple interfaces (composite device).
+   See https://docs.microsoft.com/windows-hardware/drivers/hid/hidclass-hardware-ids-for-top-level-collections
+   and https://docs.microsoft.com/windows-hardware/drivers/install/standard-usb-identifiers
+*/
```
Youw, line 419:
```suggestion

    hardware_id is always expected to be uppercase.
*/
```

windows/hid.c
```diff
@@ -693,7 +720,7 @@ struct hid_device_info HID_API_EXPORT * HID_API_CALL hid_enumerate(unsigned shor
 		/* Check the VID/PID to see if we should add this
 		   device to the enumeration list. */
 		if ((vendor_id == 0x0 || attrib.VendorID == vendor_id) &&
-		    (product_id == 0x0 || attrib.ProductID == product_id)) {
+			(product_id == 0x0 || attrib.ProductID == product_id)) {
```
Youw, line 723: the alignment "before" is intentional, please revert

windows/hid.c
```diff
@@ -754,7 +781,7 @@ HID_API_EXPORT hid_device * HID_API_CALL hid_open(unsigned short vendor_id, unsi
 	cur_dev = devs;
 	while (cur_dev) {
 		if (cur_dev->vendor_id == vendor_id &&
-		    cur_dev->product_id == product_id) {
+			cur_dev->product_id == product_id) {
```
Youw, line 784: revert

windows/hid.c
DJm00n: `DEVPKEY_Device_HardwareIds` is mandatory property of any device node in Windows device tree. I think this check should be there just for consistency.

windows/hid.c
DJm00n: fixed

windows/hid.c
DJm00n: fixed

windows/hid.c
```diff
@@ -412,17 +412,43 @@ static void hid_internal_get_ble_info(struct hid_device_info* dev, DEVINST dev_n
...
+   See https://docs.microsoft.com/windows-hardware/drivers/hid/hidclass-hardware-ids-for-top-level-collections
+   and https://docs.microsoft.com/windows-hardware/drivers/install/standard-usb-identifiers
+
+   hardware_id is always expected to be uppercase.
+*/
+static int hid_internal_get_interface_number(const wchar_t* hardware_id)
+{
+	int interface_number;
+	wchar_t *startptr, *endptr;
+	const wchar_t *interface_token = L"MI_";
```
Youw, line 426:
```suggestion
	const wchar_t *interface_token = L"&MI_";
```

Youw (approved):

windows/hid.c
DJm00n: fixed

windows/hid.c
Youw: ok, at least it is consistent with the other parts in the same function

PR #362, opened 2021-12-23
Ported code from SetupApi to CfgMgr32

DJm00n:
- Ported code from SetupApi to CfgMgr32. See: https://docs.microsoft.com/windows-hardware/drivers/install/porting-from-setupapi-to-cfgmgr32
- Use Unicode Windows API internally where possible. Windows operates natively in UTF-16 (WCHAR) so do we too.

### DIFF
--- a/windows/CMakeLists.txt
+++ b/windows/CMakeLists.txt
@@ -4,7 +4,6 @@ add_library(hidapi_winapi
 )
 target_link_libraries(hidapi_winapi
     PUBLIC hidapi_include
-    PRIVATE setupapi
 )
 
 set_target_properties(hidapi_winapi
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -21,7 +21,7 @@
 ********************************************************/
 
 #if defined(_MSC_VER) && !defined(_CRT_SECURE_NO_WARNINGS)
-// Do not warn about mbsrtowcs, wcsrtombs and wcsncpy usage.
+// Do not warn about wcsncpy usage.
 // https://docs.microsoft.com/cpp/c-runtime-library/security-features-in-the-crt
 #define _CRT_SECURE_NO_WARNINGS
 #endif
@@ -35,6 +35,7 @@ typedef LONG NTSTATUS;
 #ifdef __MINGW32__
 #include <ntdef.h>
 #include <winbase.h>
+#define WC_ERR_INVALID_CHARS 0x00000080
 #endif
 
 #ifdef __CYGWIN__
@@ -324,13 +325,13 @@ static int lookup_functions()
 }
 #endif
 
-static HANDLE open_device(const char *path, BOOL open_rw)
+static HANDLE open_device(const wchar_t *path, BOOL open_rw)
 {
 	HANDLE handle;
 	DWORD desired_access = (open_rw)? (GENERIC_WRITE | GENERIC_READ): 0;
 	DWORD share_mode = FILE_SHARE_READ|FILE_SHARE_WRITE;
 
-	handle = CreateFileA(path,
+	handle = CreateFileW(path,
 		desired_access,
 		share_mode,
 		NULL,
@@ -447,11 +448,9 @@ static int hid_internal_get_interface_number(const wchar_t* hardware_id)
 	return interface_number;
 }
 
-static void hid_internal_get_info(struct hid_device_info* dev)
+static void hid_internal_get_info(const wchar_t* interface_path, struct hid_device_info* dev)
 {
-	const char *tmp = NULL;
-	wchar_t *interface_path = NULL, *device_id = NULL, *compatible_ids = NULL, *hardware_ids = NULL;
-	mbstate_t state;
+	wchar_t *device_id = NULL, *compatible_ids = NULL, *hardware_ids = NULL;
 	ULONG len;
 	CONFIGRET cr;
 	DEVPROPTYPE property_type;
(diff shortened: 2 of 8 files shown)


windows/hid.c
```diff
@@ -605,113 +622,69 @@ static struct hid_device_info *hid_get_device_info(const char *path, HANDLE hand
...
-			   SPDRP_DRIVER, NULL, (PBYTE)driver_name, sizeof(driver_name), NULL);
-		if (!res)
-			goto cont;
+	/* Iterate over each device interface in the HID class, looking for the right one. */
+	for (wchar_t* device_interface = device_interface_list; *device_interface; device_interface += wcslen(device_interface) + 1) {
+		HANDLE read_handle = INVALID_HANDLE_VALUE;
+		HIDD_ATTRIBUTES attrib;
 
-		//wprintf(L"HandleName: %s\n", device_interface_detail_data->DevicePath);
+		//wprintf(L"Device Path: %s\n", device_interface);
```
Youw, line 678: Sinse we are refactoring - lets get rid of this comment

Youw (commented):
Once you've removed dependency on SetupAPI - we should remove link dependency on it too: from CMakeLists, autotools and vsproj files

Youw:
This is nice refactoring, thanks.

windows/hid.c
```diff
@@ -164,30 +140,47 @@ static struct hid_api_version api_version = {
...
+	size_t input_report_length;
+	USHORT feature_report_length;
+	unsigned char *feature_buf;
+	void *last_error_str;
+	DWORD last_error_num;
+	BOOL read_pending;
+	char *read_buf;
+	OVERLAPPED ol;
+	OVERLAPPED write_ol;
+	struct hid_device_info* device_info;
```
Youw, line 183: None of it is relevant to the topic of the PR.
Please revert all the white-space changes and shuffling around of the order of declarations.

It makes _unnecessary noise in the git history_, which I often need to consult.

windows/hid.c
```diff
@@ -21,74 +21,48 @@
...
-	#define IOCTL_HID_GET_FEATURE                   HID_OUT_CTL_CODE(100)
-	#define IOCTL_HID_GET_INPUT_REPORT              HID_OUT_CTL_CODE(104)
-
-#ifdef __cplusplus
-} /* extern "C" */
-#endif
-
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
```
Youw, line 73: whese are removed, not moved - is that mean we don't use any of `stdlib.h`/`string.h` anymore?

windows/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ add_library(hidapi_winapi
 )
 target_link_libraries(hidapi_winapi
     PUBLIC hidapi_include
-    PRIVATE setupapi
+    PRIVATE cfgmgr32 hid
```
Youw, line 7: And with this change you will break the build for everywone who don't have WDK installed (which is a large amount of users).

To have a support of DDK builds, it has to be done properly, e.g. a CMake option, and link against `hid.lib` only when the user explicitly needs to.

Youw (commented):
Remove all of the link dependencies on `cfgmgr32` and `hid` from all of build scripts.

windows/CMakeLists.txt
Youw: Just to be clear: DDK support is out of scope of this PR.
And it would require a lot more than I've mentioned in the previous comment.

windows/hid.c
DJm00n: alright

windows/hid.c
DJm00n: it seems. or these are included by some other headers. not sure how to check.

windows/CMakeLists.txt
DJm00n: Alright. I'll do another PR for that.

windows/hid.c
Youw: > or these are included by some other headers

That is a time bomb. I've had a failed builds after a compiler/library upgrade, because some of the headers where implicitly included where they shouldn't.

windows/hid.c
```diff
@@ -588,144 +589,96 @@ static struct hid_device_info *hid_get_device_info(const char *path, HANDLE hand
...
+		cr = CM_Get_Device_Interface_ListW(&interface_class_guid, NULL, device_interface_list, len, CM_GET_DEVICE_INTERFACE_LIST_PRESENT);
+	} while (cr == CR_BUFFER_SMALL);
 
-		/* Make sure this device has a driver bound to it. */
-		res = SetupDiGetDeviceRegistryPropertyA(device_info_set, &devinfo_data,
-			   SPDRP_DRIVER, NULL, (PBYTE)driver_name, sizeof(driver_name), NULL);
-		if (!res)
-			goto cont;
+	if (cr != CR_SUCCESS) {
+		return NULL;
```
Youw, line 654: `device_interface_list` leak here

windows/hid.c
```diff
@@ -588,144 +589,96 @@ static struct hid_device_info *hid_get_device_info(const char *path, HANDLE hand
...
+			free(device_interface_list);
 		}
 
-		/* Populate devinfo_data. This function will return failure
-		   when the device with such index doesn't exist. We've already checked it does. */
-		res = SetupDiEnumDeviceInfo(device_info_set, device_index, &devinfo_data);
-		if (!res)
-			goto cont;
-
+		device_interface_list = (wchar_t*)calloc(len, sizeof(wchar_t));
```
Youw, line 649: check for non-`NULL` return

Youw (commented):
except for the potential leak - looks good

Youw (approved):

Sunbreak:
A little curious why porting SetupApi to CfgMgr32?

> UWP will continue to be supported with "bug, reliability, and security fixes", but Windows App SDK is the "single path forward"

https://github.com/flutter/flutter/issues/14967#issuecomment-1033337989

PR #365, opened 2021-12-30
libusb: add hid_get_report_descriptor implementation

Youw:
- based on LIBUSB_REQUEST_GET_DESCRIPTOR/LIBUSB_DT_REPORT;
- improved logging of LIBUSB errors;

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -324,6 +324,16 @@ static int get_usage(uint8_t *report_descriptor, size_t size,
 }
 #endif /* INVASIVE_GET_USAGE */
 
+static int hid_get_report_descriptor_libusb(libusb_device_handle *dev, int interface_num, unsigned char *buf, size_t buf_size)
+{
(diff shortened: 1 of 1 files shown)


Youw:
I've tested this one with a few devices that I have, and it appears to be technically correct, but not usefull very much.

For instance: hidraw (#364) implementation gives consistent results with what macOS implementation gives (already in get-descriptor branch), i.e. the descriptors match perfectly.
But libusb implementation (current PR) either doesn't give the descriptor at all (libusb transfer fails wit IO error), or the returned descriptor is not the same as reported by hidraw.

I assume it is because of luck of the support of `LIBUSB_REQUEST_GET_DESCRIPTOR/LIBUSB_DT_REPORT` on device side.
Or maybe the problem is that my device is a Mouse/Keyboard device, and libusb is locked by the kernel for those reports.

I assume that this could be fixed by reading raw usb desciptors from sysfs and getting the HID report descriptor from it by manually parsing it (I've done something similar in the past), but it will take some implementation effort to do so and I can't say I have spare time to get my hands to it nor I require this functionality in any of my projects.

---
TL;DR;: this is not perfect, but based on a correct USB/HID specification and works for devices that supports it.
Better solution is possible if/when someone needs and implements it.

JoergAtGithub:
I googled this similar case: https://community.nxp.com/t5/i-MX-Processors/Libusb-get-report-descriptor-error-on-imx6dl/m-p/1065766#M156631

PR #369, opened 2022-01-22
Reorganize Windows source structure

Youw:


Youw:
This is mostly preparation for further changes in [windows_report_descriptor_reconstructor_unit_test](https://github.com/libusb/hidapi/tree/windows_report_descriptor_reconstructor_unit_test) branch.

Youw:
@DJm00n your commen on this PR would be nice to get.

DJm00n:
`hidsdi.h`/`hidpi.h`/`hidclass.h` AFAIK comes with usual Windows SDK since Vista (2006!).
`cfgmgr32.h` was in there since Windows 2000.
https://en.wikipedia.org/wiki/Windows_Driver_Kit

I don't think we should support ancient Windows versions by 2022.
Also some cfgmgr32 methods (that we are using now after refactor) was added in Vista.

Youw:
someday I'll comeup with HIDAPI 1.0 - with a bit different API (e.g. explicit contexts like libusb has, etc.) - then I'll remove all of that ancient  garbage.

But let's have it as is for now. Specifally when it is separated into its own headers and the hid.c implementation become much cleaner.

PR #378, opened 2022-02-08
Add Mac-specific API to get device location ID

triplef:
As discussed in https://github.com/libusb/hidapi/discussions/252#discussioncomment-2089707.

Not sure if there’s a simpler way to do this, but as I need the location ID during enumeration it seems I first need to create a `IOHIDDeviceRef` from the path in order to get the location ID.

### DIFF
--- a/mac/hid.c
+++ b/mac/hid.c
@@ -34,7 +34,7 @@
 #include <unistd.h>
 #include <dlfcn.h>
 
-#include "hidapi.h"
+#include "hidapi_darwin.h"
 
 /* As defined in AppKit.h, but we don't need the entire AppKit for a single constant. */
 extern const double NSAppKitVersionNumber;
@@ -1186,6 +1186,17 @@ int HID_API_EXPORT_CALL hid_get_indexed_string(hid_device *dev, int string_index
 	/* TODO: */
 
 	return 0;
+}
+
+int HID_API_EXPORT_CALL hid_darwin_get_location_id(hid_device *dev, uint32_t *location_id)
+{
+	int res = get_int_property(dev->device_handle, CFSTR(kIOHIDLocationIDKey));
+	if (res != 0) {
+		*location_id = (uint32_t) res;
+		return 0;
+	} else {
+		return -1;
+	}
 }
 
 
--- a/mac/hidapi_darwin.h
+++ b/mac/hidapi_darwin.h
@@ -1,0 +1,50 @@
+/*******************************************************
+ HIDAPI - Multi-Platform library for
+ communication with HID devices.
+
+ libusb/hidapi Team
+
+ Copyright 2022, All Rights Reserved.
+
+ At the discretion of the user of this library,
+ this software may be licensed under the terms of the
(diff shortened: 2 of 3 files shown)


mac/Makefile.am
```diff
@@ -4,6 +4,6 @@ libhidapi_la_LDFLAGS = $(LTLDFLAGS)
 AM_CPPFLAGS = -I$(top_srcdir)/hidapi/
 
 hdrdir = $(includedir)/hidapi
-hdr_HEADERS = $(top_srcdir)/hidapi/hidapi.h
+hdr_HEADERS = $(top_srcdir)/hidapi/hidapi.h hidapi_darwin.h
```
Youw, line 7: revert

autotools are deprecated now
we don't want to develope/support new feature for it - that will be a good motivation to migrate to CMake for projects that are still using it autotools and are actively developed

mac/hidapi_darwin.h
```diff
@@ -0,0 +1,49 @@
...
+
+		/** @brief Get the location ID for a HID device.
+
+			@ingroup API
+			@param path The path name of the device
+
+			@returns
+				This function returns the location ID, or 0 on error.
+		*/
+		unsigned HID_API_EXPORT_CALL hid_darwin_get_location_id(const char *path);
```
Youw, line 43: make it a:

`int HID_API_EXPORT_CALL hid_darwin_get_location_id(hid_device *dev, uint32_t *location_id);`

returns `-1` on failure - like all other function

---
and if someone wants to use it with `path`: use `hid_open_path` -> `hid_darwin_get_location_id` -> `hid_close`

Youw (commented):
Thanks!

Just a few comments

triplef:
Looks like the AppVeyor build failed spuriously:
> Error uploading artifact to the storage: The request was aborted: Could not create SSL/TLS secure channel.

Youw:
That is not an issue for us

mac/hid.c
```diff
@@ -1176,6 +1176,17 @@ int HID_API_EXPORT_CALL hid_get_serial_number_string(hid_device *dev, wchar_t *s
 	return get_serial_number(dev->device_handle, string, maxlen);
 }
 
+int HID_API_EXPORT_CALL hid_darwin_get_location_id(hid_device *dev, uint32_t *location_id)
+{
+	int res = get_int_property(dev->device_handle, CFSTR(kIOHIDLocationIDKey));
```
Youw, line 1181: I'm pretty sure that the `location_id` is an unsigned int property
lets add a `get_uint_property` function, to be strictly correct

Youw (approved):

mac/hid.c
triplef: Yes I also think it’s an unsigned property, and the return parameter is declared as such (`uint32_t *location_id`).

Since it’s the same bit size I don’t see any reason to add a separate getter. In fact `get_int_property()` is already used in `get_vendor_id()` to return an unsigned property the same way. But I added a cast just to be explicit.

PR #395, opened 2022-03-15
libusb: Fix double free of buffer pointer in hid_close()

Sneha1P:
In hid_close() buffer pointer is freed and next to that the
libusb_free_transfer function is called which checks for if
the LIBUSB_TRANSFER_FREE_BUFFER flag is set and the buffer
pointer is not NULL. when this condition evaluates true, it
tries to free memory for a buffer which is already free and
a crash occurs.
Make buffer pointer NULL, once it is freed.

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1408,6 +1408,7 @@ void HID_API_EXPORT hid_close(hid_device *dev)
 
 	/* Clean up the Transfer objects allocated in read_thread(). */
 	free(dev->transfer->buffer);
+	dev->transfer->buffer = NULL;
 	libusb_free_transfer(dev->transfer);
 
 	/* release the interface */


Sneha1P:
what will be the correct approach to fix the crash?

        free(dev->transfer->buffer);
+      dev->transfer->buffer = NULL;
        libusb_free_transfer(dev->transfer);

OR

+      dev->transfer->flags |= LIBUSB_TRANSFER_FREE_BUFFER;
        libusb_free_transfer(dev->transfer);

Sneha1P:
@Youw, Yes I have hit the issue.

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f3597f8f3fa in __GI_abort () at abort.c:89
#2  0x00007f3597fcbbd0 in __libc_message (do_abort=do_abort@entry=2, fmt=fmt@entry=0x7f35980c0dd0 "*** Error in `%s': %s: 0x%s ***\n")
    at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007f3597fd1f96 in malloc_printerr (action=3, str=0x7f35980c0f28 "double free or corruption (fasttop)", ptr=<optimized out>,
    ar_ptr=<optimized out>) at malloc.c:5049
#4  0x00007f3597fd27de in _int_free (av=0x7f3590000020, p=0x7f35900008b0, have_lock=0) at malloc.c:3905
#5  0x00007f3598db8ed3 in hid_close (dev=0x5631eccc4500) at hid.c:1249
```
Using hidapi-0.9.0

Youw (approved):
Current solution is acceptable as is.

Better question: what version of libusb are you using?

For the sake of getting to bottom of this: any chance you could debug to find out who sets the "FREE_BUFFER" flag?

Sneha1P:
I am using libusb-1.0.20

yes, it's true that in hidapi flag is not getting set, but In libusb I found the function "libusb_control_transfer" which sets this FREE_BUFFER flag, and this particular libusb function is called from hid_write() in /libusb/hid.c

Youw:
> "libusb_control_transfer" which sets this FREE_BUFFER flag, and this particular libusb function is called from hid_write()

That doesn't sound right. `libusb_control_transfer` creates a temporary transfer and deletes it on its own. It has nothing to do with `dev->transfer` that we use only for read/interrupt reports.

Something else is going on here.

mcuee:
> I am using libusb-1.0.20

Hmm, that is a very old version. Can you try later version like 1.0.25 release?

Issue #344, opened 2021-10-13
Composite hid device that's also a keyboard on Mac OS X


dnaq:
I have an issue with implementing a HID-based protocol for a BLE device that is also a keyboard an getting it to work under Mac OS X. My guess is that the operating system refuses to give us access to the hid device due to the fact of it also being a keyboard. When using USB the solution is pretty simple, just use a different endpoint for the custom protocol, but for BLE we can only have a single HOG endpoint. I'm wondering if there is, or if it would be possible to implement, a way for hidapi to not try to grab the whole device, but instead just grab the reports of the usages we're interested in.

Issue #27, opened 2019-06-18
mac: Option to open device in non-exclusive mode


z3ntu:
Example: If you open a mouse currently (on macOS), the mouse input doesn't reach the OS, so you can't move the cursor with the mouse anymore while the device is opened in hidapi. That can be solved by not opening the device in exclusive mode, so https://github.com/z3ntu/hidapi/commit/fc9077c470803ee89d4748b2dcbd62328fc224f8 / https://github.com/signal11/hidapi/pull/297

For my specific use case I have to look if it's possible to open a different usb device interface from the same device that doesn't block mouse input, but it seems that this feature is wanted by at least one other person too.

Youw:
I think https://github.com/signal11/hidapi/pull/335 refers to the same general issue, but on Windows
we need to consider an API that would allow specifying access mode

z3ntu:
a.k.a https://github.com/libusb/hidapi/pull/1

On Windows it seems it's not possible to open some devices in exclusive mode. On Mac both is possible, so no fallback is possible on the hidapi side. About Linux, I don't know :).

Youw:
hmm, yes, it is different
would be nice to find out what possible regressions could https://github.com/z3ntu/hidapi/commit/fc9077c470803ee89d4748b2dcbd62328fc224f8 possibly cause, if there is any
after going thru documentation - I'd say, that there shouldn't be any (direct) issues

on the other hand, we could add something like:
`hid_open_path_opt(char* path, int bit_options)` and define different options for different platforms (like OPEN_EXCLUSIVE for mac)

z3ntu:
https://github.com/signal11/hidapi/issues/295#issuecomment-272234180

mattsains:
I opened PR #148 to solve this:

Youw:
So I give this problem some thinking, and here what I suggest:
- accept #148 as is;
- add new API: `hid_open<_path>_exclusive` to cover current behavior;

Motivation:
1) On Windows `hid_open/_path` is non-exclusive in any case, so #148 will make the behavior consistent, which is great;
   - still need to check how `hidraw`/`libusb` behaves - don't really know yet;

    It will be very convenient for many users who find exclusive mode on macOS inconvenient/unusable;

2) `hid_open<_path>_exclusive` will make possible to preserve the old behavior for users which may rely on it;

3) Such approach won't break anything on both API/ABI levels for users who don't care how the device is open on macOS.

We still have 2 options for `hid_open<_path>_exclusive`:
[...]

PR #397, opened 2022-03-16
darwin: add option to open device in non-exclusive mode

Youw:
Resolves: #27
Resolves: #344

### DIFF
--- a/mac/hid.c
+++ b/mac/hid.c
@@ -108,8 +108,21 @@ struct input_report {
 	struct input_report *next;
 };
 
+static struct hid_api_version api_version = {
+	.major = HID_API_VERSION_MAJOR,
+	.minor = HID_API_VERSION_MINOR,
+	.patch = HID_API_VERSION_PATCH
+};
+
+/* - Run context - */
+static	IOHIDManagerRef hid_mgr = 0x0;
+static	int is_macos_10_10_or_greater = 0;
+static	IOOptionBits device_open_options = 0;
+/* --- */
+
 struct hid_device_ {
 	IOHIDDeviceRef device_handle;
+	IOOptionBits open_options;
 	int blocking;
 	int uses_numbered_reports;
 	int disconnected;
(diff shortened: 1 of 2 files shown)


Youw:
@dnaq try if this works for you.

dnaq:
That looks like it would work beautifully for my use case.

DJm00n:
I thought that hidapi idea is to provide generic crossplatform API if possible.
We can add such option for Windows too (do not pass `FILE_SHARE_READ` and `FILE_SHARE_WRITE` to [CreateFile](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew)). Maybe its better to do not-os-specific `hid_open_path` instead?

Youw:
I already had this conversation somewhere (but I can't find it, so here it goes again):
1) HID Standard doesn't define such things like sharing the device on host side, so we can't reference to any existing specification or smth;
2) Not all platforms supports such notion at all:
    - You can do Exclusive or Non-Exclusive Open on macOS;
    - (apparently) You can do Shared or Non-Shared Open on Windows (I never checked the actual behavior when you do Non-Shared Open on Windows);
    - LIBUSB nor HIDRAW backends doesn't support Non-Exclusive Open due to OS/API specifics;

Having a "cross-platform" function that is known only to work for _some_ platforms defeats the whole "cross-platform" idea.

Pretty much the same thing we can tell about similar option: ReadWrite or ReadOnly Open: supported by Windows and HIDRAW, but not supported by macOS and LIBUSB. More than that - on Window you can open a device without specifying ReadWrite or ReadOnly, and on HIDRAW you have to specify one of them.

---
There is also a question of consistency, e.g.: "shouldn't a cross-platform application behave consistently across platforms?".
I.e. since we open on Windows in shared mode, why don't we open in Non-Exclusive mode on macOS by default?
I already asked myself this question [here](https://github.com/libusb/hidapi/issues/27#issuecomment-590607926).
In addition to that: on LIBUSB we open with exclusive access (and there is no option to do otherwise), so current behavior on macOS is consistent with LIBUSB.

---
Given all of the above, my strategy is:
- try&_keep the existing behavior whenever possible_ (unless it is a bug, which better be fixed, like #334);
- (recently new strategy) Add platform-specific options for options/platforms where it makes sense.

PR #398, opened 2022-03-16
Fire v0.12.0

Youw:
- add convenient macros to check HIDAPI version in runtime;
- mark in which version all recent HIAPI API functions where added;


windows/hidapi_winapi.h
```diff
@@ -19,6 +19,8 @@
 
 /** @file
  * @defgroup API hidapi API
+
```
Youw, line 22:
```suggestion
 *
```

PR #404, opened 2022-04-25
Fix find_package(hidapi) warnings

Youw:
- FindPkgConfig.cmake warns it is not called by find_package;
- CMake package `config` file and `exports` file names should not match,
otherwise it gets included recursively by GLOB expression;


stsichler:
this commit seems to break detection of `find_package()` of specific versions, see #482

Youw:
Looks like so.
We need a CI check for scenario described in #482.
Thanks for finding it.

Any chance you could provide a PR?

PR #405, opened 2022-04-25
libusb: cmake: Proper dependency on Iconv

Youw:
- explicitly add Iconv as a dependent library;
- check if iconv requires pointer-to-const as input (introduce ICONV_CONST check);
- NetBSD CI (has external Iconv library implementation that uses all of the above);

Closes: #44


Youw:
OMG!

Several hours of coding (with some "dummy" tries on Linux) and _no build failures on CI for the first try_!

PR #409, opened 2022-05-04
v0.12.0-RC1 found fixes

Youw:


mcuee:
Tested to be okay under macOS. No more warnings and hidtest build is okay.

mcuee (approved):
Looks good to me. But I am probably not a good reviewer for all the changes. So it is better to have one more approvals.

Youw:
> I am probably not a good reviewer

well:

> Tested to be okay under macOS.

That matters _a lot_.

PR #425, opened 2022-06-11
fix: make the behaviour of methods which set hid_error more consistent

Julusian:
From reading through the code I found that the behaviour of `hid_error()` was rather inconsistent at times.
Often methods would only set the value if they had a new error, which would make it close to impossible to figure out if the value of `hid_error()` was meaningful.

The windows implementation was especially bad and would only set the device error if it encountered an error, and would never clear the last error.
The linux hidraw implementation would sometimes clear it, but not always.

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -399,6 +399,8 @@ extern "C" {
 			still contain the Report ID, and the report data will
 			start in data[1].
 
+			This function sets the return value of hid_error().
+
 			@ingroup API
 			@param device A device handle returned from hid_open().
 			@param data A buffer to put the read data into, including
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -1053,6 +1053,8 @@ int HID_API_EXPORT hid_send_feature_report(hid_device *dev, const unsigned char 
 {
 	int res;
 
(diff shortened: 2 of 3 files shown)


Julusian:
I suppose that it could be fine as it, as you know that there is a new error from the return code, so maybe none of this is necessary other than the docs fix

Youw:
You are right, there is so much wrong about `hid_error` in general.
Starting with the inconsistency in the documentation and implementation, ending up with real thread-safety issues (despite the documentation).

We've introduced a lot problems by trying to make an [improvement](https://github.com/libusb/hidapi/pull/23).
We tried to make `hid_error` mimic the [`errno`](https://man7.org/linux/man-pages/man3/errno.3.html) behavior, on the thread-safety part, but that was practically impossible to do so for a few reasons:
- hid_error has a per-device version of the function, and the device (struct) member cannot contain thread-local members;
- unlike `errno`, which is an `int` - hid_error is a dynamically allocated string, which has to be `free`d by each thread, where the error has been happening, and that is not true for `hidraw` (the only backend that has global errors implementation);

Let me finally start making it [right](https://github.com/libusb/hidapi/tree/hid_error).

PR #410, opened 2022-05-04
Meson support

Youw:
As a wrapepr over CMake.
Still experimental support and definitely not a primary build system.

PR #426, opened 2022-06-12
add meson

neheb:
Helps to avoid iconv problems that the CMake build has.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,47 @@
+project('hidapi', 'c', version: files('VERSION'),meson_version: '>=0.60')
+
+iconv_dep = dependency('iconv', required: get_option('libusb'))
+libusb_dep = dependency('libusb-1.0', required: get_option('libusb'))
+udev_dep = dependency('libudev', required: get_option('hidraw'))
+
+incdirs = include_directories('hidapi')
+
+install_headers('hidapi/hidapi.h', subdir: 'hidapi')
+
+pconf = import('pkgconfig')
+
+if libusb_dep.found() and iconv_dep.found()
+  libhidapi = library('hidapi-libusb',
+    'libusb/hid.c',
+    dependencies: [ iconv_dep, libusb_dep ],
+    include_directories: incdirs,
+    version: meson.project_version(),
+    install: true
+  )
+
+  pconf.generate(
+    libhidapi,
+    description: 'C Library for USB HID device access from Linux, Mac OS X, FreeBSD, and Windows. This is the libusb implementation.',
+    url: 'https://github.com/libusb/hidapi',
+    subdirs: 'hidapi',
+  )
+endif
+
+if udev_dep.found() and host_machine.system() == 'linux'
+  hidraw_lib = library('hidapi-hidraw',
+    'linux/hid.c',
+    dependencies: udev_dep,
+    include_directories: incdirs,
+    version: meson.project_version(),
+    install: true
+  )
+
+  pconf.generate(
+    hidraw_lib,
+    description: 'C Library for USB/Bluetooth HID device access from Linux, Mac OS X, FreeBSD, and Windows. This is the hidraw implementation.',
+    url: 'https://github.com/libusb/hidapi',
+    subdirs: 'hidapi',
+  )
+
+  install_headers('libusb/hidapi_libusb.h', subdir: 'hidapi')
+endif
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -0,0 +1,53 @@
+project(
+  'hidapi',
+  'c',
+  version : '0.12.0',
```
eli-schwartz, line 4:
```suggestion
  version : files('VERSION'),
```

The other build systems read the version from here, might as well do it for Meson too. ;)

meson.build
```diff
@@ -0,0 +1,53 @@
...
+incdirs = include_directories('hidapi')
+
+install_headers('hidapi/hidapi.h', subdir: 'hidapi')
+
+pconf = import('pkgconfig')
+
+if libusb_dep.found() and iconv_dep.found()
+  libhidapi = library('hidapi-libusb',
+    'libusb/hid.c',
+    dependencies: [ iconv_dep, libusb_dep ],
```
eli-schwartz, line 22: This should depend on threads, I think.

Also, it is only called "-libusb" on linux.

And on Windows or macOS it uses the sources from windows/ or mac/ instead. See e.g. https://github.com/libusb/hidapi/blob/59e84ca74fae41a43f8ff3ea96b4f956c79c69b4/Makefile.am#L18-L40

meson.build
```diff
@@ -0,0 +1,53 @@
...
+    libhidapi,
+    description: 'C Library for USB HID device access from Linux, Mac OS X, FreeBSD, and Windows. This is the libusb implementation.',
+    url: 'https://github.com/libusb/hidapi',
+    subdirs: 'hidapi',
+  )
+endif
+
+if udev_dep.found()
+  hidraw_lib = library('hidapi-hidraw',
+    'linux/hid.c',
```
eli-schwartz, line 38: I think this is only built on Linux.

eli-schwartz (changes requested):

meson.build
neheb: threads probably implicitly gets included with libusb.

meson.build
```diff
@@ -0,0 +1,47 @@
...
+install_headers('hidapi/hidapi.h', subdir: 'hidapi')
+
+pconf = import('pkgconfig')
+
+if libusb_dep.found() and iconv_dep.found()
+  libhidapi = library('hidapi-libusb',
+    'libusb/hid.c',
+    dependencies: [ iconv_dep, libusb_dep ],
+    include_directories: incdirs,
+    version: files('VERSION'),
```
eli-schwartz, line 18: This works for project, specifically.

For libraries just use meson.project_version()...

meson_options.txt
```diff
@@ -0,0 +1,7 @@
+option('hidraw', type : 'feature',
+  description : 'Build hidraw. Requires udev.',
```
JoergAtGithub, line 2:
```suggestion
description : 'Build hidapi with hidraw backend. Requires udev.',
```

meson_options.txt
```diff
@@ -0,0 +1,7 @@
...
+)
+
+option('libusb', type : 'feature',
+  description : 'Build hidapi. Requires libusb.',
```
JoergAtGithub, line 6:
```suggestion
description : 'Build hidapi with libusb backend. Requires libusb.',
```

Youw:
What does it solve compared to #410?

> avoid iconv problems that the CMake build has

There is 0 known issues related to `iconv`: https://github.com/libusb/hidapi/issues?q=is%3Aissue+is%3Aopen+iconv

Can you elaborate?

neheb:
> What does it solve compared to #410?
> 
> > avoid iconv problems that the CMake build has
> 
> There is 0 known issues related to `iconv`: https://github.com/libusb/hidapi/issues?q=is%3Aissue+is%3Aopen+iconv
> 
> Can you elaborate?

didn't know there was another PR.

Anyway, issue is that when both external and internal iconv implementations exist, CMake's find_package(Iconv) fails to find iconv and terminates the build. More info here:

https://github.com/openwrt/packages/actions/runs/2477509891
https://github.com/openwrt/packages/pull/18738

solved by using dependency('iconv') which just works.

Youw:
I'm pretty sure, we can find a solution to CMake's `find_package(Iconv)`.

---
I do not support having yet another alternative build system for HIDAPI when it can be avoided - you'd have to duplicate the entire build logic we already have in CMake.
Lets focuse on #410 instead.

PR #428, opened 2022-06-12
hid_error improvements

Youw:
Finally make the documentation and the implementation of `hid_error` consistent.

Closes: #425

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -399,6 +399,8 @@ extern "C" {
 			still contain the Report ID, and the report data will
 			start in data[1].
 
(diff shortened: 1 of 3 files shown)


Youw:
@Julusian if you have comments regarding this - feel free to leave them here.

Julusian:
Yeah this looks fine to me.

Initially it felt like a shame to lose the thread safety of the error on linux, but as long as hid_open, hid_enumerate, hid_init are contained to a single thread (which it looks like they should be anyway), then I dont see it losing any safety.

Youw:
> which it looks like they should be anyway

Yeap, still need to write the documentation for it.

Issue #429, opened 2022-06-13
find_package(Iconv) fails when external iconv implementations is provided as CFLAGS/LDFLAGS


Youw:
Reported at: https://github.com/libusb/hidapi/pull/426
Failed sample: https://github.com/openwrt/packages/actions/runs/2477509891

Youw:
> https://github.com/openwrt/packages/runs/6838046113?check_suite_focus=true

@neheb would it be possible to see the content of:
> "/home/build/openwrt/build_dir/target-x86_64_musl/hidapi-hidapi-0.12.0/CMakeFiles/CMakeOutput.log".
> "/home/build/openwrt/build_dir/target-x86_64_musl/hidapi-hidapi-0.12.0/CMakeFiles/CMakeError.log".

Maybe you can run the build locally or so.


---
And the content of:
> "/home/build/openwrt/staging_dir/target-x86_64_musl/usr/lib/libiconv-stub/lib"

I woulder what CMake tries (and fails) to find there.

Youw:
Looks like this is a very specific case, during the cross-compilation, and when iconv is cross-compiled as well, and passed to CMake build externally.
This might be an issue specific to openwrt build infrastructure.

Youw:
Problem analysis:

From the logs I can tell that openwrt doesn't follow minimal guidelines for CMake [cross-compilation](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html) - no propper toolchain or specifying paths/[environment](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html#variable:CMAKE_PREFIX_PATH) to the libraries, the way CMake can recognize it.

One way of fixing it on the environment side is to set the `CMAKE_PREFIX_PATH` to:
`/home/mangix/devstuff/openwrt/staging_dir/target-mips_24kc_musl/usr;/home/mangix/devstuff/openwrt/staging_dir/target-mips_24kc_musl/usr/lib/libiconv-full;/home/mangix/devstuff/openwrt/staging_dir/target-mips_24kc_musl/usr/lib/libintl-full`.

But that way, obviously, requres changes to `cmake.mk` (maybe even more), which I don't expect to be done any time soon.
[...]

PR #430, opened 2022-06-18
Ensure Iconv is found when provided via CFLAGS/LDFLAGS

Youw:
- by default find_file/find_library doesn't respect CFLAGS/LDFLAGS, and FindIconv fails to find Iconv;
- by explicitly trying to link against `-liconv` - we're checking if library is available in such way;
- additionally: if Iconv is detected as BUILT_IN, no need to explicitly depend on `Iconv::Iconv`;

Fixes: #429


Youw:
Right now I don't have the environment to check this properly (except what is setup on CI).

@neheb can you check this as part of the openwrt build?

neheb:
As soon as I get home.

neheb:
nope:

```
-- The C compiler identification is GNU 12.1.0
-- Check for working C compiler: /home/mangix/devstuff/openwrt/staging_dir/toolchain-mips_24kc_gcc-12.1.0_musl/bin/mips-openwrt-linux-musl-gcc - skipped
-- hidapi: v0.12.0
-- Looking for pthread.h
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found PkgConfig: /home/mangix/devstuff/openwrt/staging_dir/host/bin/pkg-config (found version "1.8.0")
-- Checking for module 'libudev'
--   Found libudev, version 243
-- Checking for module 'libusb-1.0>=1.0.9'
--   Found libusb-1.0, version 1.0.24
-- Performing Test Iconv_IS_BUILT_IN
-- Performing Test Iconv_IS_BUILT_IN - Failed
-- Could NOT find Iconv (missing: Iconv_LIBRARY)
-- Could NOT find Iconv (missing: Iconv_LIBRARY)
CMake Error at libusb/CMakeLists.txt:41 (message):
  Iconv is not found, make sure your build environment is right


-- Configuring incomplete, errors occurred!
See also "/home/mangix/devstuff/openwrt/build_dir/target-mips_24kc_musl/hidapi-hidapi-0.12.0/CMakeFiles/CMakeOutput.log".
See also "/home/mangix/devstuff/openwrt/build_dir/target-mips_24kc_musl/hidapi-hidapi-0.12.0/CMakeFiles/CMakeError.log".
```

neheb:
Seems to have done it. A lot of work as compared to

```
dependency('iconv')
```

though.

Youw:
Not a lot of work, compared to _uncompleted_ #426 :)

Besides, I'll ask CMake community, if this can be improved by CMake itself, even though openwrt CMake support is too incomplete as of yet.

---
Thanks for testing it.

libusb/CMakeLists.txt
```diff
@@ -22,11 +22,53 @@ target_link_libraries(hidapi_libusb PRIVATE Threads::Threads)
...
+                    #include <stddef.h>
+                    #include <iconv.h>
+                    int main() {
+                    char *a, *b;
+                    size_t i, j;
+                    iconv_t ic;
+                    ic = iconv_open(\"to\", \"from\");
+                    iconv(ic, &a, &i, &b, &j);
+                    iconv_close(ic);
+                    }
```
Youw, line 56:
```suggestion
                    int main() {
                        char *a, *b;
                        size_t i, j;
                        iconv_t ic;
                        ic = iconv_open(\"to\", \"from\");
                        iconv(ic, &a, &i, &b, &j);
                        iconv_close(ic);
                    }
```

neheb:
I do agree CMake's Find_Iconv is not that great, Fixing it upstream would be good.

The OpenWrt comment is interesting. I don't really know what the "proper" way of handling cross compilation is like this. Although cmake.mk probably predates whatever "proper" is.

Youw:
> I don't really know what the "proper" way of handling cross compilation is like this.

Let me share with you a few links, as a general reference:
- The best way to cross-compile with CMake is to prepare a CMake _toolchain_ file. Here is the official documentation: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html
So instead of specifying so many CMake arguments and environment varibales - a single additional CMake argument should be passed: `-DCMAKE_TOOLCHAIN_FILE=<toolchain for cross-compilation>`;
- a very good example of CMake _toolchain_ file is the one provided by Android [NDK](https://developer.android.com/ndk/downloads). The one I used recently is located at: `ndk/22.1.7171670/build/cmake/android.toolchain.cmake`;
- some while ago I've been using Yocto builds for one of the development boards; Yocto has mature CMake support - that I remember, even though cannot really share a specific link to check out;

---
BTW:
> the one provided by Android

I'm successfully using with HIDAPI (and not only) in several of my projects.

PR #435, opened 2022-07-15
Win32: modified hid_winapi_get_container_id()  for consistency

DJm00n:
Make use of `hid_internal_get_devnode_property()` method.


Youw (commented):
As much as I like consistency, new implementation makes extra memory allocation/deallocation, and the old one - doesn't.

I don't think overall readability is better enough to pay the runtime price.

DJm00n:
This method is not meant to be called frequently so I think we can disregard this 8 byte malloc.

Youw (approved):

Youw:
Actually, there is one more issue with the new implementation: no more explicit check for the buffer size.
It is highly unlikely that WinAPI will report a `DEVPROP_TYPE_GUID` but `len` of smaller than `sizeof(GUID)`, but I like it much better if there is no implicit assumptions in the code.

DJm00n:
>It is highly unlikely that WinAPI will report a DEVPROP_TYPE_GUID but len of smaller than sizeof(GUID)

According to the documentation, this should not be a problem:

<html>
<body>


Property data type | Property value size requirement
-- | --
A fixed-length base-data-type value | The specified size of the supplied data must be the number of bytes in the base data type.


</body>
</html>

https://docs.microsoft.com/en-us/windows-hardware/drivers/install/property-value-requirements

>Set the PropertyType parameter to DEVPROP_TYPE_GUID, set the PropertyBuffer parameter to a pointer to a buffer that contains a GUID value, and set the PropertyBufferSize parameter to sizeof(GUID).

https://docs.microsoft.com/en-us/windows-hardware/drivers/install/devprop-type-guid

Youw:
> According to the documentation

Unfortunately I've personally seen cases when WinAPI didn't behave _as per documentation_.

My consern stands:
- when we pass a buffer that is bigger than an API uses - that's fine;
- if we ask how large buffer API needs, allocate memory for it, and then _don't actually check its size_ and try to read it - that is a potential issue;

PR #441, opened 2022-08-13
Reattach kernel driver when failed to claim interface

Youw:
Potential execution path may happen if kernel driver successfully detached,
but libusb failed to claim the requested interface.

Otherwise the device remains with detached kernel driver.


mcuee (approved):
Looks good to me.

PR #446, opened 2022-08-20
linux: Open files with `O_CLOEXEC` to not leak fds to child processes

ids1024:
In general, `open()` should always be called with `O_CLOEXEC`, except immediately before an `exec*` call that will use the fd.

Previously the fd was leaked to a child process, as seen here:

```c

int main() {
    assert(hid_open_path("/dev/hidraw0") > 0);
    char *args[] = {"/bin/sh", "-c", "readlink /proc/self/fd/*", NULL};
    execve("/bin/sh", args, NULL);
}
```

In some cases this could be a security issue if child processes may be less privileged and shouldn't have access to the file.

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -384,7 +384,7 @@ static int get_hid_report_descriptor(const char *rpt_path, struct hidraw_report_
 	int rpt_handle;
 	ssize_t res;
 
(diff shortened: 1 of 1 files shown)


Youw (approved):
Thanks!

PR #457, opened 2022-09-13
libusb: stop `read_thread` from looping indefinitely (#456)

### LINKED ISSUE #456: hidapi-libusb: if `libusb_submit_transfer` fails in read_thread, `hid_close` hangs
I have a usb device (that is finicky with its endpoints, it identifies itself as a hid_device and responds to hid requests, but is missing an IN interrupt endpoint, so Linux does not recognize it as HID) that I open using `hid_libusb_wrap_sys_device` with a linux file descriptor. I can send and get feature reports just fine; however, on `hid_close`, it hangs: [specifically in checking for the transfer loop to finish (hidapi/libusb/hid.c:1001)](https://github.com/libusb/hidapi/blob/0f2cf886e5a91ff13f222a67d8931c0822244c9a/libusb/hid.c#L1001) . I ran with the env variable `LIBUSB_DEBUG=4` and saw that the call to `submit_bulk_transfer` in libusb fails. 

I'm not sure how this interacts with the `read_callback`, but `dev->transfer_loop_finished` never gets set to `1` and `libusb_handle_events_completed(usb_context, &dev->transfer_loop_finished);` takes the full 60 seconds to complete before restarting the while loop.

thesilvanator:
- When `libusb_submit_transfer` in `read_thread` fails, `read_callback` never gets called and never sets `transfer_loop_finished` to true, causing `read_thread` to loop indefinitely.

- Fix this by skipping the handle_events for the transfer

- Fixes #456

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -952,6 +952,7 @@ static void read_callback(struct libusb_transfer *transfer)
 
 static void *read_thread(void *param)
 {
+	int res;
 	hid_device *dev = param;
 	uint8_t *buf;
 	const size_t length = dev->input_ep_max_packet_size;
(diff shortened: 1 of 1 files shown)


Youw (approved):
This does look about right

thesilvanator:
I see that you have approved this, but the one of the CI checks seems to be still waiting for completion. Is there anything I can do to fix this?

Youw:
No worries, that is not a blocker.

thesilvanator:
Ah, ok. So when does this eventually get merged in?

PR #469, opened 2022-10-20
Use %ls format specifier for README example

endolith:
I finally got it running, but it was printing only the first character.  %ls is for wide strings and works.

https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823

"If you want your code to work both on runtimes that use the Windows classic printf rules as well as those that use C standard printf rules, you can limit yourself to %hs for narrow strings and %ls for wide strings, and you’ll get consistent results regardless of whether the format string was passed to sprintf or wsprintf."

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,19 +120,19 @@ int main(int argc, char* argv[])
 
 	// Read the Manufacturer String
 	res = hid_get_manufacturer_string(handle, wstr, MAX_STR);
-	wprintf(L"Manufacturer String: %s\n", wstr);
+	wprintf(L"Manufacturer String: %ls\n", wstr);
 
 	// Read the Product String
 	res = hid_get_product_string(handle, wstr, MAX_STR);
-	wprintf(L"Product String: %s\n", wstr);
+	wprintf(L"Product String: %ls\n", wstr);
 
 	// Read the Serial Number String
 	res = hid_get_serial_number_string(handle, wstr, MAX_STR);
-	wprintf(L"Serial Number String: (%d) %s\n", wstr[0], wstr);
+	wprintf(L"Serial Number String: (%d) %ls\n", wstr[0], wstr);
 
 	// Read Indexed String 1
 	res = hid_get_indexed_string(handle, 1, wstr, MAX_STR);
-	wprintf(L"Indexed String 1: %s\n", wstr);
+	wprintf(L"Indexed String 1: %ls\n", wstr);
 
 	// Toggle LED (cmd 0x80). The first byte is the report number (0x0).
 	buf[0] = 0x0;


Youw (approved):
Thanks!

endolith:
(It probably should exit when hid_open returns NULL, also, or [it produces SIGSEGV when the device isn't plugged in](https://github.com/signal11/hidapi/issues/285#issuecomment-239636071).  Confusing for beginners like me.)

endolith:
Also apparently [you can't output with `printf` and `wprintf` in the same file](https://stackoverflow.com/a/8682010/125507)...

Youw:
I guess for beginners the [hidtest](https://github.com/libusb/hidapi/blob/master/hidtest/test.c) would be a better sample and it was updated/improved over time a lot.

I you want to improve sample in the README (e.g. replace `wprintf` with `printf` as it is used in `hidtest`) - you're very welcome to do so.

README.md
```diff
@@ -120,19 +120,19 @@ int main(int argc, char* argv[])
 
 	// Read the Manufacturer String
 	res = hid_get_manufacturer_string(handle, wstr, MAX_STR);
-	wprintf(L"Manufacturer String: %s\n", wstr);
+	wprintf(L"Manufacturer String: %ls\n", wstr);
```
endolith, line 123:
```suggestion
	printf("Manufacturer String: %ls\n", wstr);
```

README.md
```diff
@@ -120,19 +120,19 @@ int main(int argc, char* argv[])
...
 
 	// Read the Product String
 	res = hid_get_product_string(handle, wstr, MAX_STR);
-	wprintf(L"Product String: %s\n", wstr);
+	wprintf(L"Product String: %ls\n", wstr);
```
endolith, line 127:
```suggestion
	printf("Product String: %ls\n", wstr);
```

README.md
```diff
@@ -120,19 +120,19 @@ int main(int argc, char* argv[])
...
 
 	// Read the Serial Number String
 	res = hid_get_serial_number_string(handle, wstr, MAX_STR);
-	wprintf(L"Serial Number String: (%d) %s\n", wstr[0], wstr);
+	wprintf(L"Serial Number String: (%d) %ls\n", wstr[0], wstr);
```
endolith, line 131:
```suggestion
	printf("Serial Number String: (%d) %ls\n", wstr[0], wstr);
```

README.md
```diff
@@ -120,19 +120,19 @@ int main(int argc, char* argv[])
...
 
 	// Read Indexed String 1
 	res = hid_get_indexed_string(handle, 1, wstr, MAX_STR);
-	wprintf(L"Indexed String 1: %s\n", wstr);
+	wprintf(L"Indexed String 1: %ls\n", wstr);
```
endolith, line 135:
```suggestion
	printf("Indexed String 1: %ls\n", wstr);
```

endolith:
@Youw Yeah, I'm just new to this and not sure of the conventions, but I'll try.

PR #470, opened 2022-10-20
README example: prevent segmentation fault and mixture of printf/wprintf

endolith:
When no device is connected (or user doesn't have permission to access
device), hid_open returns NULL.  If NULL is passed to
hid_get_manufacturer_string(), it crashes with SIGSEGV, which could be
confusing to newcomers trying to get example to work.

wprintf and printf aren't compatible in the same stream

Now it's more similar to https://github.com/libusb/hidapi/blob/master/hidtest/test.c#L129

Closes: #469

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,19 +120,19 @@ int main(int argc, char* argv[])
 
 	// Read the Manufacturer String
 	res = hid_get_manufacturer_string(handle, wstr, MAX_STR);
-	wprintf(L"Manufacturer String: %s\n", wstr);
+	wprintf(L"Manufacturer String: %ls\n", wstr);
 
 	// Read the Product String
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,22 +117,26 @@ int main(int argc, char* argv[])
 	// Open the device using the VID, PID,
 	// and optionally the Serial number.
 	handle = hid_open(0x4d8, 0x3f, NULL);
+	if (!handle) {
+		printf("Unable to open device\n");
+ 		return 1;
```
Youw, line 122:
```suggestion
		printf("Unable to open device\n");

		hid_exit()
 		return 1;
```

Youw (approved):

README.md
endolith: Good point, but make the same change in the other test.c 😁

Issue #480, opened 2022-11-09
Windows: lib import/export


OgreTransporter:
https://github.com/libusb/hidapi/blob/2b70208feabeb4cb524dff58c450af63544790bc/hidapi/hidapi.h#L33
This does not make sense. `dllexport` is normally always used in combination with `dllimport` (https://learn.microsoft.com/en-us/cpp/cpp/dllexport-dllimport).

Youw:
"does not make" doesn't mean it is broken.
I remember 0 complains about this part.
And I personally use HDIAPI both as a shared and as a static library on multiple platforms, including Windows.

Please share a specific use-case where the old behavior has compilation/runtime issues. If there is such - we'll try to find a backward-compatible solution for all build systems.

OgreTransporter:
I never claimed that errors occur. But I can also argue the other way around, that just because you haven't had an error yet, it doesn't always have to work. After all, the Windows API provides for the combination of `dllimport` and `dllexport`. This is a general improvement suggestion to make the code cleaner. Many libraries like Boost (https://github.com/boostorg/system/blob/develop/include/boost/system/config.hpp) use this kind of macro mechanism to implement the Windows API correctly.

Youw:
You're perfectly right about how things should be in the ideal world.
And I'm definitely considering making it right for HIDAPI v1+.

But right now - my strong position is: "don't try to fix what is not broken". And the #480 breaks all windows builds on our CI.
And our CI is the kind of "golden sample" for those who seek to start using HIDAPI or a reflection of how HIDAPI is already used by many others.

YgorSouza:
> "does not make" doesn't mean it is broken. I remember 0 complains about this part. And I personally use HDIAPI both as a shared and as a static library on multiple platforms, including Windows.
> 
> Please share a specific use-case where the old behavior has compilation/runtime issues. If there is such - we'll try to find a backward-compatible solution for all build systems.

For what it's worth, I did have problems with that. I linked HIDAPI statically into my DLL, and then noticed that all of its functions were accessible and visible with for example `dumpbin /exports`. Then I looked at the header and saw that there was no way to disable that, so I just copied the header to my project and edited out the dllexport. Now I am planning on using the hidapi Rust crate, which uses this repo as a submodule, and I realized I'll have to fork this repo and then fork the crate repo to make this same change.  That `#ifndef HIDAPI_STATIC` guard would simplify this a lot.

[...]

PR #481, opened 2022-11-09
Proposed solution for issue #480

OgreTransporter:


Youw:
While I totally understand the motivation for the change you're proposing, in HIDAPI v0 we choose to keep the backward compatibility as much as possible.

For this to work, it has to be done for all build systems: autotools, MSVC proj, Manual Makefiles, CMake.
And to be hones - I'd rather not touch *what is not broken*.

Unless you know a specific use-case where the old implementation has actual compilation/runtime issues - I don't believe any change is required as of right now.
A broken CI builds after the change supports my statement.

windows/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 list(APPEND HIDAPI_PUBLIC_HEADERS "hidapi_winapi.h")
+set(CMAKE_DEBUG_POSTFIX d)
```
Youw, line 2: this is a huge change from CMake/build perspective
HIDAPI choosed not to use it since the beginning

windows/CMakeLists.txt
```diff
@@ -20,26 +21,51 @@ target_link_libraries(hidapi_winapi
...
+            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+            PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/hidapi"
+        )
+    endif()
+else()
+    target_compile_definitions(hidapi_winapi PUBLIC HIDAPI_STATIC)
+    set_target_properties(hidapi_winapi
+        PROPERTIES
+            EXPORT_NAME "winapi"
+            OUTPUT_NAME "hidapi-static"
```
Youw, line 51: this is a breaking/backward incompatible changes

I'd say that most CMake changes are backward-incompatible and not even directly related to the original issue

windows/CMakeLists.txt
OgreTransporter: This is a great pity. With `CMAKE_DEBUG_POSTFIX` it is easier to distinguish debug and release builds. Otherwise, when using `INSTALL`, the two configurations overwrite each other. If you use CMake for further projects this option is not an API change either, because the exported targets are adapted as well.

windows/CMakeLists.txt
OgreTransporter: This is quite correct. Without the changes all produced libraries have the same name `hidapi`, no matter what configuration (`debug`, `release`) and build mode (`static`, `shared`). This is a bit of a hindrance during installation.

windows/CMakeLists.txt
Youw: if that is required for your specific case - you can always set that variable as an additional CMake parameter:
`-DCMAKE_DEBUG_POSTFIX=d`
or, in case if you're using HIDAPI as a subfolder - set the veriable before adding HIDAPI as a subfolder

windows/CMakeLists.txt
Youw: Can you share more info about your use-case, where you need to build/install both shared and static versions of the library?

windows/CMakeLists.txt
OgreTransporter: Currently, I also solve it this way. Currently, I also solve it this way. I build HIDAPI separately and copy static and dynamic versions together into an SDK for further use.

windows/CMakeLists.txt
OgreTransporter: I use HIDAPI for communication with self-built Ardunio-based input devices. For these to work in Unity I need the DLL, for all other applications in C/C++ I prefer to use the static variant. Everything needs to be debugged as well, so I need both the static and the dynamic library in the Debug and Release configurations.

Youw: For debug/non-debug I'd have to think a bit.
As for Shared/Static - looks like you're using the library in entirely different projects.
What stops you from having a separate shared and static builds under different locations?

OgreTransporter: Of course I can also create four directories and build each configuration individually. Then, of course, I must always take care of four directories when copying and distributing between projects and computers. To have everything in one directory is much clearer. Works also with most other libraries. Especially with the configurations different names are very practical, because these are also written into the CMake-Target-Exports. So I can add HIDAPI to a project via CMake, depending on the configuration the appropriate library is used automatically. In different directories I always have to run CMake with a different path for HIDAPI when changing the configuration.

One directory with four files and e.g. the CMake targets `hidapi::hidapi` and ` hidapi::hidapi_static` makes a lot easier. You can simply add HIDAPI to the project and choose between static and shared. Debug/release is done by CMake during project generation. Having everything at once also ensures that you don't work with different versions.

Youw: Are your projects CMake-based?

Youw:
We can continue the conversation to find posisble improvements for HIDAPI, but this is not goint to be accepted, and [this is](https://github.com/libusb/hidapi/pull/481#discussion_r1017905208) the main motivation why.

PR #486, opened 2022-11-30
Add support for HID over SPI

vaaronna:
This adds (or more accurately completes, since it was already started) support for HID over SPI devices. 
Handling is almost identical to HID over I2C.


Youw (approved):
This make me think that we should probably just generalize this routine for the general case/unknown BUSes, as it doesn't really affects anything except enumeration/device details/info.

But thanks anyway.

PR #488, opened 2022-12-02
Link gcc helpers statically on MinGW

jefdriesen:
When compiling a 32bit dll with the mingw-w64 compiler, some 64bit integer arithmetic operations are implemented using functions from libgcc (e.g. __udivdi3 and __umoddi3 from libgcc_s_dw2-1.dll). This unexpected dependency is inconvenient for applications.

The run-time dependency can be avoid by linking statically.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,6 +123,7 @@ case $host in
 	os="windows"
 	threads="windows"
 	win_implementation="mingw"
+	LDFLAGS="${LDFLAGS} -static-libgcc"
 	;;
 *-msys*)
 	AC_MSG_RESULT([ (Windows back-end, using MSYS2)])
@@ -130,6 +131,7 @@ case $host in
 	os="windows"
 	threads="windows"
 	win_implementation="mingw"
+	LDFLAGS="${LDFLAGS} -static-libgcc"
 	;;
 *-cygwin*)
 	AC_MSG_RESULT([ (Windows back-end, using Cygwin)])


Youw:
will that work with clang on Windows?

jefdriesen:
I've never tried building with clang on Windows. Libusb does this too, so I assume it's not causing any problems:
https://github.com/libusb/libusb/blob/master/configure.ac#L132
https://github.com/libusb/libusb/commit/f9162d36bf7253b3743d2973e385d32166ea0795

I'm using autotools because that's the build system I'm familiar with. My own project is also using autotools, and thus being able to build the dependencies with the same setup is convenient.

Youw:
1) Libusb does it only for [mingw and msys](https://github.com/libusb/libusb/blob/f3619c4078a921c3d6bf3238ff7a8dc70d60e40e/configure.ac#L126) for some reason, while this PR suggests it for all Windows platforms, including cygwin (I'm not sure how to use clang with autotools on Windows either, but still);

    While I'm not against this improvement, I want to make sure we don't break anything accidentally.

2) Autotools are deprecated for HIDAPI. You're still in liberty to use it, but I encourage you to try CMake instead, if you can.

jefdriesen:
I assume you'll still need mingw or msys when building with clang? For the windows header files and everything the autotools system needs (autoconf, automake, etc). Clang vs gcc is only the compiler part. But I don't know for sure either.

For Windows build, I'm actually cross-compiling on Linux. That's trivial with autotools (e.g. `./configure --host=${ARCH}-w64-mingw32`), but I don't know how to do the same with CMake.

Youw:
1) Lets play it safe anyway: apply this change only for mingw and msys,  similar way as Libusb does.

2) Just in case you're curios: to cross-compile anything with CMake, the [documentation](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html) suggests using Toolchain Files.
For Linux/Mingw there are some provided by the community, e.g.: https://gist.github.com/ebraminio/2e32c8f6d032a8e01606f7f564d2b1ee

jefdriesen:
Something like the updated version?

Youw (approved):
looks good

JoergAtGithub:
To build with clang on Windows, you just need to call the different compiler (clang-cl instead of cl) all libraries remain the same. Other than with mingw, the generated executable is ABI compatible, to that generated by the Microsoft compiler.
You just benefit from the better code optimizer in clang, and everything else remains the same.

Youw:
> To build with clang on Windows, you just need to call the different compiler (clang-cl instead of cl) all libraries remain the same. Other than with mingw, the generated executable is ABI compatible, to that generated by the Microsoft compiler. You just benefit from the better code optimizer in clang, and everything else remains the same.

Yeah, but autotolls doesn't use `cl` AFAIK.
And it is not the point. I just wanted to check either `clang-cl` (or any other Windows-specific variant of clang) would accept `-static-libgcc` compilation option as suggested in the very first commit in this PR (my guess is that it woudn't accept it).

PR #499, opened 2023-01-26
Add missing SOVERSION to Windows CMake build

swt2c:


Youw:
I don't believe it is missing per se.

Not sure what `SOVERSION` property does to Windows targets, but `SUFFIX` as introduced - definitely "unnatural" for Win32/DLL platforms.

If you really need those for you project/use-case - you can apply the same properties to `hidapi_winapi` in your build tree after you include HIDAPI as a subdirectory.

swt2c:
It's inconsistent with what the autoconf build does on Windows and also inconsistent with the other platforms on CMake.

Citation for "definitely "unnatural" for Win32/DLL platforms"?

Youw:
autoconf build on Windows cannot produce MSVC build and never been a native development environment, so that is not an authoritative reason.

If you see that CMake+MingGW/Cygwin VS Autotools producing different results - _maybe_ only that needs to be fixed.

As for MSVC/Win32 builds in general - SONAME-ing never used for binaries versioning on Windows by all mature projects, including Windows/VisualStudio developers itself - only embedded manifest/metadata. There is even a [WinAPI](https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfoa?redirectedfrom=MSDN) function to read that version/metadata.

Youw:
And there is still [`hidapi.sln`](https://github.com/libusb/hidapi/blob/master/windows/hidapi.sln) which _used to be_ the most official source of binaries on Windows and it always has been producing `hidapi.dll` w/o any additional name suffix(es)/prefix(es).

swt2c:
> autoconf build on Windows cannot produce MSVC build and never been a native development environment, so that is not an authoritative reason.

Sorry, I meant specifically mingw + autoconf.  That's how we've been building hidapi for Windows for close to 10 years:
https://src.fedoraproject.org/rpms/mingw-hidapi/blob/rawhide/f/mingw-hidapi.spec

Switching to cmake now, but as you say, the soname is unexpected, so we'll drop it.

Youw:
> we'll drop it

And you think it won't make any backward compatibility issues in a project like that? Ok then.

In any case I'll be happy to help if any assistance needed.

swt2c:
> > we'll drop it
> 
> And you think it won't make any backward compatibility issues in a project like that? Ok then.
> 
> In any case I'll be happy to help if any assistance needed.

No, nothing in the distribution uses it.  I have been building it for use mostly in one of my personal projects, so it should be fine.  Thanks though.

PR #500, opened 2023-01-27
windows: Get Bluetooth device Model Number String  instead of Device Name into product string

DJm00n:
Use `Device Name` only if `Model Number String` is not available. This should better match what we have in  `USB_DEVICE_DESCRIPTOR.iProduct`

See [Bluetooth Device Information Service](https://www.bluetooth.com/specifications/specs/device-information-service-1-1/) spec for more info.


windows/hid.c
```diff
@@ -409,32 +409,46 @@ static void* hid_internal_get_device_interface_property(const wchar_t* interface
...
-	/* Get devnode grandparent to reach out Bluetooth LE device node */
-	if (CM_Get_Parent(&dev_node, dev_node, 0) != CR_SUCCESS)
-		return;
+	if (wcslen(dev->product_string) == 0) {
+		/* Model Number String (UUID: 0x2A24) */
+		wchar_t* product_string = hid_internal_get_devnode_property(dev_node, (const DEVPROPKEY*)&PKEY_DeviceInterface_Bluetooth_ModelNumber, DEVPROP_TYPE_STRING);
+		if (!product_string) {
+			/* Get devnode grandparent to reach out Bluetooth LE device node */
+			if (CM_Get_Parent(&dev_node, dev_node, 0) != CR_SUCCESS)
+				return;
```
Youw, line 442: I'm not a fan of this early return here
let's avoid it and have the read of `DEVPKEY_NAME` under an `if`

Youw (commented):
No more comments, thanks

Youw (approved):
@DJm00n please check if I accidentally didn't break anything. I don't have an actual BLE device to check this.

DJm00n:
@Youw looks good. thanks

PR #504, opened 2023-02-11
Unified the target output directory.

hkeepl:
It's hard to run or debug the the test executable program because the library and test target are compiled in seperate directories

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,10 @@ unset(_build_type_index)
 
 project(hidapi LANGUAGES C)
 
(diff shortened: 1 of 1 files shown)


Youw:
This will not be merged.

- I have been developing/debugging HIDAPI for several years, and it hasn't been an issue for me even once;
I presume you're the Windows-only user. On other platforms that is not even required due to mechanisms like embedded RUNPATH.
- You may set those variable variables for you own CMake build yourself. I don't see reasons why this should be a default for all users.
- Once you set it here - no other users may override it w/o source modifications. We don't want to have such limitation.
- If you use an IDE with mature CMake support - it automatically adds all library paths to your PATH, so it runs/debugs just fine (e.g. I use Qt Creator Community edition).

PR #509, opened 2023-02-19
macOS: silence sprintf deprecation

Youw:
I've started getting this recently


sezero:
The rationale is wrong: snprintf has always been available in macosx, since its initial 10.1 release

sezero:
> I have to admit - I didn't check the macOS-specific background of this. Only checked that it is not available in C99 standard. If anyone provides a PR with a fix before I can get my hands to it - that would be great.

Here you go: https://github.com/libusb/hidapi/pull/511

PR #512, opened 2023-02-25
Suggestion for a clean implementation of #480.

OgreTransporter:
Update pr #507 for a clean implementation of issue #480.


Youw:
For as long as CI is not happy - this is not a solution

PR #515, opened 2023-03-12
win: Have a separate definition flag for export

Youw:
On some compilers an empty definition (e.g. -DHID_API_EXPORT) means the same as =1 (e.g. -DHID_API_EXPORT=1), but the HID_API_EXPORT macro should be empty in static library build case.

Having a separate compiler definition flag solves that issue.

Closes: #512
Relates: #480

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -29,18 +29,17 @@
 
 #include <wchar.h>
 
+/* #480: this is to be refactored properly for v1.0 */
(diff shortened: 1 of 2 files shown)


Youw:
@SamuelC0507 this should fix the issue you've faced.

SamuelC0507:
> @SamuelC0507 this should fix the issue you've faced.

Just tried it and yes indeed it does, thanks!

PR #520, opened 2023-03-27
Suggestion for new windows import/export macro handling

OgreTransporter:
#480

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -29,15 +29,24 @@
 
 #include <wchar.h>
 
-/* #480: this is to be refactored properly for v1.0 */
-#ifdef _WIN32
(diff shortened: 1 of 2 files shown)


Youw:
I understand this is a "step into a right direction", but what does it fix here?
The fact that you need to modify the Makefiles and MSVC project - means the changes breaks the build for those who desided to embedde HIDAPI directly as a source (as was originally advertized by the HIDAPI author) - [sample](https://github.com/trezor/cython-hidapi/blob/8e92053d0d7f5a5dd9bdfeb6a78191ef717cdccf/setup.py).
It is not a big deal to fix it, e.g. for cython-hidapi, but that is exactly a maintenance burden I'd like to avoid completely until v1.0.

OgreTransporter:
> I'd like to avoid completely until v1.0
Ok, I can understand that.

PR #533, opened 2023-04-26
Fix API definitions for clang/gcc flag -Wstrict-prototypes

oleh-yudin-ua:


Youw (approved):
Thanks!
Can you apply the same fix for all backends (windows/, linux/, libusb/)?

oleh-yudin-logi:
> Thanks! Can you apply the same fix for all backends (windows/, linux/, libusb/)?

sure. added the same for other backends too

Youw (approved):

PR #574, opened 2023-05-26
hidapi/mac: option to only enumerate IOHIDDevices that are likely to be joysticks

slouken:
Touching HID devices with keyboard usages will trigger a keyboard capture permission prompt on macOS 11+. See [#4887](https://github.com/libsdl-org/SDL/issues/4887)

This PR adds a compile time option (disabled by default) of only enumerating controllers.

### DIFF
--- a/mac/hid.c
+++ b/mac/hid.c
@@ -38,6 +38,11 @@
 
 #include "hidapi_darwin.h"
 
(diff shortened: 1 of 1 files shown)


slouken:
This has been folded into higher level SDL code, enabled on all platforms in https://github.com/libsdl-org/SDL/commit/0ffeca8a1ce42984759585ef3a35ecda85b64938

PR #575, opened 2023-05-26
hidapi/windows: avoid enumerating devices that may hang when queried

slouken:


windows/hid.c
```diff
@@ -756,6 +756,28 @@ static struct hid_device_info *hid_internal_get_device_info(const wchar_t *path,
 	return dev;
 }
 
+static int hid_blacklist(unsigned short vendor_id, unsigned short product_id)
```
JoergAtGithub, line 759: We use also such a denylist in our app. But we use more filter criteria: USB-Interface-Number, Usage-Page, Usage and Bus-Type

slouken: Yeah, that logic could be moved into SDL, but since we've definitely run into problems with those devices I thought it would be worth sharing upstream.

slouken:
These are mostly from Steam customer reports, running Windows. I'm not sure whether they vary based on OS or driver version. There was one device which I observed hanging on Linux, and it was the device itself that was hanging while processing a string request, not the driver. Unfortunately this was years ago and I don't remember which one it was.

slouken:
The Speedlink Competition Pro was one I observed myself, and it was the strangest behavior. The controller actually disconnected and re-enumerated as a different (non-functional) device when queried.

Youw:
What I am afraid in hard-coded lists like this is that it makes it really difficult to support and reason about, e.g.:
- what if the device/driver/OS got fixed over time and someone really needs to use it?
- what if it was never a device/driver problem (but a cable/HUB/etc.)?
- what if all of it works since Windoes 11 vXXXX, and none of it is needed anymore?
- what if all of those are valid broken devices and we should filter them out on all platforms?

I've very scepticaly of having it as part of HIDAPI implementation.

---
I'd like to hear/investigate from someone who has one of those faulty devices.

slouken:
These are all reasonable concerns. I'll go ahead and leave this in SDL and close this PR.

Youw:
We could do both: allow SDL filter out devices like this, _and_ avoid patching HIDAPI sources if we would introduce something like:
```c
typedef int (*HID_API_CALL hidapi_ven_dev_filter_func)(unsigned short vendor_id, unsigned short product_id);
struct hid_device_info HID_API_EXPORT * HID_API_CALL hid_enumerate_ven_dev_filter(hidapi_ven_dev_filter_func);
```
But that is a different effort, and would definitely need an implementation on all platforms.

slouken:
Yup, agreed. And for what it's worth the SDL filter also uses the usage page and usage, so that might be useful in this model.

Let me know if you'd like me to create a PR for this approach?

Youw:
If you're up to it - the improvements are always welcome.
You've already introduced a lot of those, thank you for that.
Would be great to finish up the existing PRs first.

Issue #237, opened 2021-01-26
Drop Windows XP support


mcuee:
From https://github.com/libusb/hidapi/issues/133.
It may be a good idea to drop Windows XP support. libusb has done that for Windows.

Youw:
Maybe.
It just don't really solve anything in context of #133. #146 - does. I've described it in [details](https://github.com/libusb/hidapi/issues/133#issuecomment-567247547).

mcuee:
Yes it does not sort the issue of #133. 

But then it may still be a good idea to upgrade from CancelIo to CancelIoEx.

Youw:
What I afraid the most - it will hide the #133 issue deep enough so some users will start thinking that there is no issue.
I always favor an explicit crash/hang/error (so it is reproducible, can be easily debugged/fixed) instead of an "undefined behavior" that _may_ hit you one in a million times.

In other words: `hid_close` better not try to interrupt `hid_read` (explicitly or implicitly) unless it is going to wait for its completion (which is not the case for any of the backends).

mcuee:
In this case, I will close this one.

mcuee:
I will keep this open even if this does not help on #133. Probably this can be considered after #133.

mcuee:
I think it is better to close this as it is clear that hidapi will support old OS when there are no extra efforts.

https://github.com/libusb/hidapi/issues/376#issuecomment-1033756841
> > does it make sense to still support Windows XP and Vista? Or old macOS versions? Or very old Linux versions?
> 
> For as long as it doesn't require the actually effort / workarounds on a code level to support those.

aitap:
When the use of [`CM_Get_DevNode_PropertyW`](https://learn.microsoft.com/en-us/windows/win32/api/cfgmgr32/nf-cfgmgr32-cm_get_devnode_propertyw) (Vista and later) has been introduced in 11cc0e8a86b33f5b760934eef32e09c55f2c8eae, it was made optional. In 2551a5d84baca323ca8d61db06464cd1b963c289, failure to resolve `CM_Get_DevNode_PropertyW` becomes a fatal error, which effectively drops Windows XP support.

This is not a complaint (though if you're interested, I could submit a pull request); I'm just providing information for the sake of someone else whose code will fail to initialise on an old machine and who will go searching for "Windows XP" in this repo.

Youw:
Thanks for observation.
If you're interested in Windows XP support and can test it - a PR with a fix is welcome.

[...]

Issue #146, opened 2020-02-09
add hid_interrupt_read


Youw:
add new API function: `int hid_interrupt_read(hid_device *dev)`

it will asynchronously interrupt blocking `hid_read` call, if any

details: TBD

mcuee:
Any activities on this one?

Youw:
not that I'm aware of

mcuee:
> not that I'm aware of

No problem. We can always delay this after 0.13 release.

Issue #133, opened 2019-12-17
Windows does not support multi-threading


FrogTheFrog:
Currently hidapi is not able to close open device handle in a thread-safe manner. This is due to `CancelIo` function being used instead of `CancelIoEx` for closing a win32 device handle.

Upgrading to `CancelIoEx` would cause a WinXP support to be dropped (https://github.com/signal11/hidapi/issues/48), however a fallback could be used to try and load `CancelIoEx` dynamically to keep a somewhat buggy WinXP experience (https://github.com/signal11/hidapi/pull/208).

Something needs to be done about this issue as hidapi is being used in multi-threaded environment more and more. I would personally propose to drop WinXP support altogether as there is no point in beating a dead horse, however I would like to hear what you guys think.

Youw:
I don't think it is safe to assume `hid_close` as a thread-safe function at all.

Any client code has to stop any read/write attempts before calling `hid_close`.

What do I miss here?

FrogTheFrog:
Let's say you have 2 threads (both are using some kind of wrapper to make sure that access to hidapi is synchronized via mutex or something else). First one is doing some thread blocking read operation which puts whole thread on sleep, and the second one is trying to abort this read. The only way to abort is via `hid_close` in second thread.

Currently `hid_close` uses `CancelIo` to abort this read operation in the first thread, however `CancelIo` can only be used in the same thread that started the read operation. Thus, we are stuck. Most users don't care and call `hid_close` on second thread anyway, which sometimes crashes. `CancelIoEx` supports the scenario I've just described.

Youw:
here is a pseudo-code, that works without issues and flexible for various use-cases:
thread1:
```cpp
while(!interrupted) {
  result = hid_read_timeout(dev, buffer, buffer_size, 1000);
  if (result == 0) // no data available
    continue;
  if (result > 0) {
    // handle data here
  }
  else {
    // handle/report error, retry, etc.
  }
}
```
thread2:
```cpp
interrupted = true;
thread1.join();
// now we're sure no other threads are trying to access the device,
// assuming there is no threads, except thread1
hid_close(dev);
```

---
such usage conforms to my previous statement, that client code should first stop any read/write attempts before trying to close the device as `hid_close` cannot not be thread-safe by design, even it it is _sometime_, for _some implementations_

Youw:
[...]

PR #576, opened 2023-05-26
hidapi/windows: fix use-after-free SBH corruption due to overlapped ReadFile in hidapi not being canceled for all threads before device close

slouken:
- hidapi already called CancelIo on hid_close but that only cancels pending IO for the current thread. Controller read/writes originate from multiple threads (serialized, but on a different thread nonetheless) but device destruction was always done on the main device thread which left any pending overlapped reads still running after hidapi's internal read buffer is deallocated leading to intermittent free list corruption.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -1273,14 +1273,33 @@ int HID_API_EXPORT HID_API_CALL hid_get_input_report(hid_device *dev, unsigned c
 	return hid_get_report(dev, IOCTL_HID_GET_INPUT_REPORT, data, length);
 }
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wcast-function-type"
+#endif
 void HID_API_EXPORT HID_API_CALL hid_close(hid_device *dev)
 {
+	typedef BOOL (WINAPI *CancelIoEx_t)(HANDLE hFile, LPOVERLAPPED lpOverlapped);
+	CancelIoEx_t CancelIoExFunc = (CancelIoEx_t)GetProcAddress(GetModuleHandle(TEXT("kernel32.dll")), "CancelIoEx");
+
 	if (!dev)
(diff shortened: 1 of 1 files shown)


mcuee:
@slouken 

There is a warning treated as error so that Windows-cmake github action build failed. Please help to fix it. Thanks.

Youw:
calling `hid_close` while some other thread is calling any other device-specific hidapi function is undefined behavior, even with this fix.

I've described this problem in details here: https://github.com/libusb/hidapi/issues/133#issuecomment-567247547

There is also a related discussion: https://github.com/libusb/hidapi/issues/237

Youw:
The only long-going blocking call is `hid_read`, and I've [described](https://github.com/libusb/hidapi/issues/133#issuecomment-567060164) the only functional and strictly correct workaround for that case.

slouken:
We're using the correct model that you're describing above. The problem here is that we're doing a read with a timeout of 0 on one thread, which can start an overlapping I/O operation. The reading thread then exits, but the overlapping I/O operation is still pending, then the main thread calls hid_close(), then the OS completes the I/O operation, writing to the memory that was just freed. It doesn't happen often, but it is something that was causing crashes for customers here.

Youw:
> We're using the correct model that you're describing above.

>  with a timeout of 0

Those two statements contradict each other. You actually cannot use the `0` timeout for `hid_read_timeout` until https://github.com/libusb/hidapi/issues/146 is implemented.
In which case you'd need:
- call `hid_interrupt_read`
- wait for the read thread to signal that `hid_read/_timeout` is no longer in use (need to use additional syncronisation mechanism for that)
- only then call `hid_close`

> then the main thread calls hid_close()

That is undefined behavior even with this fix. The hid_close may successfully interrupt the IO operation, then free the `dev` object in memory, while `hid_read` is still trying to access (some of) its fileds.

Issue #513, opened 2023-02-26
Windows: Bluetooth and WriteFile API issues


DJm00n:
Looks like Windows Bluetooth drivers have issues with sending output reports with `WriteFile`.
Need to investigate this and try to use `HidD_SetOutputReport` as a fallback.

https://stackoverflow.com/questions/29480144/hidapi-sending-packet-smaller-than-caps-outputreportbytelength
https://social.msdn.microsoft.com/Forums/ie/en-US/834d2cf2-672d-424b-a126-dbc69f3f75a2/hiddsetoutputreportwritefile-bluetooth-low-energy-windows-81

Related https://github.com/libsdl-org/SDL/issues/7224

DJm00n:
>The GATT Read Characteristic Value or Read Long Characteristic Value subprocedures are used to read a Report characteristic containing Output Report data. This procedure maps to a Get_Report (Output) request in USB HID [2]. 
>The GATT Write Characteristic Value sub-procedure is used to write to a Report characteristic containing Output Report data. This procedure maps to a Set_Report (Output) request in USB HID [2]. The GATT Write Without Response sub-procedure is 
also used to write to a Report characteristic containing Output Report data, and this procedure maps to Data Output in USB HID [2]. 

https://www.bluetooth.com/specifications/specs/human-interface-device-service-1-0/

DJm00n:
>HidD_SetOutputReport becomes IOCTL_HID_SET_OUTPUT_REPORT
>WriteFile becomes IOCTL_HID_WRITE_REPORT

https://stackoverflow.com/questions/63115984/what-is-the-difference-between-read-report-get-input-report-write-report

>Output items make Output reports accessible via the Control pipe with a Set_Report (Output) command. 
>Output type reports can optionally be sent via an Interrupt Out pipe.

>The Control pipe is used for: ... Transmitting data when polled by the HID class driver (using the Get_Report 
request). 


>The Interrupt Out pipe is optional. If a device declares an Interrupt Out endpoint then Output reports are transmitted by the host to the device through the Interrupt Out endpoint. If no Interrupt Out endpoint is declared then Output reports are transmitted to a device through the Control endpoint, using Set_Report(Output) requests. 

https://www.usb.org/document-library/device-class-definition-hid-111

Youw:
The later only proves that we cannot unconditionally replace `WriteFile` with `SetOutputReport`, I guess.

[...]

PR #578, opened 2023-05-26
hidapi/windows: fixed PS4 controllers over Bluetooth on Windows 7

slouken:


windows/hid.c
```diff
@@ -1003,6 +1030,16 @@ HID_API_EXPORT hid_device * HID_API_CALL hid_open_path(const char *path)
 	return dev;
 }
 
+static int hid_write_output_report(hid_device *dev, const unsigned char *data, size_t length)
+{
+	BOOL res;
+	res = HidD_SetOutputReport(dev->device_handle, (void *)data, (ULONG)length);
```
Youw, line 1036: please see https://github.com/libusb/hidapi/pull/224
I'm confident we need to do the same thing for `HidD_SetOutputReport` as well.

Youw:
This looks similar to a problem described in https://github.com/libusb/hidapi/issues/513.
@DJm00n what do you think?

mcuee:
@DJm00n 

Sorry to ping.

windows/hid.c
```diff
@@ -1003,6 +1030,16 @@ HID_API_EXPORT hid_device * HID_API_CALL hid_open_path(const char *path)
 	return dev;
 }
 
+static int hid_write_output_report(hid_device *dev, const unsigned char *data, size_t length)
```
DJm00n, line 1033: I'd prefer `hid_internal_write_output_report` - so it is easily distinguishable from public method

windows/hid.c
```diff
@@ -183,8 +185,28 @@ struct hid_device_ {
 		OVERLAPPED ol;
 		OVERLAPPED write_ol;
 		struct hid_device_info* device_info;
+		BOOL use_hid_write_output_report;
 };
 
+static BOOL IsWindowsVersionOrGreater(WORD wMajorVersion, WORD wMinorVersion, WORD wServicePackMajor)
```
DJm00n, line 191: I'd prefer `hid_internal_is_windows_version_or_greater` - so it is easily distinguishable from public method

DJm00n (commented):
Overall looks good

slouken:
Okay, I've updated the PR with your suggestions.

DJm00n (approved):

windows/hid.c
```diff
@@ -1003,6 +1030,16 @@ HID_API_EXPORT hid_device * HID_API_CALL hid_open_path(const char *path)
...
 }
 
+static int hid_internal_write_output_report(hid_device *dev, const unsigned char *data, size_t length)
+{
+	BOOL res;
+	res = HidD_SetOutputReport(dev->device_handle, (void *)data, (ULONG)length);
+	if (res)
+		return (int)length;
+	else
+		return -1;
```
Youw, line 1040: either here or at the place of usage (in case of error) we need to register an error string

windows/hid.c
Youw: @slouken any thoughs/comments?

windows/hid.c
slouken: Yep, good point. Added!

windows/hid.c
JoergAtGithub: @slouken It seems, you missed to respond to this question

slouken: It looks like #224 was just pre-allocating the output buffer. Were you referring to another change that made sure the buffer size matched feature_report_length? I think, but I'm not sure, that the feature_report_length is the wrong size in this case.

Youw: I mean that the buffer size has to be at least `output_report_length` bytes in size.
If user passes less data - logically it is not a proble, but WinAPI implementation is known to crash if bufer is too small.
The workaround (as in #224) is to pre-allocate buffer that is large-enough _if needed_.

mcuee:
Now that Windows 7 has been out of support for a while, I am not so sure if we still care about Windows 7 or not.

mcuee:
On the other hand, #513 does not seem to be related to Windows 7.
* https://github.com/libusb/hidapi/issues/513

PR #580, opened 2023-05-26
hidapi/linux: fixed crash if uevent info isn't available

slouken:


Youw (approved):
I wounder how come no one (e.g. Coverity) didn't catch this earlier.
Thanks.

Issue #581, opened 2023-05-26
Building libusb support on Windows and macOS


slouken:
SDL has a change to allow building hidapi with libusb support on Windows. It was done by abstracting the threading interface, so it can be replaced by other implementations, and the pthread implementation is provided by default.

Please let me know if you'd like this change as a PR!
https://github.com/libsdl-org/SDL/commit/57b33aec01a86c28b30797ecbae394a2f07ddd42

mcuee:
Yes please file a PR.

This will also address the following issue.
* https://github.com/libusb/hidapi/issues/319

mcuee:
@slouken 

If possible, you may also want to add macOS support. macOS libusb support is not so useful for real application (requires sudo or some difficult-to-attain entitlement), but it may be useful to troubleshooting purpose and maybe useful for some devices.

slouken:
The current code works on macOS without any changes. I tested with it a few days ago.

PR #582, opened 2023-05-27
libusb: Allow building using non-POSIX runtime

slouken:
This isn't a full Windows or macOS port, but abstracts the libusb threading code so that it can be easily ported to other platforms.

Fixes https://github.com/libusb/hidapi/issues/581

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -37,7 +37,6 @@
 #include <sys/ioctl.h>
 #include <sys/utsname.h>
 #include <fcntl.h>
-#include <pthread.h>
 #include <wchar.h>
 
 /* GNU / LibUSB */
@@ -51,66 +50,10 @@
 
 #include "hidapi_libusb.h"
 
-#if defined(__ANDROID__) && __ANDROID_API__ < __ANDROID_API_N__
-
-/* Barrier implementation because Android/Bionic don't have pthread_barrier.
-   This implementation came from Brent Priddy and was posted on
-   StackOverflow. It is used with his permission. */
-typedef int pthread_barrierattr_t;
-typedef struct pthread_barrier {
-    pthread_mutex_t mutex;
-    pthread_cond_t cond;
-    int count;
-    int trip_count;
-} pthread_barrier_t;
-
-static int pthread_barrier_init(pthread_barrier_t *barrier, const pthread_barrierattr_t *attr, unsigned int count)
-{
-	if(count == 0) {
-		errno = EINVAL;
-		return -1;
-	}
-
-	if(pthread_mutex_init(&barrier->mutex, 0) < 0) {
-		return -1;
-	}
-	if(pthread_cond_init(&barrier->cond, 0) < 0) {
-		pthread_mutex_destroy(&barrier->mutex);
-		return -1;
-	}
-	barrier->trip_count = count;
-	barrier->count = 0;
-
-	return 0;
-}
-
-static int pthread_barrier_destroy(pthread_barrier_t *barrier)
-{
-	pthread_cond_destroy(&barrier->cond);
-	pthread_mutex_destroy(&barrier->mutex);
-	return 0;
-}
-
-static int pthread_barrier_wait(pthread_barrier_t *barrier)
-{
-	pthread_mutex_lock(&barrier->mutex);
-	++(barrier->count);
-	if(barrier->count >= barrier->trip_count)
-	{
-		barrier->count = 0;
-		pthread_cond_broadcast(&barrier->cond);
-		pthread_mutex_unlock(&barrier->mutex);
-		return 1;
-	}
-	else
-	{
-		pthread_cond_wait(&barrier->cond, &(barrier->mutex));
-		pthread_mutex_unlock(&barrier->mutex);
-		return 0;
-	}
-}
-
+#ifndef HIDAPI_THREAD_MODEL_INCLUDE
+#define HIDAPI_THREAD_MODEL_INCLUDE "hidapi_thread_pthread.h"
 #endif
+#include HIDAPI_THREAD_MODEL_INCLUDE
 
 #ifdef __cplusplus
 extern "C" {
@@ -168,10 +111,7 @@ struct hid_device_ {
 	int blocking; /* boolean */
 
 	/* Read thread objects */
-	pthread_t thread;
-	pthread_mutex_t mutex; /* Protects input_reports */
-	pthread_cond_t condition;
-	pthread_barrier_t barrier; /* Ensures correct startup sequence */
+	hidapi_thread_state thread_state;
 	int shutdown_thread;
 	int transfer_loop_finished;
 	struct libusb_transfer *transfer;
@@ -201,9 +141,7 @@ static hid_device *new_hid_device(void)
 	hid_device *dev = (hid_device*) calloc(1, sizeof(hid_device));
 	dev->blocking = 1;
 
-	pthread_mutex_init(&dev->mutex, NULL);
-	pthread_cond_init(&dev->condition, NULL);
-	pthread_barrier_init(&dev->barrier, NULL, 2);
+	hidapi_thread_state_init(&dev->thread_state);
 
 	return dev;
 }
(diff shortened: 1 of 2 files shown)


mcuee:
I think CMake files need to be updated as well.

slouken:
I don't think so, it work as-is...?

mcuee:
> I don't think so, it work as-is...?

If you look at the CMake file, I think you need to add options to build both native Windows version and Windows libusb version, same for native macOS version and macOS libusb versoon.
https://github.com/libusb/hidapi/blob/d3013f0af3f4029d82872c1a9487ea461a56dee4/CMakeLists.txt#L41-L44
```
    if(CMAKE_SYSTEM_NAME MATCHES "Linux")
        option(HIDAPI_WITH_HIDRAW "Build HIDRAW-based implementation of HIDAPI" ON)
        option(HIDAPI_WITH_LIBUSB "Build LIBUSB-based implementation of HIDAPI" ON)
    endif()
```

slouken:
The macOS change can be made as a separate PR.
The Windows change requires a Windows threads implementation. I used SDL, but that's not applicable here, so that would be a separate change.

mcuee:
No problem, you can create a new PR for macOS. Thanks.

mcuee:
@slouken 

BTW, just FYI, there is a bug with the libusb backend which is difficult to fix. It may not affect many devices though.
* https://github.com/libusb/hidapi/issues/274

mcuee:
@slouken 

I'd like to test this PR but unfortuantely I do not see how to build this PR without the changes to CMake or the legacy auto-tools script. Please help to explain how you build this PR. Thanks.

windows/hid.c
```diff
@@ -56,9 +56,11 @@ typedef LONG NTSTATUS;
 #include "hidapi_hidclass.h"
 #include "hidapi_hidsdi.h"
 
+#ifndef HIDAPI_USING_SDL_RUNTIME
```
JoergAtGithub, line 59: This change looks unrelated

Youw:
My guess if that SDL has something like:
`sdl_libusb_hid.c`:
```c
typedef struct
{
//...
} hid_device_thread_state;
// more thread_state defines
...
#define HIDAPI_USING_SDL_RUNTIME
#include "hidapi/libusb/hid.c"
```
am I correct on the approach this is used?

Youw:
While https://github.com/libsdl-org/SDL/commit/57b33aec01a86c28b30797ecbae394a2f07ddd42 makes perfect sense in the context of SDL project, this change is not at its best in the context of HIDAPI project.

First of all: the "SDL" term shouldn't be used among HIDAPI code.
Secondly: if we're introducing a threading abstraction - we better explicitly do so.

What I suggest:
- lets introduce explicit naming: compared to the original code, replace `pthread_` part with `hidapi_thread_`, so we would get:
    - `hidapi_thread_state_init`
    - `hidapi_thread_mutex_lock`
    - etc.
- introduce a new source file: `libusb/hidapi_thread_pthread.h` with the includes/declarations/definitions for pthread implementation;
- to get keep backward compatibility _and_ extensibility, inside `libusb/hid.c `:
    ````c
    #ifndef HIDAPI_THREAD_MODEL_INCLUDE
    #define HIDAPI_THREAD_MODEL_INCLUDE "hidapi_thread_pthread.h"
    #endif
    #include HIDAPI_THREAD_MODEL_INCLUDE
    ````
    and use the functions below as suggested.

It would also greatly simplify making a native HIDAPI port of libusb backend on Windows.

Youw:
> I think CMake files need to be updated as well.

This change doesn't actually make it possible to build libusb backend of HIDAPI on Windows, but rather it is just a nesessary refactoring to allow replacing pthread with any other threading library.

I'm generally agree with the aproach, just have a few conserns regarding implementation (my prev comment).

slouken:
I've updated the pull request and scoped it just to create the threading abstraction.

libusb/hid.c
```diff
@@ -1128,15 +1064,15 @@ static int hidapi_initialize_device(hid_device *dev, int config_number, const st
...
 
 	/* Wait here for the read thread to be initialized. */
-	pthread_barrier_wait(&dev->barrier);
+	hidapi_thread_barrier_wait(&dev->thread_state);
 	return 1;
 }
 
 
-hid_device * HID_API_EXPORT hid_open_path(const char *path)
+HID_API_EXPORT hid_device *hid_open_path(const char *path)
```
Youw, line 1075: this change is unrelated

libusb/hidapi_thread_pthread.h
```diff
@@ -0,0 +1,163 @@
+/*******************************************************
+ HIDAPI - Multi-Platform library for
+ communication with HID devices.
+
+ Alan Ott
+ Signal 11 Software
+
+ libusb/hidapi Team
+
+ Copyright 2022, All Rights Reserved.
```
Youw, line 10:
```suggestion
 Alan Ott
 Signal 11 Software

 libusb/hidapi Team
 
 Sam Lantinga

 Copyright 2023, All Rights Reserved.
```

libusb/hidapi_thread_pthread.h
```diff
@@ -0,0 +1,163 @@
...
+	pthread_mutex_destroy(&barrier->mutex);
+	return 0;
+}
+
+static int pthread_barrier_wait(pthread_barrier_t *barrier)
+{
+	pthread_mutex_lock(&barrier->mutex);
+	++(barrier->count);
+	if(barrier->count >= barrier->trip_count)
+	{
```
Youw, line 70: I understand this is moved as is, but lets introduce some consistency: next line for functions, same line for statements
```suggestion
	if(barrier->count >= barrier->trip_count) {
```

libusb/hidapi_thread_pthread.h
```diff
@@ -0,0 +1,163 @@
...
+		barrier->count = 0;
+		pthread_cond_broadcast(&barrier->cond);
+		pthread_mutex_unlock(&barrier->mutex);
+		return 1;
+	}
+	else
+	{
```
Youw, line 77:
```suggestion
	else {
```

libusb/hidapi_thread_pthread.h
```diff
@@ -0,0 +1,163 @@
...
+
+static void hidapi_thread_join(hidapi_thread_state *state)
+{
+	pthread_join(state->thread, NULL);
+}
+
+static void hidapi_thread_gettime(struct timespec *ts)
+{
+	clock_gettime(CLOCK_REALTIME, ts);
+}
```
Youw, line 162: interesting, this is not exactly related to threading model
more to platform itself, (i.e. POSIX/non-POSIX)
but I guess it is fine to keep it here for now as exception

but lets introduce an alias for the timespec type:
`#define hidapi_timespec timespec` to avoid any possible name collisions

Youw (approved):
Perfect, just a few small comments.

slouken:
Okay, I think this addresses everything. Please let me know if you have any other suggestions!

libusb/hidapi_thread_pthread.h
```diff
@@ -0,0 +1,174 @@
...
+{
+	pthread_join(state->thread, NULL);
+}
+
+static void hidapi_thread_gettime(hidapi_timespec *ts)
+{
+	clock_gettime(CLOCK_REALTIME, ts);
+}
+
+static void hidapi_thread_addtime(hidapi_timespec *ts, int milliseconds)
```
Youw, line 166: oh, I haven't though of this one, but great suggestion, thanks

Youw (approved):
Great work!

Now it should be relatively straight-forward to implement the native Windows port of libusb backend.
I'll keep this change in review for a few more days, so the comunity has a chance to catch up/review.

libusb/hidapi_thread_pthread.h
```diff
@@ -0,0 +1,174 @@
...
+
+static void hidapi_thread_join(hidapi_thread_state *state)
+{
+	pthread_join(state->thread, NULL);
+}
+
+static void hidapi_thread_gettime(hidapi_timespec *ts)
+{
+	clock_gettime(CLOCK_REALTIME, ts);
+}
```
JoergAtGithub, line 164: I wonder if it would make sense to seperate the two time functions into an own entity like `HIDAPI_TIME_REF_INCLUDE`.
This would allow to use a application specific time reference for future features like time stamping of incoming reports.
Other time references could be a the soundcard clock in an audio application or the network clock in a streaming application.

Youw: As of this moment (and for many years before that) the only place where this time related thing is used - is for `thread_cond_timedwait`.
As of this moment - this looks compact and we avoid additional abstraction/implementation.
If we need an abstraction in the future - we can easily introduce it later.

PR #583, opened 2023-05-27
Fixed double-free during multi-threaded hidapi access

slouken:
The error string is not protected by a mutex, and can be set from multiple threads at the same time. Without this change, it can be double-freed. It can still be double-allocated, leading to a memory leak, but at least it won't crash now.


slouken:
This isn't the right approach. Even if the string itself isn't double-freed, it can be returned to the application and then freed while the application is trying to use it. This really needs to be in thread-local storage to be completely safe.

In SDL we already have a global thread-local error string, so I'm going to make an SDL-specific change to handle the error strings safely.

Youw:
This is one of the reasons why HIDAPI is claimed to be not thread-safe.

PR #602, opened 2023-07-25
Fixed Visual C++ warning C4214

slouken:


Youw:
What is the original warning?
In theory I might understand why change ULONG to UINT, but BOOLEAN to UINT is not as traight-forward.

That struct has to be 100% ABI stable. It is reversed-engineered, based on some pieces found online.

Please see that the CI build is failing.

I'd consider explicitly supressing some warnings, instead of changing types as suggested in this PR.

slouken:
Investigating test failures...

slouken:
Okay, I'll make that change instead. I think the change of BOOLEAN to UINT may be incorrect here.

Youw (approved):

Youw:
thanks

Issue #607, opened 2023-08-05
Broken Archlinux build


Youw:
Sample: https://github.com/libusb/hidapi/actions/runs/5770941461/job/15644421387
It is either broken/incompatible docker image (so just wait/re-check after shortwhile, until it is updated/fixed), or need to switch from `archlinux:latest` to something else (`archlinux:base`?).
Or some package needs to be installed/updated in runtime after one of the recent image updates.

PR #610, opened 2023-08-14
CI: Fix Archlinux build

Youw:
Apparently when installing/upgrading packages on Archlinux - need to explicitly make sure `glibc` is upgraded as well.

Fixes: #607

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -400,7 +400,7 @@ jobs:
     - name: Install dependencies
       run: |
         pacman -Sy
-        pacman -S --noconfirm gcc pkg-config autoconf automake libtool libusb libudev0 cmake make
+        pacman -S --noconfirm glibc lib32-glibc gcc pkg-config autoconf automake libtool libusb libudev0 cmake make
     - name: Configure CMake
       run: |
         rm -rf build install


mcuee:
Interesting to know about this. 

I usually run Ubuntu LTS version (20.04 and 22.04) as well Debain (11/12) or distros based on them for ARM SBCs (Raspberry Pi OS, Ubuntu for ARM64, or Armbian based on Debian/Ubuntu). 

Recently I played with Arch Linux and a few Arch Linxu based distros as VMs under my Intel J4105 mini PC running Proxmox Virtual Environment 7.4. I tried the build test of libusb/hidapi/libftdi/avrdude and I did not encounter issues. But I have never used an Arch Linux based container, my LXC containers are again mainly using Ubuntu and Debain.

Youw:
Apparently what was happening here:

- we use some docker image which has some packaged (pre)installed, like `glibc`;
- by `pacman -Sy` we're updating a list of packages available on _the remote_; but it does not upgrade existing packages;
- we install a bunch of packanges, including `cmake` which depends on `libstdc++` which happens to depend on a never version of `glibc` than we had preinstalled, but does not upgrade the existing package (i believe `apt` wouldn't allow installing such package);

Youw:
The "before" used to work, only because until recently the very latest version of `libstdc++` was depending on the same (or older) version of `glibc` that was already preinstalled on the image.

Issue #587, opened 2023-06-09
Mac: hid_enumerate crash on macOS 14 when linked to AppKit and inside a `fork`


nguyen-phillip:
The example code below crashes in macOS 14 on a call to IOHIDManagerCreate when linked against AppKit and the --fork flag is passed. Removing the AppKit dependency allows it to run as expected. I've filed a case with Apple since this seems to be an OS bug.

However it seems that the only reason why the hidapi code needs to be linked against AppKit is because it references NSAppKitVersionNumber to check for if running on macOS >= 10.10 Yosemite

https://github.com/libusb/hidapi/blob/80e8d62bc1cc70876d57fed7e76d7478757b4da8/mac/hid.c#L469

Would it be reasonable to send a pull request that dropped support for macOS < 10.10? This would simplify the code a bit and remove the requirement for linking with AppKit entirely.

```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <hidapi.h>

int main(int argc, char **argv) {
  if (argc > 1 && !strcmp(argv[1], "--fork")) {
    pid_t pid = fork();
    printf("fork: %d\n", pid);
    if (pid != 0) {      
      printf("parent exiting\n");
      exit(0);
    }
  }  
  struct hid_device_info *info = hid_enumerate(0, 0);
  printf("hid_device_info: %lx\n", (long)info);
  return 0;
}
```

Youw:
I don't think I have enough info to tell if we should drop the 10.10 support

can we instead have compile-time desision based on the `-mmacosx-version-min` compilation flag (I'm not sure if there is corresponding compile-time macro definition - I presume there is)
that should be both safe and backward compatible

nguyen-phillip:
Maybe something like

```
is_macos_10_10_or_greater = MAC_OS_X_VERSION_MIN_REQUIRED > MAC_OS_X_VERSION_10_10;
```

would work?

Youw:
Unfortunately no. It has to be more like:
```c
is_macos_10_10_or_greater = 
#if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_10
    true;
#else
    (NSAppKitVersionNumber >= 1343);
#endif
```
and the declaration of `NSAppKitVersionNumber` has to be under similar `ifdef`
so is the link library in `CMakeList.txt`

nguyen-phillip:
[...]

PR #611, opened 2023-08-15
macOS: Check for macOS 10.10 using kCFCoreFoundationVersionNumber instead of AppKit

mrpippy:
Fixes: #587

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,7 +74,7 @@ case $host in
 	backend="mac"
 	os="darwin"
 	threads="pthreads"
-	LIBS="${LIBS} -framework IOKit -framework CoreFoundation -framework AppKit"
+	LIBS="${LIBS} -framework IOKit -framework CoreFoundation"
 	;;
 *-freebsd*)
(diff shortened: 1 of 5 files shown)


Youw:
This looks perfect.
@nguyen-phillip can you confirm this fixes the issue in your environment?

nguyen-phillip:
Yes, this should work -- the issue only manifests when linked against AppKit. Thanks!

Youw:
> should work

Would you be able to actually run and check?
Thanks.

nguyen-phillip:
We've patched out this version check entirely as we don't need to support macOS predating 10.10. However the minimal repro code for the (still open) Apple bug only has problems when AppKit is linked in. CoreFoundation and IOKit are fine.

Youw (approved):
I wasn't able to reproduce the original #587 with current master. Maybe because I'm using M2 mac machine. Thus I cannot confirm if it really fixes the original issue or not.

@mrpippy can you confirm that you had the original issue reproduced locally and not reproducible after this change?

In any case - this make one dependency less compared to "before" which is a great thing on its own.

Issue #608, opened 2023-08-12
Native BSD backend using uhid


ghost:
I was curious about the pragmatics of supporting various BSDs. There seems to be a fairly common API for HID over USB on most common BSDs called `usbhid` which is intended to be used with the `uhid` kernel driver for generic HID devices. Support for this is likely a necessity because several BSD ports of `libusb` cannot be used with USB devices already reserved by a kernel driver.

I'm too unfamiliar with HID myself to work on this at present, but is there anyone else willing to put in the effort to add this as a `hidapi` backend?

Youw:
Duplicate of #543 ?

braewoods:
> Duplicate of #543 ?

I don't believe so. There's more BSDs than just FreeBSD. And #543 is about the `hidraw` compatibility layer that is only on FreeBSD. This is about the HID layer that exists on multiple BSDs, though they have diverged a bit over the years. It may be feasible to manage that with some clever use of the preprocessor.

Youw:
For those who potentially want to start implementing this but don't know where to start (from HIDAPI POV) - here is an example me adding an experimental WinRT backend: https://github.com/libusb/hidapi/compare/winrt

braewoods:
Some relevant links for `uhid`:

https://man.openbsd.org/uhid.4
https://man.netbsd.org/uhid.4
https://man.freebsd.org/cgi/man.cgi?query=uhid&sektion=4
https://leaf.dragonflybsd.org/cgi/web-man?command=uhid&section=4

braewoods:
I don't believe a native port for DragonFly is feasible unless they fix their UHID ioctl situation. It doesn't appear to support getting any information about the device behind the UHID device such as the USB device information. The other 3 still appear to be viable.

mcuee:
I am not so sure if there is any interest to have the uhid backend for FreeBSD as the existing libusb backedn works fine (libusb is part of FreeBSD now, and it is using FreeBSD's own libusb implementation with BSD license). hidraw backend seems to be more interesting for FreeBSD.

But I agree that it will be good to have uhid backend for NetBSD and OpenBSD, if it is possible. HIDAPI libusb backend for NetBSD/OpenBSD require rebuilding the kernel last time I checked, so it is not really that useful.

braewoods:
[...]

PR #612, opened 2023-08-17
Initial NetBSD/UHID native backend implementation

braewoods:
This is the culmination of the research I did for `uhid` on NetBSD. It should be as complete as a native backend can be given the limitations of NetBSD's kernel and `uhid` as a whole. See related issue, #608.


CMakeLists.txt
```diff
@@ -42,6 +42,9 @@ elseif(NOT WIN32)
         option(HIDAPI_WITH_HIDRAW "Build HIDRAW-based implementation of HIDAPI" ON)
         option(HIDAPI_WITH_LIBUSB "Build LIBUSB-based implementation of HIDAPI" ON)
     endif()
+    if(CMAKE_SYSTEM_NAME MATCHES "NetBSD")
```
Youw, line 45: Just to double-check: is `uhid` backend specific to NetBSD or is it available on other BSD systems?

braewoods: Technically it is available to other BSD systems since it has existed since sometime in 1990. But they have diverged heavily over the years so they are all at varying levels of compatibility not to mention differences in internal kernel behavior. The compatible subset I found is extremely limited and would not make for a very useful generic approach. I thought it best to combine `uhid` with platform specific support code to help workaround its limitations. NetBSD probably has the most advanced `uhid` implementation in terms of overall features.

Youw: Thanks, this is helpfull.
I agree, better keep it for NetBSD only.

braewoods:
Minor code cleanup. I noticed some improvements I could make when reusing some of the code for another project of mine.

mcuee:
Nice, the first simple test seems to work fine. I need to modify hidapitester to test this more.
```
localhost$ cmake -B build -D HIDAPI_BUILD_HIDTEST=1
-- The C compiler identification is GNU 7.5.0
-- Check for working C compiler: /usr/bin/cc - skipped
-- hidapi: v0.14.0
-- Looking for pthread.h
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Found PkgConfig: /usr/pkg/bin/pkg-config (found version "1.9.4")
-- Checking for module 'libusb-1.0>=1.0.9'
--   Found libusb-1.0, version 1.0.24
-- Check for Iconv
-- Performing Test Iconv_IS_BUILT_IN
-- Performing Test Iconv_IS_BUILT_IN - Success
-- Found Iconv: built in to C library
-- Performing Test HIDAPI_ICONV_CONST
-- Performing Test HIDAPI_ICONV_CONST - Success
-- Building hidtest
-- Configuring done
-- Generating done
-- Build files have been written to: /home/mcuee/build/hidapi_netbsd/build
localhost$ cmake --build build
[ 16%] Building C object src/netbsd/CMakeFiles/hidapi_netbsd.dir/hid.c.o
[ 33%] Linking C shared library libhidapi-netbsd.so
[ 33%] Built target hidapi_netbsd
[ 50%] Building C object src/libusb/CMakeFiles/hidapi_libusb.dir/hid.c.o
[ 66%] Linking C shared library libhidapi-libusb.so
[ 66%] Built target hidapi_libusb
[ 83%] Building C object hidtest/CMakeFiles/hidtest.dir/test.c.o
[100%] Linking C executable hidtest
[100%] Built target hidtest
localhost$ sudo ./build/hidtest/hidtest
Password:
hidapi test/example tool. Compiled with hidapi version 0.14.0, runtime version 0.14.0.
Compile-time version matches runtime version of hidapi.

Device Found
  type: 0925 7001
  path: uhidev0
  serial_number:
  Manufacturer: Microchip Technology Inc.
  Product:      Generic HID
  Release:      1
  Interface:    -1
  Usage (page): 0x1 (0xffa0)
  Bus type: 1 (USB)

  Report Descriptor: (47 bytes)
0x06, 0xa0, 0xff, 0x09, 0x01, 0xa1, 0x01, 0x09, 0x03, 0x15,
0x00, 0x26, 0xff, 0x00, 0x75, 0x08, 0x95, 0x3f, 0x81, 0x02,
0x09, 0x04, 0x15, 0x00, 0x26, 0xff, 0x00, 0x75, 0x08, 0x95,
0x3f, 0x91, 0x02, 0x09, 0x05, 0x15, 0x00, 0x26, 0xff, 0x00,
0x75, 0x08, 0x95, 0x3f, 0xb1, 0x02, 0xc0,
unable to open device

localhost$ uname -a
NetBSD localhost 9.3 NetBSD 9.3 (GENERIC) #0: Thu Aug  4 15:30:37 UTC 2022  mkrepro@mkrepro.NetBSD.org:/usr/src/sys/arch/amd64/compile/GENERIC amd64

localhost$ sudo usbdevs
addr 1: UHCI root hub, NetBSD
addr 0: xHCI root hub, NetBSD
addr 0: xHCI root hub, NetBSD
 addr 1: Generic HID, Microchip Technology Inc.
```

braewoods:
> Nice, the first simple test seems to work fine. I need to modify hidapitester to test this more.
> 
> I am using NetBSD 9.3 VM running under Proxmox and I attach a generic HID device to the VM.

Feel free to refine it further but I have reached the limit of what I can see to do. I am considering writing a FreeBSD backend after NetBSD has been merged. It is probably the last native BSD backend that has enough kernel API functionality to work with the current library abstraction. I may end up creating two versions for FreeBSD due to it having the option of using a `hidraw` compatibility layer as well as the older `uhid` one.

mcuee:
> I am considering writing a FreeBSD backend after NetBSD has been merged. It is probably the last native BSD backend that has enough kernel API functionality to work with the current library abstraction. I may end up creating two versions for FreeBSD due to it having the option of using a `hidraw` compatibility layer as well as the older `uhid` one.

That would be wonderful. I tend to think the 1hidraw` backend is the way to go for FreeBSD.

braewoods:
> That would be wonderful. I tend to think the 1hidraw` backend is the way to go for FreeBSD.
Preliminary investigation suggests this has its own problems due to how new this backend is. It isn't enabled by default on fresh installations. Of the two, `uhid` and `hidraw`, only one can be attached to any particular device.

`uhid` is the default if `hw.usb.usbhid.enable` is set to 0. `hidraw` can only be used if `hw.usb.usbhid.enable` is set to 1 and the `hidraw` module is loaded explicitly. If `hw.usb.usbhid.enable` is set to 1 but `hidraw` is not available, there won't be anything setup to provide userspace HID access.

Any solution would probably have to be prepared to use whichever API the FreeBSD environment is configured to use as I don't think it would be a good practice to forcefully enable `hidraw` against the wishes of the end user.

mcuee:
> > That would be wonderful. I tend to think the 1hidraw` backend is the way to go for FreeBSD.
> 
> Preliminary investigation suggests this has its own problems due to how new this backend is. It isn't enabled by default on fresh installations. Of the two, `uhid` and `hidraw`, only one can be attached to any particular device.
> 
> `uhid` is the default if `hw.usb.usbhid.enable` is set to 0. `hidraw` can only be used if `hw.usb.usbhid.enable` is set to 1 and the `hidraw` module is loaded explicitly. If `hw.usb.usbhid.enable` is set to 1 but `hidraw` is not available, there won't be anything setup to provide userspace HID access.
> 
> Any solution would probably have to be prepared to use whichever API the FreeBSD environment is configured to use as I don't think it would be a good practice to forcefully enable `hidraw` against the wishes of the end user.

Thanks for the analysis.

However, for the FreeBSD OS, there is already a libusb backend which will be the default as it already works fine (other than common libusb issues: https://github.com/libusb/hidapi/issues?q=is%3Aopen+is%3Aissue+label%3Alibusb). So libusb backend will be the preferred backend in the foreseeable future.

I believe ugen driver is also attached to generic USB HID device along with uhid (or hidraw). That is why it works.

I see this as similar to Linux side last time -- libusb was the preferred and hidraw was secondary as there was too many limitations previously. Now hidraw is the preferred.

But of course you can develop uhid backend first as it is more common for now. However, I tend to believe hidraw may be the future for FreeBSD as well (and maybe NetBSD/OpenBSD).

mcuee:
@Youw 

This PR does not support the existing auto-tools build system. Do we want to keep the auto-tools support?

braewoods:
> This PR does not support the existing auto-tools build system. Do we want to keep the auto-tools support?
I left it out because I found comments saying that it was deprecated so I assumed it was going to be removed in the near future.

Youw:
> Do we want to keep the auto-tools support?

No.

> I left it out because I found comments saying that it was deprecated so I assumed it was going to be removed in the near future.

This is exactly why.

---
BTW: I've reviewed all CMake changes - very consistent with what we already have. Very nice, thanks.

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+ *
+ * This function can be called repeatedly until it returns non-0
+ * Usage is found. pos is the starting point (initially 0) and will be updated
+ * to the next search position.
+ *
+ * The return value is 0 when a pair is found.
+ * 1 when finished processing descriptor.
+ * -1 on a malformed report.
+ */
+static int get_next_hid_usage(const uint8_t *report_descriptor, uint32_t size, uint32_t *pos, uint16_t *usage_page, uint16_t *usage)
```
Youw, line 244: I've just merged additional change to linux/hidraw to master that changes this function
lets update here as well

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+		uint16_t page;
+		uint16_t usage;
+
+		pos = page = usage = 0;
+
+		/*
+		 * Parse the first usage and usage page
+		 * out of the report descriptor.
+		 */
+		if (!get_next_hid_usage(ucrd->ucrd_data, ucrd->ucrd_size, &pos, &page, &usage)) {
```
Youw, line 363: this is probably a copy-paste (or not), but since technically we're writing this from scratch - I'd prefer reading such conditionas as:
```suggestion
		if (get_next_hid_usage(ucrd->ucrd_data, ucrd->ucrd_size, &pos, &page, &usage) == 0) {
```
generally speaking, having a negation (`!`) automatically triggers human brain to look for a failed call (even if often not the case in unix world)

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+{
+	return (!strncmp(s, "uhidev", 6) && isdigit((int) s[6]));
+}
+
+static int is_uhid_device(const char *s)
+{
+	return (!strncmp(s, "uhid", 4) && isdigit((int) s[4]));
+}
+
+static void walk_device_tree(int drvctl, const char *dev, int depth, char arr[static 256][16], size_t *len, int (*cmp) (const char *))
```
Youw, line 418: lets have a (global) named constant/define for 256 (`HIDAPI_MAX_SUPPORTED_DEVICES`)?

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+	strlcpy(dla.l_devname, dev, sizeof(dla.l_devname));
+	dla.l_childname = childname;
+	dla.l_children = 256;
+
+	res = ioctl(drvctl, DRVLISTDEV, &dla);
+	if (res == -1)
+		return;
+
+	if (dla.l_children > 256)
+		return;
```
Youw, line 439: how about:
```suggestion
	l_children_to_enumerate = min(dla.l_children, HIDAPI_MAX_SUPPORTED_DEVICES);
```

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+
+	if (hed->vendor_id != 0 && hed->vendor_id != udi->udi_vendorNo)
+		return;
+
+	if (hed->product_id != 0 && hed->product_id != udi->udi_productNo)
+		return;
+
+	for (size_t i = 0; i < USB_MAX_DEVNAMES; i++) {
+		const char *parent_dev;
+		char arr[256][16];
```
Youw, line 481: is this `256` is same as above or not?
(use same or different named constant accordingly)

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+{
+	const char *locale;
+
+	/* indicate no error */
+	register_global_error(NULL);
+
+	/* Set the locale if it's not set. */
+	locale = setlocale(LC_CTYPE, NULL);
+	if (!locale)
+		setlocale(LC_CTYPE, "");
```
Youw, line 613: can you confirm this is required for this backend and not just a copy-paste?

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+	walk_device_tree(drvctl, "", 0, arr, &len, is_usb_controller);
+
+	hed.root = NULL;
+	hed.end = NULL;
+	hed.drvctl = drvctl;
+	hed.vendor_id = vendor_id;
+	hed.product_id = product_id;
+
+	for (size_t i = 0; i < len; i++) {
+		char devpath[32];
```
Youw, line 654: I'm seeing `16` and `32` magic numbers often - lets have a named constant for that too

Youw: in all places

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+
+		strlcpy(path, dev->path, sizeof(path));
+
+		break;
+	}
+
+	hid_free_enumeration(devs);
+
+	if (*path == '\0') {
+		register_global_error("Device with requested VID/PID/(SerialNumber) not found");
```
Youw, line 717: NOTE: global error never clears
if the previous call has failed, and the next one didn't - the global error still going to point to the previous failure

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+		return NULL;
+	}
+
+	return hid_open_path(path);
+}
+
+HID_API_EXPORT hid_device * HID_API_CALL hid_open_path(const char *path)
```
Youw, line 724: global error not cleared

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+		char devpath[32];
+		int uhid;
+		int rep_id;
+		struct pollfd *ph;
+
+		child_dev = arr[i];
+		strlcpy(devpath, "/dev/", sizeof(devpath));
+		strlcat(devpath, child_dev, sizeof(devpath));
+
+		uhid = open(devpath, O_RDWR | O_CLOEXEC);
```
Youw, line 771: so we always open all report handles
lets havea a `README.md` under `netbsd/` and describe important details like that

Youw (commented):
Overall very consistent implementation.
I didn't check all of the implementation/logic (specially what consernes the `uhid` API/usage) but overall looks very good.
Just a few comments/improvements to consider.

braewoods:
Re-based on latest `master` branch.

netbsd/hid.c
```diff
@@ -0,0 +1,1121 @@
...
+
+#include "hidapi.h"
+
+struct hid_device_ {
+	wchar_t *last_error_str;
+	struct hid_device_info *device_info;
+	size_t poll_handles_length;
+	struct pollfd poll_handles[256];
+	int report_handles[256];
+	int device_handle;
```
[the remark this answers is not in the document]
braewoods: As you wish. This is just the result of me trying to order struct fields in terms of size to reduce the need for compiler added padding for data alignment.

netbsd/hid.c
braewoods: These are tied to an implementation defined size. I'll find a suitable macro they already defined to use.

netbsd/hid.c
braewoods: It is written this way for a reason. If the caller doesn't provide enough space, NetBSD instead says how much it would have used. There's nothing to iterate over if this happens. The array would be full of garbage data.

braewoods: The reason it exists is to be a failsafe if my assumptions somehow end up being wrong on some NetBSD system that has a very large number of child devices on a given parent device.

netbsd/hid.c
braewoods: I can do this. But seeing as this is related to the other comment, what do you prefer? I was using stack-based fixed size allocations to not have to deal with dynamic memory within a recursive function. This is faster overall than the other approach I could use to retrieve the size before doing the actual data request. I figured that it is very unlikely that any given device node will have more than 256 child devices. This really is about how many child nodes you can expect to encounter in any given parent node in the NetBSD device tree.

netbsd/hid.c
braewoods: It is identical to the others. I am reusing the device tree walker function because the system call approach is the same, it just starts with a specific parent device instead of the empty string (which is the root of the device tree here).

netbsd/hid.c
braewoods: Resolved by clearing global error before `hid_hid_open_path` returns success. All other exit points set a new error.

netbsd/hid.c
braewoods: Resolved by clearing global error before `hid_open_path` returns success. All other exit points set a new error.

netbsd/hid.c
braewoods: Apparently it isn't. I kept it originally because I was assuming it was required for the multi-byte functions to work correctly on utf8 input like it is with `glibc`.

braewoods:
Ok. That should handle the requested changes. Anything else?

netbsd/hid.c
Youw: I'm perfectly fine with pre-allocated stack memory
my only point is to have a single name/definition for the magic number

thanks

netbsd/hid.c
Youw: then lets have an explicit comment about that
I might not be the only one who may thing about an "optimisation" like that

Youw:
just one last comment

braewoods:
There. That comment should cover it.

Youw (approved):

PR #625, opened 2023-09-15
Only enumerate each interface once

slouken:
If a device has multiple altsettings, like an Xbox Series X controller, make sure we only enumerate each interface once, not once for each altsetting.


Youw (approved):
LGTM
thanks

PR #627, opened 2023-09-25
windows: change MAX_STRING_WCHARS to 126

vovkos:
Win32 HID API doc says: For USB devices, the maximum string length is 126 wide characters (not including the terminating NULL character).

For certain USB devices, using a buffer larger or equal to 127 wchars results in successful completion of HID API functions, but a broken string is stored in the output buffer. This behaviour persists even if HID API is bypassed and HID IOCTLs are passed to the HID driver directly (IOCTL_HID_GET_MANUFACTURER_STRING, IOCTL_HID_GET_PRODUCT_STRING, etc).

So, the buffer MUST NOT exceed 126 wchars.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -599,18 +599,17 @@ static void hid_internal_get_ble_info(struct hid_device_info* dev, DEVINST dev_n
 	}
 }
 
-/* unfortunately, BLUETOOTH vs BLE can't be distinguished using hid_device_info alone
-   so we'll return it as a flag in the high 32-bit (low 32-bits will carry the DEVINST)
+/* Unfortunately, HID_API_BUS_xxx constants alone aren't enough to distinguish between BLUETOOTH and BLE.
+   As such, we'll introduce an internal flag to help us with that; once we're done, we'll clear it away.
 */
 
-#define HID_INTERNAL_INFO_FLAG_BLE 0x100000000ULL
+#define HID_API_BUS_INTERNAL_BLE 0x10
 
-static ULONGLONG hid_internal_get_info(const wchar_t* interface_path, struct hid_device_info* dev)
+static DEVINST hid_internal_detect_bus_type(const wchar_t* interface_path, struct hid_device_info* dev)
 {
 	wchar_t *device_id = NULL, *compatible_ids = NULL;
 	CONFIGRET cr;
-	DEVINST dev_node;
-	ULONGLONG result = 0;
+	DEVINST dev_node = 0;
 
 	/* Get the device id from interface path */
 	device_id = hid_internal_get_device_interface_property(interface_path, &DEVPKEY_Device_InstanceId, DEVPROP_TYPE_STRING);
@@ -631,8 +630,6 @@ static ULONGLONG hid_internal_get_info(const wchar_t* interface_path, struct hid
 	compatible_ids = hid_internal_get_devnode_property(dev_node, &DEVPKEY_Device_CompatibleIds, DEVPROP_TYPE_STRING_LIST);
 	if (!compatible_ids)
 		goto end;
-
-	result = dev_node;
 
 	/* Now we can parse parent's compatible IDs to find out the device bus type */
 	for (wchar_t* compatible_id = compatible_ids; *compatible_id; compatible_id += wcslen(compatible_id) + 1) {
@@ -656,8 +653,7 @@ static ULONGLONG hid_internal_get_info(const wchar_t* interface_path, struct hid
 
 		/* Bluetooth LE devices */
 		if (wcsstr(compatible_id, L"BTHLEDEVICE") != NULL) {
-			dev->bus_type = HID_API_BUS_BLUETOOTH;
-			result |= HID_INTERNAL_INFO_FLAG_BLE;
+			dev->bus_type = (hid_bus_type)(HID_API_BUS_BLUETOOTH | HID_API_BUS_INTERNAL_BLE);
 			break;
 		}
 
@@ -679,7 +675,7 @@ static ULONGLONG hid_internal_get_info(const wchar_t* interface_path, struct hid
 end:
 	free(device_id);
 	free(compatible_ids);
-	return result;
+	return dev_node;
 }
 
 static char *hid_internal_UTF16toUTF8(const wchar_t *src)
@@ -721,7 +717,7 @@ static struct hid_device_info *hid_internal_get_device_info(const wchar_t *path,
 	wchar_t string[MAX_STRING_WCHARS + 1];
 	ULONG len;
 	ULONG size;
-	ULONGLONG flags;
+	DEVINST dev_node;
 
 	/* Create the record. */
 	dev = (struct hid_device_info*)calloc(1, sizeof(struct hid_device_info));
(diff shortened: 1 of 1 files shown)


Youw:
> a broken string is stored in the output buffer

Is this causing un undefined behavior (like reading uninitialized memory) or just is it more a security consern (getting data from some buffer in the memory which we wheren't supposed to get normally.

> For certain USB devices

Is it specific to driver or specific to device descriptor?

> For certain USB devices
> the buffer MUST NOT exceed 126 wchars

Taking into account size limitation for other buses (e.g. Bluetooth) this parameter probably need to be bus-specific, e.g. have it as is (256) by default, and have it 126 for device bus is known and is USB.

vovkos:
> What kind of USB HID device is showing such problem?

I can't quite put my finger on what exactly triggers this behaviour, but here's an example (Windows10 Pro 21H2, Version 10.0.19044 Build 19044)

Before the patch:

```
---------------------------------------
Enumerating HIDAPI devices
---------------------------------------
Loading HID DB...
Enumerating HID devices...

... (skipped)

\\?\HID#VID_14A5&PID_A011#6&1e2f4adc&0&0000#{4d1e55b2-f16f-11cf-88cb-001111000030}
    VID:          14a5
    PID:          a011
    Manufacturer: U
    Product:      U
    Serial:       
    Release:      74
    Usage page:   Generic Desktop
    Usage:        Mouse
    Bus:          USB
    Interface:    0
    Instance:     HID\VID_14A5&PID_A011\6&1e2f4adc&0&0000
    DEVINST:      0x0004

... (skipped)
```

After the patch:

```
---------------------------------------
Enumerating HIDAPI devices
---------------------------------------
Loading HID DB...
Enumerating HID devices...

... (skipped)

\\?\HID#VID_14A5&PID_A011#6&1e2f4adc&0&0000#{4d1e55b2-f16f-11cf-88cb-001111000030}
    VID:          14a5
    PID:          a011
    Manufacturer: USB Gaming Mouse
    Product:      USB Gaming Mouse
    Serial:       
    Release:      74
    Usage page:   Generic Desktop
    Usage:        Mouse
    Bus:          USB
    Interface:    0
    Instance:    HID\VID_14A5&PID_A011\6&1e2f4adc&0&0000
    DEVINST:     0x0004

... (skipped)    
```

Symptoms are as follows:

On return, the output buffer is zeroed except for the first letter; the first letter is pretty much random on every run (sometimes, it matches the first letter of the actual string descriptor).

Tried calling [`HidD_GetManufacturerString`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/hidsdi/nf-hidsdi-hidd_getmanufacturerstring) or even sending [`IOCTL_HID_GET_MANUFACTURER_STRING`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/hidclass/ni-hidclass-ioctl_hid_get_manufacturer_string) directly with the same result.

I was able to find similar reports from other users, for example: https://stackoverflow.com/questions/2329572/trouble-reading-manufacturer-string-from-hid-device-using-hid-dll-api.

In the case of my device, passing a 127 wchars long buffer still results in a broken output (126 wchars work).

Please note that all of the above only applies to this particular USB mouse (other HID devices work fine with the original 256 wchars long buffer).

> Will this PR cause side effects for other HID device like Bluetooth?

That's a good question. [HidD_GetManufacturerString](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/hidsdi/nf-hidsdi-hidd_getmanufacturerstring) and friends fail if the buffer is big enough, so the patch could affect devices with string descriptor exceeding 126 wchars (are there even devices like that?). 

To minimize possible impact, it probably makes sense to make this patch specific to the USB bus (i.e., only limit the buffer size for `HID_API_BUS_USB`.)

> Reference: it seems to me Windows API allows the buffer size to be up to 4093 Bytes, https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/hidsdi/nf-hidsdi-hidd_getindexedstring

Well, the same page also states that USB string descriptors are limited to 126 wchars:

> For USB devices, the maximum string length is 126 wide characters (not including the terminating NULL character).

The doc is a bit vague on whether it's OK to pass a 127 wchars long buffer to accommodate the NULL-terminator, but my experiments showed that it's not (fails for 127 wchars, works for 126 wchars).

vovkos:
> Is this causing un undefined behavior (like reading uninitialized memory) or just is it more a security consern (getting data from some buffer in the memory which we wheren't supposed to get normally.

It doesn't look like it's actually breaking anything or leading to a crash; just the output buffer is wrong (a random letter, then zeroes). 

> Is it specific to driver or specific to device descriptor?

Can't really tell whether it's caused by a particular device, USB driver, or HID driver (but not the Win32 HID library, because I tried passing HID IOCTLs directly). But again, it's not unique to my mouse -- there are similar reports on the web.

> Taking into account size limitation for other buses (e.g. Bluetooth) this parameter probably need to be bus-specific, e.g. have it as is (256) by default, and have it 126 for device bus is known and is USB.

Good point. I'll make the changes and make this limitation specific to the USB bus.

windows/hid.c
```diff
@@ -590,11 +599,18 @@ static void hid_internal_get_ble_info(struct hid_device_info* dev, DEVINST dev_n
...
 }
 
-static void hid_internal_get_info(const wchar_t* interface_path, struct hid_device_info* dev)
+/* unfortunately, BLUETOOTH vs BLE can't be distinguished using hid_device_info alone
+   so we'll return it as a flag in the high 32-bit (low 32-bits will carry the DEVINST)
+*/
+
+#define HID_INTERNAL_INFO_FLAG_BLE 0x100000000ULL
+
+static ULONGLONG hid_internal_get_info(const wchar_t* interface_path, struct hid_device_info* dev)
```
Youw, line 608: this is no longer `hid_internal_get_info`
lets name it explicitly: `hid_internal_get_bus_type`

Youw (commented):
other then my last commnet - looks good

windows/hid.c
vovkos: Makes sense. Also, I refactored the internal BLE flag processing to avoid OR-ing it with a `DEVINST`. Also, if you don't mind, let's call the function `hid_internal_detect_bus_type`, not `hid_internal_get_bus_type`  (because it returns `DEVINST`, not `hid_bus_type`). 

Take a look.

Youw: if we're to bring this conversation, to my taste - both are hacks
I'd rather make a separate struct with DEVINST+flags as separate members (as an out parameter)

as for the suggestion naming - I'm fine with it

vovkos: Fair enough; let's return flags explicitly. `hid_internal_detect_bus_type` now returns a struct with two fields (`dev_node` and `bus_flags`).

Youw (approved):

vovkos:
The `windows-cmake` check still fails, but it has nothing to do with this PR (compilation errors in `hid_report_reconstructor_test`)

Youw:
No worries. I'll override that check when it is time to merge.
I always give a few days for community to have a chance to review the PR even after it has been approved.

DJm00n:
According to the [USB 3.2 Revision 1.1 Specification](https://www.usb.org/document-library/usb-32-revision-11-june-2022). Section 9.6.9.:

image

>The UNICODE string descriptor (shown in Table 9-31) is not NULL-terminated. The string length is computed by subtracting two from the value of the first byte of the descriptor.

So maximum size in BYTES is 255-2=253. Or 127 WCHARS with NULL-termination.

I like the idea of this workaround for possible Windows bug, but do not like this `hid_internal_detect_bus_type_result` in implementation. I think it will be better to add `DEVINST hid_internal_get_devnode(const wchar_t* interface_path)` instead.

Youw:
> better to add `DEVINST hid_internal_get_devnode(const wchar_t* interface_path)`

I agree this would do good from Single Responsibility principle,
but from implementation POV I believe it will be non-trivial to have a separate `hid_internal_get_devnode` and a function that would detect if the device is a BLE device w/o some code duplication.
I believe current aproach is a good balance in the overall design.

Youw (approved):

Issue #630, opened 2023-10-19
hidapi 0.14.0 for Windows with C++-Builder 11


Stoccarda:
I have been using hid.c in a version from before September 2019 for
32- and 64-bit Windows programs.
Compiler is CLANG in C++ Builder 11.3 (and earlier) IDE.
hid.c was compiled and embedded into the main program, no dll.
The main program had in most cases a GUI.

It worked for me, except for rising an exception when closing the program, and
hence closing the HID device.

When compiling the current 0.14.0 several errors / warning occurred:

---

in hidapi_descriptor_reconstruct.h:

#if defined(__MINGW32__) || defined(__CYGWIN__)
...
#else
	union {
		hid_pp_cap caps[];
		hid_pp_link_collection_node LinkCollectionArray[];
	};
#endif

[bcc32c error] hidapi_descriptor_reconstruct.h(220): type name requires a specifier or qualifier
[bcc32c error] hidapi_descriptor_reconstruct.h(220): expected ';' at end of declaration list

When extending the first line to
#if defined(__MINGW32__) || defined(__CYGWIN__) || defined(__clang__)
it works. It also works with __BORLANDC__ .

---

Compiler warning "[bcc64 Warnung] hidapi_hidsdi.h(43): redefinition of typedef 'PHIDP_PREPARSED_DATA' is a C11 feature"

typedef struct _HIDP_PREPARSED_DATA * PHIDP_PREPARSED_DATA;

was previously defined in hidapi_hidpi.h(39)
Won't be the typedef in hidapi_hidpi.h(39) sufficient?

---

Second warning from hid.c in line 267:
int printf_written = swprintf(msg, msg_len + 1, L"%.*ls: (0x%08X) %.*ls", (int)op_len, op, error_code, (int)system_err_len, system_err_buf);

[bcc64 Warnung] hid.c(267): incompatible integer to pointer conversion passing 'unsigned long' to parameter of type 'const wchar_t *' (aka 'const unsigned short *')

---

There is a linker error with the 32-bit compiler version:
[ilink32 Fehler] Fatal: Invalid VIRDEF Fixup-Index in Modul 'hid.c'

It works when creating a 64-bit version.

---

When running a program 

	res = hid_get_input_report_length(handle);
	ShowMessage((String)"input_report_length: " + res);
shows a length of zero.

	res = hid_get_output_report_length(handle);
	ShowMessage((String)"output_report_length: " + res);
returns the correct value.	

The following code was used for it:

[...]

PR #631, opened 2023-10-22
Added clang-cl build to CI 

JoergAtGithub:
- Using -G"NMake Makefiles", because I couldn't get the other generators to work with clang-cl
- ASAN=off because it's not supported by clang-cl yet

Please not that the MinGW fail already occured with the last commit on the master branch and is unrelated to this PR.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -181,7 +181,7 @@ jobs:
     - name: Build CMake MSVC
       working-directory: build/msvc
       run: cmake --build . --config RelWithDebInfo --target install
-    - name: Check artifacts MSVC
+    - name: Check artifacts CMake MSVC
       uses: andstor/file-existence-action@v2
       with:
         files: "install/msvc/lib/hidapi.lib, \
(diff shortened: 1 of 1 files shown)


Youw (approved):
Interesting this doesn't catch the #630
But it is good to have such variation of the build checked by CI anyway, thanks.

Youw:
BTE: the MinGW failure - there is a valid issue in the code, as per log.
Would be great to handle that too (separately of course).

JoergAtGithub:
> Interesting this doesn't catch the #630

This replaces only the compiler, but not the C runtime. I guess only the clang CRT shows these issues.

.github/workflows/builds.yml
```diff
@@ -237,10 +237,48 @@ jobs:
...
+          -D CMAKE_C_COMPILER=clang-cl ^
+          -Dhidapi_ROOT=install\clang_cl ^
+          -DCMAKE_INSTALL_PREFIX=install\clang_cl_test ^
+          "-DCMAKE_C_FLAGS=%MSVC_COMPILE_FLAGS%"
+        cd build\clang_cl_test
+        cmake --build . --target install
+    - name: Run CTest NMake ClangCL
+      shell: cmd
+      working-directory: build/clang_cl
+      run: ctest -C RelWithDebInfo --rerun-failed --output-on-failure
```
Youw, line 280:
```suggestion
      run: ctest -C RelWithDebInfo --no-compress-output --output-on-failure
```

PR #635, opened 2023-10-26
Disable -Wformat-extra-args form MinGW

JoergAtGithub:
It seems that the warning implementations for the format strings in MinGW compiler and MinGW std-crt are not in sync: https://stackoverflow.com/questions/68900199/how-to-get-mingw-gcc-to-recognize-the-zu-format-specifier-for-size-t/68907731#68907731

I don't know a better solution than disabling this check.

### DIFF
--- a/windows/test/hid_report_reconstructor_test.c
+++ b/windows/test/hid_report_reconstructor_test.c
@@ -6,6 +6,7 @@
 
 #if defined(__MINGW32__)
 #pragma GCC diagnostic ignored "-Wformat"
+#pragma GCC diagnostic ignored "-Wformat-extra-args"
 #endif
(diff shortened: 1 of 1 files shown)


Youw:
What about `__USE_MINGW_ANSI_STDIO`?
Have you tried to define it instead?

cmorty:
@Youw  as `-Wformat` is already disabled: Do you see a real issue with also disabling `-Wformat-extra-args`? Especially `extra-args`  should be - opposed to `-Wformat` - covered by other platforms, right?

cmorty:
...ok. I'm going to really fix this....

cmorty:
@Youw Is #641 more to your liking?

Youw:
It very much is, thanks.
Closing in favor of #641

PR #637, opened 2023-11-07
Don't try to memcpy() to a NULL destination.

icculus:
This function is called with a literal NULL for that pointer later in the file.

We had a user run into this in SDL; checking for NULL before memcpy()'ing resolved it.


Youw (approved):
This is a good one, thanks.
I wounder why Coverity didn't catch this.
It is clear (and I know it only now, when I checked explicitly for that) that it is an undefined behavior if either of `src` or `dest` of `memcpy` is a null pointer _even of the size is `0`_.

Youw:
> We had a user run into this in SDL

I wounder what's your user's environment. Clearly that check wasn't required for years, otherwise someone else would hit it by now.
My guess that the implementation of C-runtime library on macOS wasn't as strict as The Standard states.

PR #639, opened 2023-11-16
CMake: Do not install pkgconfig

cmorty:
If `HIDAPI_INSTALL_TARGETS` is set to off, pkgconfig is still installed. This a problem when statically linking against libhid and then using CMake's install-feature.


Youw (approved):
Good catch, thanks.

PR #640, opened 2023-11-16
Doc: Fix target documentation

cmorty:


BUILD.cmake.md
```diff
@@ -214,14 +214,14 @@ This is done to let the host project's developer decide what is important (what
     In a _subdirectory build_, even if not set, those variables remain unchanged, so a host project's developer has a full control over the HIDAPI build configuration.
 
 Available CMake targets after `add_subdirectory(hidapi)` _are the same as in case of [standalone build](#standalone-package-build)_, and a few additional ones:
-- `hidapi_include` - the interface library; `hidapi::hidapi` is an alias of it;
+- `hidapi_include` - the interface library; `hidapi::include` is an alias of it;
```
Youw, line 217: thanks

BUILD.cmake.md
```diff
@@ -214,14 +214,14 @@ This is done to let the host project's developer decide what is important (what
...
 - `hidapi_winapi` - library target on Windows; `hidapi::winapi` is an alias of it;
 - `hidapi_darwin` - library target on macOS; `hidapi::darwin` is an alias of it;
 - `hidapi_libusb` - library target for libusb backend; `hidapi::libusb` is an alias of it;
 - `hidapi_hidraw` - library target for hidraw backend; `hidapi::hidraw` is an alias of it;
 - `hidapi-libusb` - an alias of `hidapi_libusb` for compatibility with raw library name;
 - `hidapi-hidraw` - an alias of `hidapi_hidraw` for compatibility with raw library name;
-- `hidapi` - an alias of `hidapi_winapi` or `hidapi_darwin` on Windows or macOS respectfully.
+- `hidapi` - not a target, but a project containing the correct version number
```
Youw, line 224: this statement is deceptive
1) this is a list of _targets_ as explicitly stated above
2) The original statement is true: [1](https://github.com/libusb/hidapi/blob/9b507edd8ab66bd8c16cbd9379f6c26ca72d6465/windows/CMakeLists.txt#L44), [2](https://github.com/libusb/hidapi/blob/9b507edd8ab66bd8c16cbd9379f6c26ca72d6465/mac/CMakeLists.txt#L32)

cmorty: You are right, I actually missed that. Wouldn't it make sense to remove those two and declare them at the same place as `hidapi::hidapi` for all targets?

Youw: not nesessary
per-platform files has per-platform definitions
documentation is in one place
we should be good as is

cmorty: Ok. The advantage of cleaning things up would be that the sentence would be:
```
`hidapi` [...] an alias of `hidapi::hidapi` 
```
but probably one shouldn't use those targets anyway. :)

Youw: > `hidapi` [...] an alias of `hidapi::hidapi` 

That would be a strong but technically invalid statement as of right now.

Youw: > probably one shouldn't use those targets anyway. :)

That's exactly why that target is added for compatibility cases only.
I'll consider removing it completely for v1.0

PR #641, opened 2023-11-16
Fix the windows build by fixing printf related issues

### LINKED ISSUE #632: Fix MinGW CI build
Recently MinGW caught an issue (several issues) in the test tool:
```sh
D:\a\hidapi\hidapi\hidapisrc\windows\test\hid_report_reconstructor_test.c:105:34: error: too many arguments for format [-Werror=format-extra-args]
  105 |                 if (sscanf(line, "pp_data->MagicKey                             = 0x%02hhX%02hhX%02hhX%02hhX%02hhX%02hhX%02hhX%02hhX\n", &pp_data->MagicKey[0], &pp_data->MagicKey[1], &pp_data->MagicKey[2], &pp_data->MagicKey[3], &pp_data->MagicKey[4], &pp_data->MagicKey[5], &pp_data->MagicKey[6], &pp_data->MagicKey[7])) continue;
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Need to handle it.

cmorty:
This merge-request removes the need for 
```C
    #pragma GCC diagnostic ignored "-Wformat"
    #pragma GCC diagnostic ignored "-Wformat-extra-args"
```
in mingw. It also turns on `-Wformat-signedness` for GNU-compatible compilers.

This fixes #632

### DIFF
--- a/hidtest/test.c
+++ b/hidtest/test.c
@@ -77,7 +77,7 @@ void print_device(struct hid_device_info *cur_dev) {
 	printf("  Release:      %hx\n", cur_dev->release_number);
 	printf("  Interface:    %d\n",  cur_dev->interface_number);
 	printf("  Usage (page): 0x%hx (0x%hx)\n", cur_dev->usage, cur_dev->usage_page);
-	printf("  Bus type: %d (%s)\n", cur_dev->bus_type, hid_bus_name(cur_dev->bus_type));
+	printf("  Bus type: %u (%s)\n", (unsigned)cur_dev->bus_type, hid_bus_name(cur_dev->bus_type));
 	printf("\n");
 }
 
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1929,7 +1929,7 @@ uint16_t get_usb_code_for_current_locale(void)
 		return 0x0;
 
 	/* Make a copy of the current locale string. */
-	strncpy(search_string, locale, sizeof(search_string));
+	strncpy(search_string, locale, sizeof(search_string)-1);
 	search_string[sizeof(search_string)-1] = '\0';
 
 	/* Chop off the encoding part, and make it lower case. */
--- a/windows/hidapi_descriptor_reconstruct.h
+++ b/windows/hidapi_descriptor_reconstruct.h
@@ -37,6 +37,7 @@
 #include <windows.h>
 
 #include "hidapi_hidsdi.h"
+#include <assert.h>
 
 #define NUM_OF_HIDP_REPORT_TYPES 3
 
@@ -124,6 +125,13 @@ typedef struct hid_pp_link_collection_node_ {
 	ULONG  Reserved : 23;
 	// Same as the public API structure HIDP_LINK_COLLECTION_NODE, but without PVOID UserContext at the end
 } hid_pp_link_collection_node, *phid_pp_link_collection_node;
+
+// Note: This is risk-reduction-measure for this specific struct, as it has ULONG bit-field.
+//       Although very unlikely, it might still be possible that the compiler creates a memory layout that is
+//       not binary compatile.
+//       Other structs are not checked at the time of writing.
+static_assert(sizeof(struct hid_pp_link_collection_node_) == 16,
+    "Size of struct hid_pp_link_collection_node_ not as expected. This might break binary compatibility");
 
 typedef struct hidp_unknown_token_ {
 	UCHAR Token; /* Specifies the one-byte prefix of a global item. */
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -3,7 +3,8 @@ name: GitHub Builds
 on: [push, pull_request]
 
 env:
-  NIX_COMPILE_FLAGS: -Wall -Wextra -pedantic -Werror
+  NIX_COMPILE_FLAGS:   -Wall -Wextra -pedantic -Werror
+  GNU_COMPILE_FLAGS: -Wall -Wextra -pedantic -Werror -Wformat-signedness
   MSVC_COMPILE_FLAGS: /W4 /WX
 
 jobs:
@@ -112,8 +113,8 @@ jobs:
     - name: Configure CMake
       run: |
         rm -rf build install
-        cmake -B build/shared -S hidapisrc -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHIDAPI_ENABLE_ASAN=ON -DCMAKE_INSTALL_PREFIX=install/shared -DHIDAPI_BUILD_HIDTEST=ON "-DCMAKE_C_FLAGS=${NIX_COMPILE_FLAGS}"
-        cmake -B build/static -S hidapisrc -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHIDAPI_ENABLE_ASAN=ON -DCMAKE_INSTALL_PREFIX=install/static -DBUILD_SHARED_LIBS=FALSE -DHIDAPI_BUILD_HIDTEST=ON "-DCMAKE_C_FLAGS=${NIX_COMPILE_FLAGS}"
+        cmake -B build/shared -S hidapisrc -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHIDAPI_ENABLE_ASAN=ON -DCMAKE_INSTALL_PREFIX=install/shared -DHIDAPI_BUILD_HIDTEST=ON "-DCMAKE_C_FLAGS=${GNU_COMPILE_FLAGS}"
+        cmake -B build/static -S hidapisrc -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHIDAPI_ENABLE_ASAN=ON -DCMAKE_INSTALL_PREFIX=install/static -DBUILD_SHARED_LIBS=FALSE -DHIDAPI_BUILD_HIDTEST=ON "-DCMAKE_C_FLAGS=${GNU_COMPILE_FLAGS}"
     - name: Build CMake Shared
       working-directory: build/shared
       run: make install
@@ -139,7 +140,7 @@ jobs:
           -S hidapisrc/hidtest \
           -Dhidapi_ROOT=install/shared \
           -DCMAKE_INSTALL_PREFIX=install/shared_test \
-          "-DCMAKE_C_FLAGS=${NIX_COMPILE_FLAGS}"
+          "-DCMAKE_C_FLAGS=${GNU_COMPILE_FLAGS}"
         cd build/shared_test
         make install
     - name: Check CMake Export Package Static
@@ -149,7 +150,7 @@ jobs:
           -S hidapisrc/hidtest \
           -Dhidapi_ROOT=install/static \
           -DCMAKE_INSTALL_PREFIX=install/static_test \
-          "-DCMAKE_C_FLAGS=${NIX_COMPILE_FLAGS}"
+          "-DCMAKE_C_FLAGS=${GNU_COMPILE_FLAGS}"
         cd build/static_test
         make install
 
@@ -282,7 +283,7 @@ jobs:
     - name: Configure CMake MinGW
       shell: cmd
       run: |
-        cmake -G"MinGW Makefiles" -B build\mingw -S hidapisrc -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHIDAPI_WITH_TESTS=ON -DHIDAPI_BUILD_PP_DATA_DUMP=ON -DCMAKE_INSTALL_PREFIX=install\mingw -DHIDAPI_BUILD_HIDTEST=ON "-DCMAKE_C_FLAGS=%NIX_COMPILE_FLAGS%"
+        cmake -G"MinGW Makefiles" -B build\mingw -S hidapisrc -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHIDAPI_WITH_TESTS=ON -DHIDAPI_BUILD_PP_DATA_DUMP=ON -DCMAKE_INSTALL_PREFIX=install\mingw -DHIDAPI_BUILD_HIDTEST=ON "-DCMAKE_C_FLAGS=%GNU_COMPILE_FLAGS%"
     - name: Build CMake MinGW
       working-directory: build\mingw
       run: cmake --build . --target install
@@ -303,7 +304,7 @@ jobs:
           -S hidapisrc\hidtest ^
           -Dhidapi_ROOT=install\mingw ^
           -DCMAKE_INSTALL_PREFIX=install\mingw_test ^
-          "-DCMAKE_C_FLAGS=%NIX_COMPILE_FLAGS%"
+          "-DCMAKE_C_FLAGS=%GNU_COMPILE_FLAGS%"
         cd build\mingw_test
         cmake --build . --target install
     - name: Run CTest MinGW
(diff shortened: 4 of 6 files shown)


mcuee:
This PR seems to work fine with MinGW build, from github action and from my local build.

My local build.
```
MINGW64 /c/work/hid/hidapi_pr641
$ cmake -B build_mingw -D CMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHIDAPI_WITH_TESTS=ON -DHIDAPI_BUILD_PP_DATA_DUMP=ON
-- Building for: Ninja
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- The C compiler identification is GNU 13.2.0
-- Check for working C compiler: C:/msys64/mingw64/bin/cc.exe - skipped
-- hidapi: v0.14.0
-- Configuring done (0.8s)
-- Generating done (0.0s)
-- Build files have been written to: C:/work/hid/hidapi_pr641/build_mingw

$ cmake --build build_mingw
Change Dir: 'C:/work/hid/hidapi_pr641/build_mingw'

Run Build Command(s): C:/msys64/mingw64/bin/ninja.exe -v
[1/8] C:\msys64\mingw64\bin\windres.exe -O coff -Dhidapi_winapi_EXPORTS -I C:/work/hid/hidapi_pr641/hidapi -I C:/work/hid/hidapi_pr641/windows  C:/work/hid/hidapi_pr641/windows/hidapi.rc src/windows/CMakeFiles/hidapi_winapi.dir/hidapi.rc.obj
[2/8] C:\msys64\mingw64\bin\cc.exe  -IC:/work/hid/hidapi_pr641/hidapi -IC:/work/hid/hidapi_pr641/windows -O2 -g -DNDEBUG -std=gnu11 -MD -MT src/windows/test/CMakeFiles/hid_report_reconstructor_test.dir/hid_report_reconstructor_test.c.obj -MF src\windows\test\CMakeFiles\hid_report_reconstructor_test.dir\hid_report_reconstructor_test.c.obj.d -o src/windows/test/CMakeFiles/hid_report_reconstructor_test.dir/hid_report_reconstructor_test.c.obj -c C:/work/hid/hidapi_pr641/windows/test/hid_report_reconstructor_test.c
[3/8] C:\msys64\mingw64\bin\cc.exe -Dhidapi_winapi_EXPORTS -IC:/work/hid/hidapi_pr641/hidapi -IC:/work/hid/hidapi_pr641/windows -O2 -g -DNDEBUG -MD -MT src/windows/CMakeFiles/hidapi_winapi.dir/hid.c.obj -MF src\windows\CMakeFiles\hidapi_winapi.dir\hid.c.obj.d -o src/windows/CMakeFiles/hidapi_winapi.dir/hid.c.obj -c C:/work/hid/hidapi_pr641/windows/hid.c
[4/8] C:\msys64\mingw64\bin\cc.exe -Dhidapi_winapi_EXPORTS -IC:/work/hid/hidapi_pr641/hidapi -IC:/work/hid/hidapi_pr641/windows -O2 -g -DNDEBUG -MD -MT src/windows/CMakeFiles/hidapi_winapi.dir/hidapi_descriptor_reconstruct.c.obj -MF src\windows\CMakeFiles\hidapi_winapi.dir\hidapi_descriptor_reconstruct.c.obj.d -o src/windows/CMakeFiles/hidapi_winapi.dir/hidapi_descriptor_reconstruct.c.obj -c C:/work/hid/hidapi_pr641/windows/hidapi_descriptor_reconstruct.c
[5/8] cmd.exe /C "cd . && C:\msys64\mingw64\bin\cc.exe -O2 -g -DNDEBUG   -shared -o src\windows\libhidapi.dll -Wl,--out-implib,src\windows\libhidapi.dll.a -Wl,--major-image-version,0,--minor-image-version,14 src/windows/CMakeFiles/hidapi_winapi.dir/hid.c.obj src/windows/CMakeFiles/hidapi_winapi.dir/hidapi_descriptor_reconstruct.c.obj src/windows/CMakeFiles/hidapi_winapi.dir/hidapi.rc.obj  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
[6/8] cmd.exe /C "cd . && C:\msys64\mingw64\bin\cc.exe -O2 -g -DNDEBUG  src/windows/test/CMakeFiles/hid_report_reconstructor_test.dir/hid_report_reconstructor_test.c.obj -o src\windows\test\hid_report_reconstructor_test.exe -Wl,--out-implib,src\windows\test\libhid_report_reconstructor_test.dll.a -Wl,--major-image-version,0,--minor-image-version,0  src/windows/libhidapi.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
[7/8] C:\msys64\mingw64\bin\cc.exe  -IC:/work/hid/hidapi_pr641/hidapi -IC:/work/hid/hidapi_pr641/windows -O2 -g -DNDEBUG -std=gnu11 -MD -MT src/windows/pp_data_dump/CMakeFiles/pp_data_dump.dir/pp_data_dump.c.obj -MF src\windows\pp_data_dump\CMakeFiles\pp_data_dump.dir\pp_data_dump.c.obj.d -o src/windows/pp_data_dump/CMakeFiles/pp_data_dump.dir/pp_data_dump.c.obj -c C:/work/hid/hidapi_pr641/windows/pp_data_dump/pp_data_dump.c
[8/8] cmd.exe /C "cd . && C:\msys64\mingw64\bin\cc.exe -O2 -g -DNDEBUG  src/windows/pp_data_dump/CMakeFiles/pp_data_dump.dir/pp_data_dump.c.obj -o src\windows\pp_data_dump\pp_data_dump.exe -Wl,--out-implib,src\windows\pp_data_dump\libpp_data_dump.dll.a -Wl,--major-image-version,0,--minor-image-version,0  src/windows/libhidapi.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
```

mcuee:
On the other hand, I can not reproduce the issue with git main either.
* https://github.com/libusb/hidapi/issues/632#issuecomment-1814565300

cmorty:
It's a compiler issue: 
* `The C compiler identification is GNU 13.2.0`  <- Your system
* `The C compiler identification is GNU 12.2.0` <- Gitlab-Runner

It seems like they turned on `-Wformat-signedness` by default, which it normally isn't. This MR also turns it on, which shouldn't hurt.

hidtest/test.c
```diff
@@ -77,7 +77,7 @@ void print_device(struct hid_device_info *cur_dev) {
 	printf("  Release:      %hx\n", cur_dev->release_number);
 	printf("  Interface:    %d\n",  cur_dev->interface_number);
 	printf("  Usage (page): 0x%hx (0x%hx)\n", cur_dev->usage, cur_dev->usage_page);
-	printf("  Bus type: %d (%s)\n", cur_dev->bus_type, hid_bus_name(cur_dev->bus_type));
+	printf("  Bus type: %u (%s)\n", cur_dev->bus_type, hid_bus_name(cur_dev->bus_type));
```
Youw, line 80:
```suggestion
	printf("  Bus type: %u (%s)\n", (unsigned)cur_dev->bus_type, hid_bus_name(cur_dev->bus_type));
```

windows/hidapi_descriptor_reconstruct.h
```diff
@@ -119,9 +119,9 @@ typedef struct hid_pp_link_collection_node_ {
 	USHORT NumberOfChildren;
 	USHORT NextSibling;
 	USHORT FirstChild;
-	ULONG  CollectionType : 8;
-	ULONG  IsAlias : 1;
-	ULONG  Reserved : 23;
+	UINT  CollectionType : 8;
+	UINT  IsAlias : 1;
+	UINT  Reserved : 23;
```
Youw, line 124: revert - this may change binary compatibility
any particular reason for this change?

libusb/hid.c
```diff
@@ -1929,7 +1929,7 @@ uint16_t get_usb_code_for_current_locale(void)
 		return 0x0;
 
 	/* Make a copy of the current locale string. */
-	strncpy(search_string, locale, sizeof(search_string));
+	strncpy(search_string, locale, sizeof(search_string)-1);
```
Youw, line 1932: I don't believe this is required, is it?

cmorty: My local build compained. ;) Otherwise I woulnd't have change it.
```
/home/morty/data/cherry/hidapi/libusb/hid.c:1932:9: error: ‘__builtin_strncpy’ specified bound 64 equals destination size [-Werror=stringop-truncation]
 1932 |         strncpy(search_string, locale, sizeof(search_string));
      |         ^
cc1: all warnings being treated as errors
```

```
/usr/bin/cc --version
cc (Ubuntu 12.3.0-1ubuntu1~23.04) 12.3.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

hidtest/test.c
cmorty: Wow yea. I hate implementation defined.

windows/hidapi_descriptor_reconstruct.h
cmorty: I was very unsure about this one. I'll just drop it. [UINT](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/52ddd4c3-55b9-4e03-8287-5392aac0627f) and [ULONG](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/32862b84-f6e6-40f9-85ca-c4faf985b822)  should be the same, as `int` and `long` are the same. But the only `int` is covered by the standard (+ implementation defined, of course). When using `ULONG` I should probably also use `%lu` instead of `%u`.

Youw: this is a part of reverce-engineered implementation of internal HID structures
unless there is an explicit statement in MS documentation that ULONG is an alias for UINT in all cases - we can't change it

cmorty: IT also allows to revert https://github.com/libusb/hidapi/commit/98ab22a33d326e5f3b139d16fd0859fb811c548a ( Suppress Visual C++ warning C4214: nonstandard extension used: bit field types other than int) ->  https://github.com/cmorty/hidapi/actions/runs/6891832863

Youw: my last statement stands

cmorty: Yea, but it's a bit field of under 32bit. Thus the binary representation is covered by the number of bits of the field, not the type. 

As all bit fields are all smaller than 32, and I think that we can agree, that there is no 16-bit Windows around, `int` and `long` should always be at least 32 bit. `long` might actually get larger, and that is likely to be the bigger risk of braking binary compatibility.

cmorty: For what it's worth https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types?redirectedfrom=MSDN stats that they are both unsigned 32bit. But, I you insist I'll drop the commit.

Youw: > Yea, but it's a bit field of under 32bit. Thus the binary representation is covered by the number of bits of the field, not the type.

But it changes the _size of the struct_ it is contained in thus it breaks binary compatibility of the array of such structs (if any).

Youw: @JoergAtGithub do you have an oppinion on this change? You probably knows this better.

cmorty: ...no it doesn't.  The type of a bit-field does not change it's size. Only the number behind the colon is relevant. And that the type is large enough to hold the bit-field.

> Bit-fields
>
> Declares a member with explicit width, in bits. Adjacent bit-field members may be packed to share and straddle the individual bytes. 

https://en.cppreference.com/w/c/language/bit_field

And the relevant sentences from the standard:

> 4)
> The expression that specifies the width of a bit-field shall be an integer constant
> expression with a nonnegative value that does not exceed the width of an object of the
> type that would be specified were the colon and expression omitted.  If the value is
> zero, the declaration shall have no declarator.
> 
> 11)
> An implementation may allocate any addressable storage unit large enough to hold a bit-
> field. If enough space remains, a bit-field that immediately follows another bit-field in a
> structure shall be packed into adjacent bits of the same unit. If insufficient space remains,
> whether a bit-field that does not fit is put into the next unit or overlaps adjacent units is
> implementation-defined. The order of allocation of bit-fields within a unit (high-order to
> low-order or low-order to high-order) is implementation-defined. The alignment of the
> addressable storage unit is unspecified.

https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf - §6.7.2.1


According to the standard compilers may actually pad bit-fields. I can't imagine though, that some compiler will ever create padded bit fields, because it will break so much stuff.

Youw: > The type of a bit-field does not change it's size.

I'm refering to the cases like this: 
https://godbolt.org/z/czEzjzKrG

JoergAtGithub: We shouldn't change these, as these are the same types as in the public headers from Microsoft. Using exactly the same is the safest way to keep the binary compatibility.
And binary compatibility of these structs is not covered by our unit tests.

JoergAtGithub: https://godbolt.org/z/5WhYsMW14

Youw: lets revert then

cmorty: Ok, I learned something and hate C/C++ even more. :-( https://godbolt.org/z/9on94frPz
Based on that using UINT seems even more reasonable, but MS's public headers are the killer argument here. I'll drop the commit.

cmorty: OK. I reassured my pedantic me by adding a `static_assert`. I hope that solution is fine with you.

libusb/hid.c
cmorty: I'd leave this, because this will probably pop up eventually when the tool chain is updated. Is that ok with you?

Youw: ok

windows/hidapi_descriptor_reconstruct.h
```diff
@@ -125,6 +126,9 @@ typedef struct hid_pp_link_collection_node_ {
 	// Same as the public API structure HIDP_LINK_COLLECTION_NODE, but without PVOID UserContext at the end
 } hid_pp_link_collection_node, *phid_pp_link_collection_node;
 
+static_assert(sizeof(struct hid_pp_link_collection_node_) == 16,
+    "Size of struct hid_pp_link_collection_node_ not as expected. This might break binary compatibility");
```
Youw, line 130: as much as I like this - there is two problems with this part:
1) this is a half-measure - if we're to static_asset it - needs to be done for all similar structs in this file
2) `static_asset` is a C++ feature, in C - it is a [bit different](https://en.cppreference.com/w/c/language/_Static_assert), as always; need to confirm if all compilers support it equally

cmorty: 1) Yes it's not perfect, but it is more than nothing. And this covers `ULONG`, `USHORT`, `USAGE` and `ULONG`-bit-fields.
2) Actually it is part of C++11 [cppref](https://en.cppreference.com/w/c/language/_Static_assert) / [MS](https://learn.microsoft.com/en-us/cpp/c-language/static-assert-c?view=msvc-170)

Youw: Hm, you got me there, I should be pay more attention to all of the wording.
I'm generally skeptic about C11 support by the compilers (in opposed to C++11), i.e. threading as introduced by C11 is not yet supported by all of the compilers...

> more than nothing

OK, lets have it.

But I'm pretty sure you're not considering that in addition to do  "more than nothing (good)" it also brings some negative side-effect, e.g. when a person sees there is a single `static_assert` it make assume one of:
- there is a similar `static_assert` for all structs for which the size has to be enforced (that is not true);
- structs that don't have it - do not have size restrictions and may be changed for convenience (that is not true either);
I've seen similar cases in my practice.

cmorty: Nothing that cannot be fixed with a comment. ;)

Youw: Right, becase:
- all developers read comments;
- compilers read comments;

(:

Youw (approved):

Issue #650, opened 2023-11-28
win32: hang forever in hid_read_timeout() when calling hid_close().


dontech:
Hello,

There have been previous reports on this:

https://github.com/signal11/hidapi/issues/88
https://github.com/libusb/hidapi/issues/128

These are assumed fix, but they are not.

I have simple cases that hang 100% of the time:

1) Create reader in thread using hid_read_timeout()
2) Call hid_close() in main thread
3) GetOverlappedResult() blocks forever.

dontech:
I discovered that CancelIoEx() has secret properties that can shutdown overlapped operations that CancelIo() cannot for some odd reason.

According to docs, CancelIo() should do the same, but I fear that for reason the system cannot find the overlapped operation from the handle when called. By calling CancelIoEx() directly with the overlapped reference, the problem is resolved.

Pull request sent.

With this fix the problem becomes un-reproducible in my case. 

Thanks,

/pedro

mcuee:
@dontech 

Thanks. Is your issue the same as #133? Your fix seems to be the same as mentioned in #133 as well -- to use `CancelIoEx`.

BTW, I do not see any issues to drop Windows XP support. But @Youw will need to agree.

* #133 
* https://github.com/libusb/hidapi/issues/237

mcuee:
Related discussion:
* #649

Youw:
Closing as duplicate of #133

mcuee:
> > BTW, I do not see any issues to drop Windows XP support. But @Youw will need to agree.
> 
> Again, it is not about WindowsXP support, but rather to catch early impropper use of the API.

What I mean is that if the fix to use CancelEx is good then there is no problem is drop XP support. But since that is not a vliad fix, that is another story.

Issue #649, opened 2023-11-28
win32: Hangs forever in hid_read_timeout() when calling hid_close()


dontech:
As the title states, the problems reported earlier around GetOverlappedResult() is still present.

There where attempted fixed in:

https://github.com/libusb/hidapi/issues/128
https://github.com/signal11/hidapi/issues/88

But for some simple scenarios this is still the case:

1) Read using hid_read_timeout() in a background thread.
2) Attempt to close using the main thread using hid_close()
3) Hang forever in GetOverlappedResult().

Youw:
It has nothing to do with #128 specifically.
HIDAPI is not thread safe. You should not attempt to `hid_close` while perforing other `hid_*` funcitons on the same device. That is undefined behavior.
_Some_ info about it: https://github.com/libusb/hidapi/issues/45

Youw:
Oh, and BTW: https://github.com/libusb/hidapi/commit/4d02971b8391ec69e204a561b64b548433a08559 does not really a fix. There is still an undefined behavior and you'll be getting crashes e.g. [here](https://github.com/libusb/hidapi/blob/master/windows/hid.c#L1154) because `hid_read*` would attempt to access the device that is already closed/freed.

dontech:
> Oh, and BTW: [4d02971](https://github.com/libusb/hidapi/commit/4d02971b8391ec69e204a561b64b548433a08559) is not really a fix. There is still an undefined behavior and you'll be getting crashes e.g. [here](https://github.com/libusb/hidapi/blob/master/windows/hid.c#L1154) because `hid_read*` would attempt to access the device that is already closed/freed.

That is only true, due to the library missing synchronization function calls during read()/write() operations and close().
I do not see any obvious reason why hidapi, in difference to other similar I/O libraries should not be thread-safe using simple O/S provided mechanisms. But i see that you are right that this is not currently the case.

I also mentioned this in: https://github.com/libusb/hidapi/issues/133

e.i, by implementing a simple EnterCriticalSection()/ExitCriticalSection() construct in the win32 port, the problem you are describing would not be there. If I find the time, I will propose a patch to show how this would look.

Youw:
You just faild to consider all multi-threading scenarios.
The failure case: https://github.com/libusb/hidapi/pull/651#discussion_r1409218821

dontech:
>> You just faild to consider all multi-threading scenarios.

What you are stating is true and is basic teachings in operating system problematics.
[...]

PR #651, opened 2023-11-28
Fix hanging GetOverlappedResult() when closing device

dontech:
This fixes:

https://github.com/libusb/hidapi/issues/650

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -1278,6 +1278,20 @@ void HID_API_EXPORT HID_API_CALL hid_close(hid_device *dev)
 	if (!dev)
 		return;
 
+	/* Cancel I/O specifically for overlapped operations. There are cases
+	 * where CancelIo() does _not_ cancel the associated overlapped
+	 * operations, causing GetOverlappedResult() to hang forever for
+	 * read/write operations.
+	 *
+	 * This is also the recommended way of cancelling pending overlapped
+	 * operations according to:
+	 *
+	 * https://learn.microsoft.com/en-us/windows/win32/fileio/canceling-pending-i-o-operations
+     */
+	CancelIoEx(dev->device_handle, &dev->ol);
+	CancelIoEx(dev->device_handle, &dev->write_ol);
+
+	/* Cancel all remaining I/O */
 	CancelIo(dev->device_handle);
 	free_hid_device(dev);
 }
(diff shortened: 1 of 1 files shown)


mcuee:
Related discussion:
* #649 
* #650

windows/hid.c
```diff
@@ -223,6 +229,9 @@ static void free_hid_device(hid_device *dev)
 	free(dev->feature_buf);
 	free(dev->read_buf);
 	hid_free_enumeration(dev->device_info);
+	DeleteCriticalSection(&dev->cs);
+	DeleteCriticalSection(&dev->cs_read);
+	DeleteCriticalSection(&dev->cs_write);
 	free(dev);
```
Youw, line 235: this is what I mentioned by "not possible to implement a thread-safe `hid_close`. Imagine a scenario:
1) One thread enters a `hid_close` and enter the critical section;
2) Another thread enters `hid_read` and waits on Critical section;
3) First thread destroys a critical section here (is it even allowed w/o first "leaving" it - shouldn't be allowed, but lets assume it is)
4) First thread `free`s a `dev`
5) Seconds threads is no longer waits on a critical section - it is either successfully entered or got an error the the critical section has been destroyed
5) Second thread tries to access `dev`/read/etc. _after_ it was `free`d by `hid_close`

dontech: This argument is invalid. It is normal for any kind of file operation on any kind of operating system that you cannot call any new file operations after calling close(). This is true for (almost) every operating system on the planet. This is the applications responsibility, just like on any other file system, as handles/pointers/tokens could be reused.

Youw: then why is this (syncronisation) even needed?
so you can safely interrupt the read by hid_close? - That has to be a responsibility of a separate function

Youw: > you cannot call any new file operations after calling close(). This is true for (almost) every operating system on the planet.

You're absolutely right, and let me extend this statement:
- you cannot call any file operation after entering the `close` function (even before it leaves);
- you cannot call `close` until you finish all file operations (leave each corresponding function);

Youw: > This argument is invalid.

The scenario shouldn't be allowed. But if it happens - the argument is valid. We should distinct the two.

dontech: Again, theoretically I agree with everything you state.

However, using this limitation which exists in basically _all_ file systems I have encountered as a reason to never have sane thread safety in a simple library is a stretch....

Youw:
Lets continue this discussin in one place: https://github.com/libusb/hidapi/issues/133

PR #677, opened 2024-04-15
Added hid_send_output_report() function

mbcinergy:
Where hid_get_input_report() was already available for some time, hid_send_output_report() was still missing. 

We have used our own fork with this function added for a couple of years now, but believe it's better to add this to the main repository as well. 

The changes are tested for Windows and on Linux (hidraw and libusb). 

The functions for mac and netbsd are also added but not tested, since we currently have no hardware available for that. It would be great if somebody can test this function for those platforms as well, but the functions are very straightforward so I don't expect any issues with them.

### DIFF
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -307,9 +307,9 @@ extern "C" {
 			single report), followed by the report data (16 bytes). In
 			this example, the length passed in would be 17.
 
-			hid_write() will send the data on the first OUT endpoint, if
-			one exists. If it does not, it will send the data through
-			the Control Endpoint (Endpoint 0).
+			hid_write() will send the data on the first interrupt OUT 
+			endpoint, if one exists. If it does not the behaviour is as 
+			@ref hid_send_output_report
 
 			@ingroup API
 			@param dev A device handle returned from hid_open().
@@ -474,6 +474,8 @@ extern "C" {
 			@returns
 				This function returns the actual number of bytes written and
 				-1 on error.
+
+			@see @ref hid_write
 		*/
 		int HID_API_EXPORT HID_API_CALL hid_send_output_report(hid_device* dev, const unsigned char* data, size_t length);
 
--- a/hidapi/hidapi.h
+++ b/hidapi/hidapi.h
@@ -445,6 +445,38 @@ extern "C" {
 		*/
 		int HID_API_EXPORT HID_API_CALL hid_get_feature_report(hid_device *dev, unsigned char *data, size_t length);
 
+		/** @brief Send a Output report to the device.
+
+			Since version 0.15.0, @ref HID_API_VERSION >= HID_API_MAKE_VERSION(0, 15, 0)
+
+			Output reports are sent over the Control endpoint as a
+			Set_Report transfer.  The first byte of @p data[] must
+			contain the Report ID. For devices which only support a
+			single report, this must be set to 0x0. The remaining bytes
+			contain the report data. Since the Report ID is mandatory,
+			calls to hid_send_output_report() will always contain one
+			more byte than the report contains. For example, if a hid
+			report is 16 bytes long, 17 bytes must be passed to
+			hid_send_output_report(): the Report ID (or 0x0, for
+			devices which do not use numbered reports), followed by the
+			report data (16 bytes). In this example, the length passed
+			in would be 17.
+
+			This function sets the return value of hid_error().
+
+			@ingroup API
+			@param dev A device handle returned from hid_open().
+			@param data The data to send, including the report number as
+				the first byte.
+			@param length The length in bytes of the data to send, including
+				the report number.
+
+			@returns
+				This function returns the actual number of bytes written and
+				-1 on error.
+		*/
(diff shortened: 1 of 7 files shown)


Youw:
So far I haven't encountered cases when you have to use `hid_send_output_report` specifically, instead of using `hid_write`.

Most implementations of `hid_write` would try to use interrupt pipe to send the data, at first, and would fallback to `"hid_send_output_report"` if there is no interrupt pipe for a specific device.

On some platforms, the implementation of `hid_write` is identical to suggested `hid_send_output_report`, and on others - should work identically.
Have you tried it?

mbcinergy:
> So far I haven't encountered cases when you have to use `hid_send_output_report` specifically, instead of using `hid_write`.
> 
> Most implementations of `hid_write` would try to use interrupt pipe to send the data if possible first, and would fallback to `"hid_send_output_report"` if there is no interrupt (output) pipe for a specific device.
> 
> On some platforms, the implementation of `hid_write` is identical to suggested `hid_send_output_report`, and on others - should work identically (but "better"). Have you tried it?

I've tried `hid_write`, but it does not work in this case. The issue is that we have a couple of devices which do have an interrupt output pipe but use the control pipe for some vendor specific reports. That's why the `hid_write` does not work out of the box for those devices. So we need a function which is able to specify the pipe we need to write to.

hidapi/hidapi.h
```diff
@@ -445,6 +445,38 @@ extern "C" {
...
+			@param data The data to send, including the report number as
+				the first byte.
+			@param length The length in bytes of the data to send, including
+				the report number.
+
+			@returns
+				This function returns the actual number of bytes written and
+				-1 on error.
+		*/
+		int HID_API_EXPORT HID_API_CALL hid_send_output_report(hid_device* dev, const unsigned char* data, size_t length);
```
Youw, line 478: Lets add to the documentation something like "see also @ref hid_write".
And lets update the documentation for `hid_write`, specifically:
> hid_write() will send the data on the first OUT endpoint, if one exists. If it does not, it will send the data through the Control Endpoint (Endpoint 0).

Would be great to update and specify that the behavior is as (newly introduced) `hid_send_output_report` in case if interrupt pipe does not exist.

libusb/hid.c
```diff
@@ -1638,6 +1638,36 @@ int HID_API_EXPORT hid_get_feature_report(hid_device *dev, unsigned char *data,
 	return res;
 }
 
+int HID_API_EXPORT hid_send_output_report(hid_device *dev, const unsigned char *data, size_t length)
```
Youw, line 1641: would be great to avoid code duplication among this function and `hid_write`

windows/hid.c
```diff
@@ -1313,6 +1315,17 @@ int HID_API_EXPORT HID_API_CALL hid_get_feature_report(hid_device *dev, unsigned
 	return hid_get_report(dev, IOCTL_HID_GET_FEATURE, data, length);
 }
 
+int HID_API_EXPORT HID_API_CALL hid_send_output_report(hid_device* dev, const unsigned char* data, size_t length)
+{
+	BOOL res = HidD_SetOutputReport(dev->device_handle, (PVOID)data, length);
```
Youw, line 1320: I'm afraid this is not enought

take a look at `hid_send_feature_report` and try to extract common aproach/implementation, so that the two function would behavio alike

Youw (approved):

JoergAtGithub:
Wouldn't it be better to read the `Endpoint Descriptor` to determine if the particular interface of the USB device has a `Interupt Out`?

Youw:
Not sure what do you mean.
Right now libusb backend looks for a first interrupt endpoint of a _particular interface_. But it reads from a USB config descriptor.

I didn't check the linux/hidraw implementation, but Windows and macOS implementation is not exactly known.

JoergAtGithub:
I meant this: https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/standard-usb-descriptors#usb-endpoint-descriptor

windows/hid.c
Youw: I'm not sure why I marked this as resolved - it is clearly not.
See here:
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/hidsdi/nf-hidsdi-hidd_setoutputreport#remarks
and see `hid_send_feature_report` implementation.

mbcinergy: This is fixed in our branch. I think something happened with the references when I rebased on the latest master branch using the update button from this pull request (probably should have done a merge instead of rebase?).

Youw: There are inifinite ways one can mess up git history. And there is (almost) always ways to fix it. Give it a try.

Youw (approved):

Youw:
> I meant this: https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/standard-usb-descriptors#usb-endpoint-descriptor

I'm pretty sure that's what we do on libusb backend.
On other backends we don't have any control, as functions like `write` are the OS/Platform implementation, which appear to have in a similar way.

PR #681, opened 2024-07-10
hidapi: fix "conversion from 'size_t' to 'int', possible loss of data"

madebr:
This warning shows up when building with libusb support using MSVC.
Upstreaming of https://github.com/libsdl-org/SDL/commit/1664ac4fcb395f14c16b738e071ec3a1b058cd7f

https://github.com/libsdl-org/SDL/pull/10199

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1030,7 +1030,7 @@ static void *read_thread(void *param)
 		dev->device_handle,
 		dev->input_endpoint,
 		buf,
-		length,
+		(int)length,
 		read_callback,
 		dev,
 		5000/*timeout*/);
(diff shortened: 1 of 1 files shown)


Youw (approved):

madebr:
FYI, it looks like your cmake script does allow for easy building of hidapi with libusb.

For this, SDL uses a [`FindLibUSB.cmake` module](https://github.com/libsdl-org/SDL/blob/main/cmake/FindLibUSB.cmake) and [uses a GitHub prebuilt release](https://github.com/libsdl-org/SDL/commit/628e083e02703c4bbec5de909d17a874fe691880) to provide binaries

Youw:
> FYI, it looks like your cmake script does not allow for easy building of hidapi with libusb.

Define what do you expect to be easy.
I'm using [libusb-cmake](https://github.com/libusb/libusb-cmake) and because of [this](https://github.com/libusb/hidapi/blob/4578ea24613af03e5631e400ed1167462702a112/libusb/CMakeLists.txt#L11-L12) (need `add_subdirectory(libusb)` before `add_subdirectory(hidapi)` ) - it is quite trivial.

Youw:
One more thing - do you thing we could update HIDAPI CI script to catch this/similar issues in the future?

Youw:
Lets continue the CMake conversation at: https://github.com/libusb/hidapi/discussions/682

PR #688, opened 2024-08-10
hidapi/windows: thread local error messages

d3xMachina:
I've implemented thread local error messages on Windows to fix the double free crash when you have a Read and a Write happening in 2 different threads on the same device or when there is a thread for Read/Write for each device.

Need testing with multiple devices per thread since I only have 1 device per thread in my setup.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -192,8 +192,7 @@ err:
 
 typedef void (*tls_destructor)(void *data, hid_device *dev, BOOLEAN all);
 
-struct tls_allocation
-{
+struct tls_allocation {
 	void *data;
 	DWORD thread_id;
 	tls_destructor destructor;
@@ -201,16 +200,14 @@ struct tls_allocation
 	struct tls_allocation *next;
 };
 
-struct device_error
-{
+struct device_error {
 	HANDLE device_handle;
 	wchar_t *last_error_str;
 
 	struct device_error *next;
 };
 
-struct tls_context
-{
+struct tls_context {
 	struct tls_allocation *allocated;
 	CRITICAL_SECTION critical_section;
 	BOOLEAN critical_section_ready;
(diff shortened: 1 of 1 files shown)


Youw:
While this is an interesting idea to solve the current issue, there is unresolved issue with it:
If you're to use the device in different threads, you'll get several thread_local variables, but hid_close will free only one of them. There is a mem-leak by design here.

d3xMachina:
Oh you're right, I'm working on a fix.

d3xMachina:
I've added a global list which contains the thread local allocations. When a thread wants to allocate memory for a thread local variable, it adds it to this list. The list is cleared as necessary. Accesses to the list are protected with a critical section for thread safety.

The only scenarios this global list is accessed are : 
- the first time an error message is created in a thread (to store the head of the list of errors)
- when the thread is terminated (it removes the data corresponding to the thread)
- when a device is closed (it removes the data corresponding to the device for all threads)
- when hid_exit is called (it removes all the data)

One caveat :
- hid_init must be called beforehand, otherwise the CriticalSection won't be initialized. Not sure how to get around that.

d3xMachina:
I've implemented my suggestion for the caveat I mentioned.

Youw:
Thanks for the implementation!

I totaly get your motivation for this change, but I have in mind a few doubts about this:
1) HIDAPI never been advertized as thread-safe;
2) The complexity of implementation is a bit higher than I'd be comfortable to accept and support;
3) Ultimately we have the same problem on other platforms and having it for Windows likely means porting (almost) the same imokementation for other backends.

I was even thinkg about an alternative implementation for this - instead of having a gloal list of thread-local vairables - have a list of per-device error_strings linked to a current thread-id. That should be a bit more simpler.

Another alternative is to introduce a `hid_read_error` API function, which would give only hid_read and hid_read_timeout errors - that implementation should be the most trivial on any platform, but breaks backward compatibility a bit...

PR #689, opened 2024-08-19
Connection callback: fix LeaveCriticalSection not being called after EnterCriticalSection

d3xMachina:


Youw (approved):
LGTM
@DJm00n would you double-check this one too?

DJm00n:
@Youw Yes, looks good. Approving.

Issue #684, opened 2024-08-05
"hid_error is not implemented yet"


fricpa:
how can we address this?

fricpa:
https://github.com/libusb/hidapi/blob/d101e5c7e4646ecde66b650b3b8ce4904acfb13a/libusb/hid.c#L1785

Youw:
> how can we address this?

Contributions are welcome.

Youw:
https://github.com/libusb/hidapi/tree/libusb-error

PR #690, opened 2024-09-01
Implement `hid_error`

matheusmoreira:
The `hid_error` function is not implemented yet, and contributions are welcome. So add error data to the device structure and return it from `hid_error`. Also set error data in the functions I need.

Closes: #684

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -340,6 +340,60 @@ static int is_language_supported(libusb_device_handle *dev, uint16_t lang)
 	return 0;
 }
 
+static wchar_t *utf8_to_wchar(char *s)
+{
+	wchar_t *w = NULL;
+
+/* we don't use iconv on Android, or when it is explicitly disabled */
+#if defined(__ANDROID__) || defined(NO_ICONV)
+
+	w = wcsdup(L"not implemented");
+
+#else
+	size_t slen = strlen(s);
+	wchar_t *wbuf = malloc((slen + 1) * sizeof(wchar_t));
+	if (!wbuf) { goto err; }
+	/* iconv variables */
+	iconv_t ic;
+	size_t inbytes;
+	size_t outbytes;
+	size_t res;
+	char ** restrict inptr;
+	char *outptr;
+	/* buf does not need to be explicitly NULL-terminated because
+	   it is only passed into iconv() which does not need it. */
+
+	/* Initialize iconv. */
+	ic = iconv_open("WCHAR_T", "UTF-8");
+	if (ic == (iconv_t)-1) {
+		LOG("iconv_open() failed\n");
+		return NULL;
+	}
+
+	/* Convert to native wchar_t (UTF-32 on glibc/BSD systems). */
+	inptr = &s;
+	inbytes = slen;
+	outptr = (char*) wbuf;
+	outbytes = slen * sizeof(wchar_t);
+	res = iconv(ic, inptr, &inbytes, &outptr, &outbytes);
+	if (res == (size_t)-1) {
+		LOG("iconv() failed\n");
+		goto err;
+	}
+
+	/* Write the terminating NULL. */
+	wbuf[slen] = 0;
+
+	w = wbuf;
+
+err:
+	iconv_close(ic);
+
+#endif
+
+	return w;
+}
+
 
 /* This function returns a newly allocated wide string containing the USB
    device string numbered by the index. The returned string must be freed
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1669,8 +1669,31 @@ int HID_API_EXPORT hid_send_feature_report(hid_device *dev, const unsigned char 
 		(unsigned char *)data, length,
 		1000/*timeout millis*/);
 
-	if (res < 0)
+	if (res < 0) {
+		const char *context = NULL;
+
+		switch (res) {
+		case LIBUSB_ERROR_TIMEOUT:
+			context = "Transfer timed out";
+			break;
+		case LIBUSB_ERROR_PIPE:
+			context = "Control request not supported by device";
+			break;
+		case LIBUSB_ERROR_NO_DEVICE:
+			context = "Device has disconnected";
+			break;
+		case LIBUSB_ERROR_BUSY:
+			context = "Called from event handling context";
+			break;
+		case LIBUSB_ERROR_INVALID_PARAM:
+			context = "Transfer size larger than supported";
+			break;
+		}
+
+		set_error(dev, res, context);
+
 		return -1;
+	}
 
 	/* Account for the report ID */
 	if (skipped_report_id)
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1888,11 +1888,38 @@ int HID_API_EXPORT_CALL hid_get_report_descriptor(hid_device *dev, unsigned char
 	return hid_get_report_descriptor_libusb(dev->device_handle, dev->interface, dev->report_descriptor_size, buf, buf_size);
 }
 
-
 HID_API_EXPORT const wchar_t * HID_API_CALL  hid_error(hid_device *dev)
 {
-	(void)dev;
-	return L"hid_error is not implemented yet";
+	static const char format_simple[] = "%s: %s";
+	static const char format_context[] = "%s: %s (%s)";
+	const char *name, *message, *context, *format;
+	char *buffer;
+	wchar_t *w;
+	size_t len;
+
+	if (dev->error == LIBUSB_SUCCESS) {
+		return NULL;
+	}
+
+	name = libusb_error_name(dev->error);
+	message = libusb_strerror(dev->error);
+	context = dev->error_context;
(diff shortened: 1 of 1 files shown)


libusb/hid.c
```diff
@@ -123,6 +123,10 @@ struct hid_device_ {
 #ifdef DETACH_KERNEL_DRIVER
 	int is_driver_detached;
 #endif
+
+	int error;
+	wchar_t *error_name;
```
Youw, line 128: the way it is used - `error_name` is useless
it would be much more useful to have a single `error_string` formatted as `<libusb_error_name>: <libusb_strerror>`

libusb/hid.c
```diff
@@ -123,6 +123,10 @@ struct hid_device_ {
 #ifdef DETACH_KERNEL_DRIVER
 	int is_driver_detached;
 #endif
+
+	int error;
```
Youw, line 127: I don't think we need to have it as a member here at all

libusb/hid.c
```diff
@@ -1582,8 +1666,31 @@ int HID_API_EXPORT hid_send_feature_report(hid_device *dev, const unsigned char
 		(unsigned char *)data, length,
 		1000/*timeout millis*/);
 
-	if (res < 0)
+	if (res < 0) {
+		wchar_t *custom_string = NULL;
+
+		switch (res) {
+		case LIBUSB_ERROR_TIMEOUT:
+			custom_string = L"Transfer timed out";
```
Youw, line 1674: why do you need a custom string here?
libusb has its own error string for that

libusb/hid.c
```diff
@@ -1582,8 +1666,31 @@ int HID_API_EXPORT hid_send_feature_report(hid_device *dev, const unsigned char
...
+			break;
+		case LIBUSB_ERROR_PIPE:
+			custom_string = L"Control request not supported by device";
```
Youw, line 1677: that's not nesessary true

libusb/hid.c
```diff
@@ -1582,8 +1666,31 @@ int HID_API_EXPORT hid_send_feature_report(hid_device *dev, const unsigned char
...
+			break;
+		case LIBUSB_ERROR_NO_DEVICE:
+			custom_string = L"Device has disconnected";
+			break;
+		case LIBUSB_ERROR_BUSY:
+			custom_string = L"Called from event handling context";
```
Youw, line 1683: that's not nesessary true

libusb/hid.c
```diff
@@ -1582,8 +1666,31 @@ int HID_API_EXPORT hid_send_feature_report(hid_device *dev, const unsigned char
...
+			break;
+		case LIBUSB_ERROR_INVALID_PARAM:
+			custom_string = L"Transfer size larger than supported";
```
Youw, line 1686: that's not nesessary true

Youw (changes requested):
this is not a full implementation

it should be generic and covered by all functions

libusb/hid.c
matheusmoreira: Not storing the error code would discard the error information returned by libusb.

In my case, value of this `int` is all I wanted. Ideally there would be an `hid_get_libusb_error_code` function.

libusb/hid.c
matheusmoreira: It is simple to compose those two values into a single string later if needed. Parsing them out into two strings is less simple. So I chose to store them separately.

libusb/hid.c
matheusmoreira: Because libusb specifies context-specific error messages in the documentation. "Device disconnected" instead of "broken pipe", for example. This context is only available at the call sites of each function, so only they can pass these contextual messages along.

libusb/hid.c
matheusmoreira: It is what the libusb documentation says though.

```
libusb_control_transfer()

Returns... LIBUSB_ERROR_INVALID_PARAM if the transfer size is larger than the operating system and/or hardware can support (see Transfer length limitations)
```

If it's not true, then the libusb documentation should be updated.

libusb/hid.c
matheusmoreira: It is what the libusb documentation says though.

```libusb_control_transfer()

Returns... LIBUSB_ERROR_PIPE if the control request was not supported by the device
```

If it's not true, then the libusb documentation should be updated.

libusb/hid.c
matheusmoreira: That's a fair point. This variable is redundant due to the stored error code. I think I just forgot to delete it.

matheusmoreira:
It implements a general stored error mechanism for `hid_error`. There's even an internal API that makes it easy to use.

All that remains is setting the error data in the appropriate functions. I added the one I needed. More can be added in future contributions.

matheusmoreira:
Fixed the issues pointed out in code review. Got rid of a lot of wide character stuff, much easier to just do everything needed with the utf8 data and convert just before returning the actual result. Improved the generated error message, now it contains the standard libusb error strings in addition to the context, if any.

libusb/hid.c
Youw: `ERROR_PIPE` might be returned in some other cases, not only when the device gracefully processed the request and sent a response that it is not supported. It is hard to document all corner cases, specially when some of them device-specific, etc.

libusb/hid.c
Youw: My oppinion on this is that best to return `LIBUSB_ERROR_BUSY` as is, and leave the interpretation/digging documentation to the user of HIDAPI, and not try to hard-code things that might be changed in the future (e.g. updated documentation for libusb, etc.).

Youw:
>  Ideally there would be an `hid_libusb_error` function

Why not - we have all means for it.

Youw:
> I added the one I needed. More can be added in future contributions.

In that case I'd accept it into the future branch for now. Having small improvements/changes/fixes for a future branch might speedup the development/readiness for master branch.

libusb/hid.c
```diff
@@ -340,6 +345,88 @@ static int is_language_supported(libusb_device_handle *dev, uint16_t lang)
 	return 0;
 }
 
+static wchar_t *utf8_to_wchar(char *s)
```
Youw, line 348: we already have a similar routine (using iconv) - would be great to have a common implementation if possible

libusb/hid.c
```diff
@@ -340,6 +345,88 @@ static int is_language_supported(libusb_device_handle *dev, uint16_t lang)
...
+{
+	wchar_t *w = NULL;
+
+/* we don't use iconv on Android, or when it is explicitly disabled */
+#if defined(__ANDROID__) || defined(NO_ICONV)
+
+	w = wcsdup(L"not implemented");
```
Youw, line 355: but we need a fallback implementation anyway

libusb/hid.c
```diff
@@ -340,6 +345,88 @@ static int is_language_supported(libusb_device_handle *dev, uint16_t lang)
...
+	return w;
+}
+
+static wchar_t *libusb_error_name_wchar(int error) {
+	return libusb_error_wchar(error, libusb_error_name);
+}
+
+static wchar_t *libusb_strerror_wchar(int error) {
+	return libusb_error_wchar(error, libusb_strerror);
+}
```
Youw, line 423: doesn't look like it is being used

libusb/hid.c
```diff
@@ -1582,8 +1669,31 @@ int HID_API_EXPORT hid_send_feature_report(hid_device *dev, const unsigned char
...
+		case LIBUSB_ERROR_NO_DEVICE:
+			context = "Device has disconnected";
+			break;
+		case LIBUSB_ERROR_BUSY:
+			context = "Called from event handling context";
+			break;
+		case LIBUSB_ERROR_INVALID_PARAM:
+			context = "Transfer size larger than supported";
+			break;
+		}
```
Youw, line 1691: if these are documented by libusb itself - even stronger reason not to have it here at all

libusb/hid.c
```diff
@@ -1778,11 +1888,38 @@ int HID_API_EXPORT_CALL hid_get_report_descriptor(hid_device *dev, unsigned char
...
+		return NULL;
+	}
+
+	snprintf(buffer, len, format, name, message, context);
+
+	w = utf8_to_wchar(buffer);
+
+	free(buffer);
+
+	return w;
```
Youw, line 1922: and this is a memory leak - by the documentation what is returned by hid_error - is owned by HIDAPI

Youw:
I've updated the implementation myself: #698 
I don't have a good environment right now to do extensive testing, @matheusmoreira would be great if you could check it.

matheusmoreira:
> > Ideally there would be an `hid_libusb_error` function
> 
> Why not - we have all means for it.

Huh, I was trying to avoid adding new functions this whole time. If we can just add that function then most of this PR becomes unnecessary since its just juggling error strings.

Looks like it's become a new pull request so I'll close this one.

PR #692, opened 2024-09-15
hidapi/linux: fix uninitialized variable

0xilis:
In `parse_hid_vid_pid_from_uevent_path`, if the file is less than 1024 bytes somehow, the read() call will return the size of the file and pass the `if (res < 0)` check, but the leftover bytes that were not copied to from the file will remain uninitialized. This commit fixed it by initializing the buffer with 0.

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -511,7 +511,7 @@ static int parse_hid_vid_pid_from_uevent_path(const char *uevent_path, unsigned 
 		return 0;
 	}
 
-	char buf[1024];
+	char buf[1024] = {0};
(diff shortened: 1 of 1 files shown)


Youw:
1)  I can't even recall it C supports `= {0}` initialisation by the standard.

2) The function ensures that the string is NULL-terminated and that buffer is always threated as a c-string, so uninitialized part of the memory would never be read. Unless you've found a an excemple where it actually fails. Did you?

Youw:
I don't see reasons having this accepted and no response from author.

PR #696, opened 2024-09-25
Add a bunch of sanity checks

Megamouse:
Some of these changes are meant to fix a couple of warnings while reading the code.
I also noticed that there were almost no sanity checks in the code, which may lead to lots of segfaults in user code.

- Win: Do not call CloseHandle in hid_open_path if device_handle is null
- Win: return errors if malloc calls fail
- Add missing NULL checks to many parameters

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -139,6 +139,9 @@ static int return_data(hid_device *dev, unsigned char *data, size_t length);
 static hid_device *new_hid_device(void)
 {
 	hid_device *dev = (hid_device*) calloc(1, sizeof(hid_device));
+	if (!dev)
+		return NULL;
+
 	dev->blocking = 1;
 
 	hidapi_thread_state_init(&dev->thread_state);
@@ -148,6 +151,9 @@ static hid_device *new_hid_device(void)
 
 static void free_hid_device(hid_device *dev)
 {
+	if (!dev)
+		return;
+
 	/* Clean up the thread objects */
 	hidapi_thread_state_destroy(&dev->thread_state);
 
@@ -169,6 +175,9 @@ static void register_error(hid_device *dev, const char *op)
    Only call with a num_bytes of 0, 1, 2, or 4. */
 static uint32_t get_bytes(uint8_t *rpt, size_t len, size_t num_bytes, size_t cur)
 {
+	if (!rpt)
+		return 0;
+
 	/* Return if there aren't enough bytes. */
 	if (cur + num_bytes >= len)
 		return 0;
@@ -198,6 +207,9 @@ static uint32_t get_bytes(uint8_t *rpt, size_t len, size_t num_bytes, size_t cur
 static int get_usage(uint8_t *report_descriptor, size_t size,
                      unsigned short *usage_page, unsigned short *usage)
 {
+	if (!report_descriptor || !usage_page || !usage)
+		return -1;
+
 	unsigned int i = 0;
 	int size_code;
 	int data_len, key_size;
@@ -536,6 +548,9 @@ static int hid_get_report_descriptor_libusb(libusb_device_handle *handle, int in
  */
 static void fill_device_info_usage(struct hid_device_info *cur_dev, libusb_device_handle *handle, int interface_num, uint16_t expected_report_descriptor_size)
 {
+	if (!cur_dev)
+		return;
+
 	unsigned char hid_report_descriptor[HID_API_MAX_REPORT_DESCRIPTOR_SIZE];
 	unsigned short page = 0, usage = 0;
 
@@ -632,6 +647,9 @@ static struct hid_device_info * create_device_info_for_device(libusb_device *dev
 
 static uint16_t get_report_descriptor_size_from_interface_descriptors(const struct libusb_interface_descriptor *intf_desc)
 {
+	if (!intf_desc)
+		return 0;
+
 	int i = 0;
 	int found_hid_report_descriptor = 0;
 	uint16_t result = HID_API_MAX_REPORT_DESCRIPTOR_SIZE;
@@ -685,6 +703,9 @@ static uint16_t get_report_descriptor_size_from_interface_descriptors(const stru
 
 static int is_xbox360(unsigned short vendor_id, const struct libusb_interface_descriptor *intf_desc)
 {
+	if (!intf_desc)
+		return 0;
+
 	static const int xb360_iface_subclass = 93;
 	static const int xb360_iface_protocol = 1; /* Wired */
 	static const int xb360w_iface_protocol = 129; /* Wireless */
@@ -733,6 +754,9 @@ static int is_xbox360(unsigned short vendor_id, const struct libusb_interface_de
 
 static int is_xboxone(unsigned short vendor_id, const struct libusb_interface_descriptor *intf_desc)
 {
+	if (!intf_desc)
+		return 0;
+
 	static const int xb1_iface_subclass = 71;
 	static const int xb1_iface_protocol = 208;
 	static const int supported_vendors[] = {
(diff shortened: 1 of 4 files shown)


libusb/hid.c
```diff
@@ -245,19 +257,23 @@ static int get_usage(uint8_t *report_descriptor, size_t size,
 		}
 
 		if (key_cmd == 0x4) {
-			*usage_page  = get_bytes(report_descriptor, size, data_len, i);
+			if (usage_page)
```
Youw, line 260: I'd rather have all of these checks at the beginning of the function
`get_usage` is designed for internal usage only, so all these checks are really shouldn't be `required` (if it where a C++ - it would have been references, which do not require checks at all), but I don't mind having those as saniti checks, for as long as it doesn't affect the actual implementation

libusb/hid.c
```diff
@@ -717,7 +738,8 @@ static int is_xbox360(unsigned short vendor_id, const struct libusb_interface_de
 		0x9886, /* ASTRO Gaming */
 	};
 
-	if (intf_desc->bInterfaceClass == LIBUSB_CLASS_VENDOR_SPEC &&
+	if (intf_desc &&
```
Youw, line 741: a separate check at the beginning of the function to separate the actual logic of the implementation and the technicalities

libusb/hid.c
```diff
@@ -750,7 +772,8 @@ static int is_xboxone(unsigned short vendor_id, const struct libusb_interface_de
 		0x3537, /* GameSir */
 	};
 
-	if (intf_desc->bInterfaceNumber == 0 &&
+	if (intf_desc &&
```
Youw, line 775: ditto

libusb/hid.c
```diff
@@ -1447,11 +1487,14 @@ int HID_API_EXPORT hid_write(hid_device *dev, const unsigned char *data, size_t
...
 {
+	if (!dev)
+		return 0;
+
 	/* Copy the data out of the linked list item (rpt) into the
 	   return buffer (data), and delete the liked list item. */
 	struct input_report *rpt = dev->input_reports;
 	size_t len = (length < rpt->len)? length: rpt->len;
-	if (len > 0)
+	if (data && len > 0)
```
Youw, line 1497: ditto

linux/hid.c
```diff
@@ -370,11 +385,14 @@ static int get_next_hid_usage(const __u8 *report_descriptor, __u32 size, struct
 			if (data_len == 4) { /* Usages 5.5 / Usage Page 6.2.2.7 */
 				ctx->usage_page = get_hid_report_bytes(report_descriptor, size, 2, ctx->pos + 2);
 				ctx->usage_page_found = 1;
-				*usage = get_hid_report_bytes(report_descriptor, size, 2, ctx->pos);
+
+				if (usage)
```
Youw, line 389: same here

windows/hid.c
```diff
@@ -318,11 +324,17 @@ static void register_string_error_to_buffer(wchar_t **error_buffer, const WCHAR
 
 static void register_winapi_error(hid_device *dev, const WCHAR *op)
 {
+	if (!dev)
```
Youw, line 327: functions like this are used only internally
is it even possible accidentally pass an invalid `dev` here? I'm pretty sure if the user passes an invalid or NULL `dev` into any of the `hid_*` functions - there woud be an undefined behavior (crash) way before we get here in most places...

Megamouse: I don't know, I didn't discriminate when I added these.
In my opinion there shouldn't be any function without sanity checks.
It's not unthinkable that this was possible before I added the other checks after all.

libusb/hid.c
Megamouse: I wasn't sure if there was a use case where you might not care about these outputs.
I can just early out at the beginning of these are supposed to be valid

linux/hid.c
```diff
@@ -347,6 +359,9 @@ struct hid_usage_iterator {
  */
 static int get_next_hid_usage(const __u8 *report_descriptor, __u32 size, struct hid_usage_iterator *ctx, unsigned short *usage_page, unsigned short *usage)
 {
+	if (!report_descriptor || !pos || !usage)
```
Megamouse, line 362: Note to self: this is nonsense :)

Megamouse:
I think I made all the necessary changes.

libusb/hid.c
```diff
@@ -546,8 +558,10 @@ static void fill_device_info_usage(struct hid_device_info *cur_dev, libusb_devic
 		get_usage(hid_report_descriptor, res,  &page, &usage);
 	}
 
-	cur_dev->usage_page = page;
-	cur_dev->usage = usage;
+	if (cur_dev) {
```
Youw, line 561: early return here as well - it doesn't make sense to call this function with NULL cur_dev at all

windows/hid.c
```diff
@@ -1243,8 +1282,15 @@ int HID_API_EXPORT HID_API_CALL hid_send_feature_report(hid_device *dev, const u
...
 	} else {
-		if (dev->feature_buf == NULL)
+		if (dev->feature_buf == NULL) {
 			dev->feature_buf = (unsigned char *) malloc(dev->feature_report_length);
+
+			if (dev->feature_buf == NULL) {
+				register_winapi_error(dev, L"hid_send_feature_report/malloc");
+				return -1;
+			}
+		}
```
Youw, line 1292: this is a good one, thank you

windows/hid.c
Youw: > In my opinion there shouldn't be any function without sanity checks.

That is actually debatable. Most of the checks in this PR are actually do more harm then help.
See my other comment below.

linux/hid.c
```diff
@@ -1326,6 +1383,9 @@ HID_API_EXPORT struct hid_device_info *HID_API_CALL hid_get_device_info(hid_devi
 
 int HID_API_EXPORT_CALL hid_get_indexed_string(hid_device *dev, int string_index, wchar_t *string, size_t maxlen)
 {
+	if (!dev)
+		return -1;
```
Youw, line 1387: all public API (specifically functions that accept `hid_device *` as first argument):
- do not make sens to call with non-valid `dev`
- should register an error string in case if anything goes wrong, so the user could use `hid_error` to check what went wrong in case if `-1` is returned

I do not support having this checks for all public API functions (marked with `HID_API_CALL`/`HID_API_EXPORT_CALL`).
We always assume that the `dev` should be valid, and that is up to the user to make sure it is. (It is fairly easy to implement in other languages that wrap this library w/o any additional runtime checks by using e.g. class invariant).
The only exception is `hid_close` function - similar to `free` it should gracefully accept NULL-device.

All other functions the accept `hid_device *` as an argument called from public API function should use the same assumption, that the `dev` is valid and doesn't require extra checks.

linux/hid.c
Youw: So don't add it :)

There is no code-path where these arguments will not be valid.
And we have regular static analyzer checks to catch things like that.

windows/hid.c
```diff
@@ -1513,6 +1618,11 @@ int HID_API_EXPORT_CALL hid_winapi_get_container_id(hid_device *dev, GUID *conta
 
 int HID_API_EXPORT_CALL hid_get_report_descriptor(hid_device *dev, unsigned char *buf, size_t buf_size)
 {
+	if (!dev) {
+		register_global_error(L"Device is NULL");
+		return -1;
+	}
```
Youw, line 1624: this is 100% not a supported scenario and should not be handled explicitly in such way
pieces like this create too much "noise/distraction" in the code that actually makes the overall implementation worse than better

please revert

Megamouse: So you'd rather have a hard crash on this than an error that the user can actually catch...

Youw: Let me rephrase that (and that is not only my opinion, but rather shared experience of many developers/projects I've seen):
if I as a user messed up some implementation - I'd rather get a crash immediately (and have a chance to fix it immediatelly), that a silent error (often hidden due to lack of error-handling code) which is missed and then the real problem (the one that affects some business scenarios on your application) hits later when something slightly changed in a different part of my application.

Megamouse: I can remove it then. Although I would probably have to double check if there isn't an unguarded call to this.
I can make a separate PR with the obvious changes that don't need discussion (like the malloc ptr checks).

Youw: > I can make a separate PR with the obvious changes that don't need discussion (like the malloc ptr checks).

Great idea, lets start with that.

Megamouse:
If you tell me what to keep and what to remove I can adjust this soon

Youw:
I guess I had too many comments/questions to this PR.
Instead I prepared #727 based on this, with all my comments applied.

Issue #703, opened 2024-10-18
`hid_get_report_descriptor`: missing unnumbered feature report


SuibianP:
On some devices, unnumbered feature reports seem omitted from the parsed `HIDP_PREPARSED_DATA`. The `hid_pp_caps_info` header indicates the count and size correctly, but marks it empty/unused.

```pike
# Feature caps_info struct:
pp_data->caps_info[2]->FirstCap           = 5
pp_data->caps_info[2]->LastCap            = 5
pp_data->caps_info[2]->NumberOfCaps       = 1
pp_data->caps_info[2]->ReportByteLength   = 91
```

```perl
# Output hid_pp_cap struct:
# Feature hid_pp_cap struct:
# Link Collections:
```

https://github.com/libusb/hidapi/blob/750bf201ae906eec1dc52c97087c8d0104c091ae/windows/hidapi_descriptor_reconstruct.h#L109-L114

If I anyhow print the trailing `phid_pp_cap` out, it consists entirely of zeros.

This is likely an existing bug/undocumented aspect of `HidD_GetPreparsedData` on MS side, and I am unsure if it could be worked around in userspace.

<details>
<summary>HID report descriptor (dumped on Linux with <code>hid-decode</code>)</summary>

```c
# device 0:0
# 0x05, 0x01,                    // Usage Page (Generic Desktop)        0
# 0x09, 0x02,                    // Usage (Mouse)                       2
# 0xa1, 0x01,                    // Collection (Application)            4
# 0x09, 0x01,                    //  Usage (Pointer)                    6
# 0xa1, 0x00,                    //  Collection (Physical)              8
# 0x05, 0x09,                    //   Usage Page (Button)               10
# 0x19, 0x01,                    //   Usage Minimum (1)                 12
# 0x29, 0x05,                    //   Usage Maximum (5)                 14
# 0x15, 0x00,                    //   Logical Minimum (0)               16
# 0x25, 0x01,                    //   Logical Maximum (1)               18
# 0x75, 0x01,                    //   Report Size (1)                   20
# 0x95, 0x05,                    //   Report Count (5)                  22
# 0x81, 0x02,                    //   Input (Data,Var,Abs)              24
# 0x75, 0x01,                    //   Report Size (1)                   26
# 0x95, 0x03,                    //   Report Count (3)                  28
# 0x81, 0x03,                    //   Input (Cnst,Var,Abs)              30
[...]

PR #707, opened 2024-10-27
Reconstruction logic for padding bytes at the end of Reports (only possible for devices without ReportID)

JoergAtGithub:
- Added reconstruction logic of const padding bytes at the end of a report for devices which don't use ReportIDs
- Enhanced padding logic to distinguish between byte and bit padding.
- Added a new test case for Razer Cobra mouse `1532_00A3_0002_0001` which has a completly con feature report of 90 bytes.
- Modified expected report descriptor reference file `045E_02FF_0005_0001_expected.rpt_desc` to include additional padding bytes.

Closes: #703


Youw:
If someone could test/verify this - I'll approve/merge.

mcuee:
Not sure if I have test devices which can test the improvement in this PR. Still I will check with my existing devices to see if there are any differences.

SuibianP (approved):
Tested with 1532:00A3. `hid_report_reconstructor_test` adds the last const report correctly.

Youw (approved):

PR #718, opened 2025-02-25
Make windows path compile for both C and C++ direct file include case

toxieainc:
also make it work for older WINVER cases by copying a few defines

this is how we intend/want to use it within https://github.com/vpinball/vpinball, so would be nice to have that in the official repo

### DIFF
--- a/windows/hidapi_descriptor_reconstruct.c
+++ b/windows/hidapi_descriptor_reconstruct.c
@@ -140,7 +140,7 @@ static struct rd_main_item_node * rd_append_main_item_node(int first_bit, int la
 		list = &(*list)->next;
 	}
 
-	new_list_node = malloc(sizeof(*new_list_node)); // Create new list entry
+	new_list_node = (struct rd_main_item_node *)malloc(sizeof(*new_list_node)); // Create new list entry
 	new_list_node->FirstBit = first_bit;
 	new_list_node->LastBit = last_bit;
 	new_list_node->TypeOfNode = type_of_node;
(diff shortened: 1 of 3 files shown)


Youw (approved):
LGTM, thanks.

Would you be willing to add a simple compilation check for this to our CI script?

toxieainc:
Willing: Yes. Capable: No.   ;)

Youw:
Maybe later then

windows/hidapi_descriptor_reconstruct.c
```diff
@@ -203,21 +203,21 @@ int hid_winapi_descriptor_reconstruct_pp_data(void *preparsed_data, unsigned cha
...
 	// Allocate memory and initialize lookup table
 	rd_bit_range ****coll_bit_range;
-	coll_bit_range = malloc(pp_data->NumberLinkCollectionNodes * sizeof(*coll_bit_range));
+	coll_bit_range = (rd_bit_range ****)malloc(pp_data->NumberLinkCollectionNodes * sizeof(*coll_bit_range));
 	for (USHORT collection_node_idx = 0; collection_node_idx < pp_data->NumberLinkCollectionNodes; collection_node_idx++) {
-		coll_bit_range[collection_node_idx] = malloc(256 * sizeof(*coll_bit_range[0])); // 256 possible report IDs (incl. 0x00)
+		coll_bit_range[collection_node_idx] = (rd_bit_range ***)malloc(256 * sizeof(*coll_bit_range[0])); // 256 possible report IDs (incl. 0x00)
 		for (int reportid_idx = 0; reportid_idx < 256; reportid_idx++) {
-			coll_bit_range[collection_node_idx][reportid_idx] = malloc(NUM_OF_HIDP_REPORT_TYPES * sizeof(*coll_bit_range[0][0]));
-			for (HIDP_REPORT_TYPE rt_idx = 0; rt_idx < NUM_OF_HIDP_REPORT_TYPES; rt_idx++) {
```
JoergAtGithub, line 211: Replacing the explicit enum type HIDP_REPORT_TYPE by int feels not like an improvement to me.

Youw: I believe this PR is to make the compiler happy with some of the strict checks.
And enums has never been intended to be used as iteration indexes.

PR #727, opened 2025-03-11
Add missing sanity checks

Youw:
Based on #696

- Add missing checks for zero/null arguments;
- Better management of error strings in case of failures;

Closes: #696


mac/hid.c
```diff
@@ -1108,7 +1108,7 @@ static int set_report(hid_device *dev, IOHIDReportType type, const unsigned char
 	register_device_error(dev, NULL);
 
 	if (!data || (length == 0)) {
-		register_device_error(dev, strerror(EINVAL));
+		register_device_error(dev, "Zero buffer/length");
```
Megamouse, line 1111: Can't you move register_device_error(dev, NULL) below this ?

mac/hid.c
```diff
@@ -1149,6 +1149,11 @@ static int get_report(hid_device *dev, IOHIDReportType type, unsigned char *data
 
 	register_device_error(dev, NULL);
 
+	if (!data || (length == 0)) {
+		register_device_error(dev, "Zero buffer/length");
```
Megamouse, line 1153: same here

Youw: there is even bigger issue here, that I missed earlier - we access `data` before we check for it/its length

Youw: fixed

PR #741, opened 2025-05-18
[Connection callback] Update to fresh Master branch

k1-801:
Just merging in `master`.
Did check that Windows builds still compile just fine, did not run any tests.
Did not check other systems.


k1-801:
It's probably better to rebase instead.

PR #747, opened 2025-05-28
Draft: linux/hidraw: Parse Report ID

tuxedo-wse:
I didn't see a way to retrive Report IDs using hidapi, but sometimes these are required because they are optional and/or different on a per device basis and must be prefixed to reports sent. So I made this quick and currently untested Patch as a conversation starter if and how this could be added.

Or am I missing something completly?


Youw:
I personally haven't used any as of yet, so me recommending anything on this behalf is as good as googling yourself (or asking an AI/LLM of any kind).

PR #765, opened 2025-12-20
[connection-callback] Introduce a Report Retry timeout for hotplug events in HIDRAW

k1-801:
This fix targets the scenario when `udev` reported a new device, but the device is not yet ready to provide a report. This results in the independent `usage` entries not generating for the device and certain fields in the device not being filled.

UPD: The patch, despite potentially solving timing related issues, proved to not be enough to solve the particular problem at hand, investigation continues. Marking as draft, going to update it in the next week.

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -637,7 +637,7 @@ next_line:
 }
 
 
-static struct hid_device_info * create_device_info_for_device(struct udev_device *raw_dev)
+static struct hid_device_info * create_device_info_for_device(struct udev_device *raw_dev, long report_timeout_msec)
 {
 	struct hid_device_info *root = NULL;
 	struct hid_device_info *cur_dev = NULL;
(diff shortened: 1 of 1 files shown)


mcuee:
Somehow Arch Linux build failed again, but not related to this PR.

error: target not found: libudev0

This needs to be changed to libudev0-shim.

mcuee:
Windows cmake build failed as well, again, nothing to do with this PR.

k1-801:
Testing revealed that the issue was *somehow* caused by the library being loaded through `dlopen` and does not replicate with regular build-time linking. Function pointers retrieved through `dlsym` did not match and the calls failed. None of the resulting pointers were NULL.

Building the aplication with `-rdynamic` linker flag also fixed the issue; without this flag, dynamically loaded through `dlopen` hidapi cannot retrieve usages - **and I don't know why**. But at least I know this PR is not necessary.

PR #767, opened 2025-12-30
Fix compilation with WIN32_LEAN_AND_MEAN and use it for hid.c

### LINKED ISSUE #766: MSVC build with hidapi in-source fails with errors related to WINAPI
Image

I am currently trying to build OpenRGB with HIDAPI built into the source (due to the need to ship the `connection-callback` branch specifically without it accidentally cross-linking to anything else). I added hidapi as a submodle specifically targeting that branch, though the specific error at hand is not related to that, but if I misconfigured the submodule - it could fail the build for you so pls let me know if that's the case.

The full project is located in here: https://gitlab.com/k1-801/OpenRGB/-/tree/hotplug_standalone . The project uses QMake build system, which is not very common, and I am using a MSVC 2019 compiler. I have included the `hid.c` files for each operating system into the project source list, as well as included the header folder in the includepath. I made no compiler flags changes for hidapi. The screenshot shows the errors I'm getting.

What could be causing this? How should I go about fixing this build.

If anyone could point me towards what else to check, that'd be awesome.

Thank you all in advance and happy holidays.

k1-801:
Fixes #766

Defining WIN32_LEAN_AND_MEAN removes some rarely-used includes from windows.h, which MSVC uses to speed up compilation. Unfortunately, hidapi_cfgmgr32.h relies on one of those headers. This PR fixes the missing header, as well as makes use of the speed-up define, simultaneously making the build scenarios more uniform regardless of whether the target project uses it or not.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -30,6 +30,11 @@
 extern "C" {
 #endif
 
+#ifndef WIN32_LEAN_AND_MEAN
+/* Reduces the overhead from windows.h */
+#define WIN32_LEAN_AND_MEAN
+#endif
+
 #include "hidapi_winapi.h"
 
 #include <windows.h>
--- a/windows/hidapi_cfgmgr32.h
+++ b/windows/hidapi_cfgmgr32.h
@@ -32,6 +32,7 @@
 /* This part of the header mimics cfgmgr32.h,
     but only what is used by HIDAPI */
 
+#include <ole2.h>
 #include <initguid.h>
 #include <devpropdef.h>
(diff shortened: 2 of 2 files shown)


k1-801:
The CI build fails in the same place as #765. Local build succeeds with a warning in the SDK.

k1-801:
this change also raises a warning in local copilation which can not be fixed, it is a bug in SDK:
https://developercommunity.visualstudio.com/t/issue-in-windows-sdk-header/383016

PR #768, opened 2025-12-31
Fix compilation with WIN32_LEAN_AND_MEAN (through UNDEF)

### LINKED ISSUE #766: MSVC build with hidapi in-source fails with errors related to WINAPI

k1-801:
Fixes #766 
This version is less complicated and just undefines the thing. The previous PR with a proper solution didn't work well with CI, trying a simpler solution.

### DIFF
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -30,6 +30,11 @@
 extern "C" {
 #endif
 
+#ifdef WIN32_LEAN_AND_MEAN
+/* This define may be set by IDE/project and it breaks compilation */
+#undef WIN32_LEAN_AND_MEAN
+#endif
+
 #include "hidapi_winapi.h"
 
 #include <windows.h>
--- a/windows/hid.c
+++ b/windows/hid.c
@@ -31,7 +31,8 @@ extern "C" {
 #endif
 
 #ifdef WIN32_LEAN_AND_MEAN
-/* This define may be set by IDE/project and it breaks compilation */
+/* It may be set by IDE/project and apparently HIDAPI relies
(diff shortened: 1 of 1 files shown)


windows/hid.c
```diff
@@ -30,6 +30,11 @@
 extern "C" {
 #endif
 
+#ifdef WIN32_LEAN_AND_MEAN
+/* This define may be set by IDE/project and it breaks compilation */
```
Youw: add details in which cases compilatino is broken (apparently not always)

Youw: tried locally, maybe "not always" part in my assumption not true, but I don't really want to investigate
adjusted the comment accordingly

Youw (approved):

PR #777, opened 2026-03-13
linux: Parse HID_FIRMWARE_VERSION from uevent for release_number

JohnAZoidberg:
Parse the HID_FIRMWARE_VERSION uevent property to populate release_number for all HID devices. This provides device version information for I2C, Bluetooth, SPI, and virtual HID devices.

For USB devices, the existing bcdDevice lookup from sysfs still takes precedence, maintaining backward compatibility.

The HID_FIRMWARE_VERSION property is being added to the Linux kernel (patch series submitted to linux-input@vger.kernel.org). On kernels without this property, release_number will remain 0 for non-USB devices, preserving existing behavior.

Link: https://lore.kernel.org/linux-input/20260313175659.268094-1-dhs@frame.work/T/#u

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -571,7 +571,8 @@ static int get_hid_report_descriptor_from_hidraw(hid_device *dev, struct hidraw_
  */
 static int parse_uevent_info(const char *uevent, unsigned *bus_type,
 	unsigned short *vendor_id, unsigned short *product_id,
-	char **serial_number_utf8, char **product_name_utf8)
+	char **serial_number_utf8, char **product_name_utf8,
+	unsigned short *version)
 {
 	char tmp[1024];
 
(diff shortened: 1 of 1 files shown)


JohnAZoidberg:
Also added here, which looks like it's going into 7.1 by @superm1: https://git.kernel.org/pub/scm/linux/kernel/git/hid/hid.git/commit/?h=for-7.1/lenovo-v2&id=6ca9029c823b7853e980585e757343e0e84227cd

superm1 (approved):

Youw (approved):
I don't believe I have a way to properly check/test this, but quick googling shows that this is perfectly fine.
Thanks!

linux/hid.c
```diff
@@ -622,6 +623,9 @@ static int parse_uevent_info(const char *uevent, unsigned *bus_type,
 			/* The caller has to free the serial number */
 			*serial_number_utf8 = strdup(value);
 			found_serial = 1;
+		} else if (strcmp(key, "HID_FIRMWARE_VERSION") == 0) {
+			/* Device version from kernel */
```
Youw, line 627: if you could add a comment what kernel version this become available (with reference) - that would be great

JohnAZoidberg: Yes, we can wait until either Mario or my patch gets merged and I can add a comment

Youw: ok, so we wait

PR #782, opened 2026-03-29
Fix high CPU usage in callback thread due to incorrect timespec

k1-801:
@CalcProgrammer1 noticed high CPU usage (50%) on one of the threads when libusb backend is used for hotplug. It was determined that the load was caused by `hidapi_thread_cond_timedwait()`, which requires an absolute timespec, while the code passes a relative value. This patch copies how the structure is filled from the other place where `hidapi_thread_cond_timedwait()` is uses, which should fix the issue.

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1215,16 +1215,16 @@ static void* callback_thread(void* user_data)
 {
 	(void) user_data;
 
-	/* 5 msec timeout seems reasonable; don't set too low to avoid high CPU usage */
-	/* This timeout only affects how much time it takes to stop the thread */
-	hidapi_timespec ts;
-	ts.tv_sec = 0;
-	ts.tv_nsec = 5000000;
-
 	hidapi_thread_mutex_lock(&hid_hotplug_context.callback_thread);
 	
 	/* We stop the thread if by the moment there are no events left in the queue there are no callbacks left */
 	while (1) {
+		/* 5 msec timeout seems reasonable; don't set too low to avoid high CPU usage */
+		/* This timeout only affects how much time it takes to stop the thread */
+		hidapi_timespec ts;
+		hidapi_thread_gettime(&ts);
+		hidapi_thread_addtime(&ts, 5);
+
 		/* Make the tread fall asleep and wait for a condition to wake it up */
 		hidapi_thread_cond_timedwait(&hid_hotplug_context.callback_thread, &ts);
 
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -1219,11 +1219,11 @@ static void* callback_thread(void* user_data)
 	
 	/* We stop the thread if by the moment there are no events left in the queue there are no callbacks left */
 	while (1) {
-		/* 5 msec timeout seems reasonable; don't set too low to avoid high CPU usage */
-		/* This timeout only affects how much time it takes to stop the thread */
+		/* Pretty much any value between 2 and 100 msec timeout seems reasonable; don't set too low to avoid high CPU usage */
+		/* This timeout only affects how much time it takes to stop the thread if for any reason thread shutdown wasn't triggered by condition varaiable */
 		hidapi_timespec ts;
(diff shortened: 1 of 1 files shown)


libusb/hid.c
```diff
@@ -1215,16 +1215,16 @@ static void* callback_thread(void* user_data)
...
-
 	hidapi_thread_mutex_lock(&hid_hotplug_context.callback_thread);
 	
 	/* We stop the thread if by the moment there are no events left in the queue there are no callbacks left */
 	while (1) {
+		/* 5 msec timeout seems reasonable; don't set too low to avoid high CPU usage */
+		/* This timeout only affects how much time it takes to stop the thread */
+		hidapi_timespec ts;
+		hidapi_thread_gettime(&ts);
+		hidapi_thread_addtime(&ts, 5);
```
Youw: I'd set it to 1 second - I've been using 1 second in multiple projects, and seems reasonable
5 second to exit the thread (effectively - to quit the application in some cases) seem a bit too much

k1-801: According to the call signature, the number is in milliseconds, not whole seconds.
image

When it comes to milliseconds, it doesn't really matter what number is set there, as execution at single milliseconds scale is likely limited by OS scheduler, not the actual number set here, so long as it's long enough to actually go idle.

A whole second here would be too long (at least to my taste), and checking every single millisecond seems a bit excessive. It can be changed later on to 20 or 40 or even 100 milliseconds if needed (i.e. if this thread makes any noticeable CPU load).

k1-801: So far, the patch seems to work:

image

("wrapped HID devices" in this context means HID devices that use libusb backend specifically)

Youw: While modern processors are fast-enough practically not to notice wait/idle/check every 5ms, my experience tells it is a bit too often, I'd make it 100ms then.

k1-801: Increased the timeout to 100

Youw:
Meanwhile I'm looking at a fundamentally different aproach to fix this issue - there should be no timeout at all: #783

k1-801:
The approach suggested there would also work; however, the proposed change simultaneously breaks the cache of connected devices, which would lead to inability to report disconnect events.

I don't remember exactly why I started with a timed `cond_wait`, it was either a precaution against infinite wait if, or the exact reason could have been resolved.

Youw:
@k1-801 
Check the PR again - there are also a few multithreading bugs fixed.

> a timed cond_wait, it was either a precaution against infinite wait if

There is a high change that because of this:
https://github.com/libusb/hidapi/pull/783/changes#diff-516ed9d234e81614e6fa65cfc81a50b8b119e8a0769b2e66a1c89166bac437f7R1286-R1288

There is exacty one reason why a timed wait should be used in cases like that - when it is impossible to single the even in a clean way in any other was. In this case - it is a 100% possible and should be done exactly like that.

mcuee:
#783 has been merged. Do we still need this one?

Youw:
Interesting, it should have been closed on its own

CalcProgrammer1:
Is there a new "finished" hotplug branch available?  If so let me know which branch and I will update my fork and test it in OpenRGB next this weekend across Windows, Linux, and MacOS.

mcuee:
> Is there a new "finished" hotplug branch available? If so let me know which branch and I will update my fork and test it in OpenRGB next this weekend across Windows, Linux, and MacOS.

https://github.com/libusb/hidapi/tree/connection-callback

Not yet finished...

CalcProgrammer1:
That's fine, I will test this latest branch.  I meant finished in terms of this issue mainly.

CalcProgrammer1:
With the updated connection-callback branch I am seeing about 0.7% CPU usage in htop from the openrgb background service on Linux, which is in line with what I would expect.  I don't have the devices on me to test the wrapped/libusb backend right now but the hidraw backend successfully detects hotplug and removal.  Both backends have their hotplug listening threads running as part of that 0.7%.

PR #674, opened 2024-04-06
RFC: Hotplug implementation

Youw:
Resolves: #238

PR #784, opened 2026-03-30
A bit of documentation details on callback API

Youw:


hidapi/hidapi.h
```diff
@@ -303,6 +303,11 @@ extern "C" {
 			This callback may be called by an internal event thread and as such it is
 			recommended the callback do minimal processing before returning.
 
+			It is NOT safe to call hid_open/hid_open_path or any other global/non-device
+			functions from within this callback. If you need to open
+			or close a device in response to a hotplug event, queue the
+			device path and handle it on your own application thread.
```
k1-801, line 309: I personally don't see any reason why opening or closing a device would be unsafe in a callback. There aren't any intersections between the hotplug subsystem and the code responsible for opening the devices, so there should be no race conditions.

Youw: What can I say...
For starters, check: https://github.com/libusb/hidapi/wiki/Multi%E2%80%90threading-Notes

HIDAPI is not, and never been a thread-safe API. New `register`/`deregister` functions are not thread-safe as well.

It *may* be safe, only in case of all calls to `hid_open_path` (and others), including the ones called from within a callback function - are all correctly serialized. But for that:

1) It needs to be explicilty documented (it is not - not right now)
2) It needs to be verified/checked against implementation (I already found multiple issues with libusb implementation (#783), even with regard to functionality that is documented to be working fine)

k1-801: > It _may_ be safe, only in case of all calls to `hid_open_path` (and others), including the ones called from within a callback function - are all correctly serialized. But for that:

So, so long as no other thread opens any device, there is nothing wrong with opening a device from a callback, right?
If so, this limitation is not any different from restricting using this call from any other 2 threads, and should be then worded accordingly :)

As in "It is NOT safe to call hid_open/hid_open_path from within a callback IF any other thread may be calling those functions at the same time (concurrently). It may be safe to do so if it is known that no other thread will be calling hid_open/hid_open_path, otherwise opening the device should be queued, and the device path should be copied by the application to be used and opened later in the queue."

Youw: There are well-known approaches to how the wording for multithreading safety should look like.
I'll get back to this later, one I re-check that it is really safe, and in which cases.

hidapi/hidapi.h
```diff
@@ -358,6 +365,9 @@ extern "C" {
 
 			This function is safe to call from within a hotplug callback.
```
Youw, line 366: Apparently, not really, here is one unresolved issue: https://github.com/libusb/hidapi/pull/674#discussion_r3015415035

Youw:
Opened #790 to expand this doc patch — keeps the intent (document the callback contract) but reworks the wording to match the Multi-threading Notes wiki style, enumerates what is safe to call, fixes the `device->next` / return-value wording, and spins off implementation follow-ups into #792, #793, #794, #795. Suggest closing this one in favour of #790.


---
_Drafted with Claude Code._

Youw:
I already asked Claude to mark all messages as "generated" when thsose are posted on my behalf.

Issue #154, opened 2020-03-27
when I use hidapi, how to use 'HidD_SetNumInputBuffers' to set Buffer size ?


gonboy:
when I use hidapi,
how to use 'HidD_SetNumInputBuffers'  function to set Buffer size ?

gonboy:
I found the file `hidapi/windows/hid.c`, include codes

```cpp
HID_API_EXPORT hid_device * HID_API_CALL hid_open_path(const char *path)
{
....
	/* Set the Input Report buffer size to 64 reports. */
	res = HidD_SetNumInputBuffers(dev->device_handle, 64);
	if (!res) {
		register_error(dev, "HidD_SetNumInputBuffers");
		goto err;
	}
...
}
```
how to call it `HidD_SetNumInputBuffers`

PR #787, opened 2026-04-17
Add hid_set_num_input_buffers() API

auxcorelabs:
Exposes a new public function `hid_set_num_input_buffers(dev, num_buffers)` to resize the per-device input report queue.

High-throughput HID devices (medical telemetry, high-poll-rate gaming peripherals, data acquisition hardware) can emit bursts of input reports that exceed the current hardcoded queue sizes (30 on macOS and Linux/libusb, 64 on Windows). When a burst exceeds the queue, reports are silently dropped with no error indication to the caller. Observed on hardware emitting ~90 reports in 200 ms bursts: both the macOS and Linux/libusb backends silently drop ~20% of frames.

### Per-backend behavior

- **macOS**: resizes the userspace input-report queue
- **Linux libusb**: resizes the userspace report queue
- **Windows**: wraps `HidD_SetNumInputBuffers` (parity with existing behavior)
- **Linux hidraw / NetBSD**: no-op (kernel manages buffering — documented as a `@note` on the API declaration in `hidapi.h`)

### Safety

- Values outside `[1, HID_API_MAX_NUM_INPUT_BUFFERS]` return -1 and register a descriptive error on the device (via the backend's `register_device_error` helper where available)
- Thread-safe: setter uses the same lock (`dev->mutex` / `dev->thread_state`) as the respective report callback

### Design notes

- **Backwards compatibility:** Purely additive public API. Defaults per backend are unchanged, so existing callers see no behavioral change. `hid_device` is opaque in `hidapi.h`, so struct field additions in backend-private files are not an ABI break.
- **Why 1024 cap:** Bounds per-device memory to roughly 1 MB on USB high-speed links, preventing memory exhaustion from a caller that passes `INT_MAX` (malicious or buggy). We've measured bursts up to ~90 reports deep in practice; 1024 leaves room for devices ~10× more demanding. Overridable at build time via `-DHID_API_MAX_NUM_INPUT_BUFFERS=N` for memory-constrained targets.
- **No getter:** hidraw and NetBSD back the queue in the kernel with sizes not portably exposed to userspace, so a cross-platform getter couldn't return a meaningful value there.
- **Why not `hid_get_input_report()` instead?** That API issues a host-initiated `GET_REPORT` request (via the control endpoint or OS equivalent). It does not drain the interrupt-endpoint queue that streaming devices fill, and in practice many devices either don't implement `GET_REPORT` for input reports or respond incorrectly. The two APIs use different USB transfer mechanisms (interrupt endpoint vs. control endpoint); this PR fixes the interrupt-streaming path.

### Userspace queue: flat pre-allocated ring buffer

The userspace queue is a fixed-size ring with inline-slot storage:

- **Pre-allocated, not MAX-sized.** Backing store is a single allocation holding the lengths region (`capacity × sizeof(size_t)` bytes) followed by the slot data region (`capacity × slot_size` bytes), allocated once at device open. The struct holds three pointers: `storage` owns the allocation, while `lengths` and `data` are non-owning views into the lengths and slot-data regions respectively.
- **slot_size determined at runtime per backend:**
  - **macOS** — `kIOHIDMaxInputReportSizeKey` (IOKit-parsed HID report descriptor; authoritative)
  - **libusb** — interrupt-IN `wMaxPacketSize` (matches the existing single-packet transfer sizing)
  - 64-byte clamp on both backends as a defensive fallback when the OS reports no value
- **O(1) push/pop.** Push is `memcpy`-only (no per-report `malloc`); pop is copy-out (`pop_into(r, dst, dst_len)`) — the ring owns its storage end-to-end, no borrowed-pointer contract.
- **Resize.** `hidapi_input_ring_resize()` allocates new storage, memcpys survivors in FIFO order, frees old, swaps under the caller-held mutex. Shrink-below-count drops oldest, matching push-time policy.
- **Implementation.** Single shared header `core/hidapi_input_ring.h`; backend `.c` files include it as a static-in-header helper.
- **Tests.** `tests/test_hidapi_input_ring.c` — test cases covering init/destroy idempotency, push/pop FIFO, drop-oldest, varying lengths, resize grow/shrink/wrap/empty/post-wrap/invalid-args, plus a 10k-iteration stress run. 

### References

- #154 — this PR also addresses the prior request to expose `HidD_SetNumInputBuffers` functionality to callers

Assisted-by: Claude Code:opus-4.6, Claude Code:opus-4.7, Codex:gpt-5.4


linux/hid.c
```diff
@@ -1199,6 +1199,24 @@ HID_API_EXPORT const wchar_t * HID_API_CALL  hid_read_error(hid_device *dev)
 	return dev->last_read_error_str;
 }
 
+
+int HID_API_EXPORT hid_set_input_report_buffer_size(hid_device *dev, int buffer_size)
+{
+	if (!dev) {
+		register_global_error("Device is NULL");
+		return -1;
+	}
```
Youw: remove this check (from all backends) - device functions do not check for validity of the device handle
the only exception is `hid_close`

linux/hid.c
```diff
@@ -1199,6 +1199,24 @@ HID_API_EXPORT const wchar_t * HID_API_CALL  hid_read_error(hid_device *dev)
...
+	if (buffer_size <= 0 || buffer_size > HID_API_MAX_INPUT_REPORT_BUFFER_SIZE) {
+		register_error_str(&dev->last_error_str, "buffer_size out of range");
+		return -1;
+	}
+	/* hidraw: kernel manages the input report buffer, no userspace queue
+	   to resize. Accept the call to preserve a consistent cross-platform
+	   API so callers do not need per-backend conditional code. */
```
Youw: too bad we don't have a logging mechanism
this looks like a perfect spot for a warning
making this completely silent might confuse some users

anyway - even though documentation mentiones that there is no user-space queue for linux backend - we should explicitly document that this function for linux and bsd backend is a no-op

mac/hid.c
```diff
@@ -1347,6 +1349,23 @@ HID_API_EXPORT const wchar_t * HID_API_CALL hid_read_error(hid_device *dev)
 	return dev->last_read_error_str;
 }
 
+
+int HID_API_EXPORT hid_set_input_report_buffer_size(hid_device *dev, int buffer_size)
+{
+	if (!dev) {
+		register_global_error("Device is NULL");
+		return -1;
+	}
```
Youw: remove

netbsd/hid.c
```diff
@@ -955,6 +955,24 @@ HID_API_EXPORT const wchar_t* HID_API_CALL hid_read_error(hid_device *dev)
 	return dev->last_read_error_str;
 }
 
+
+int HID_API_EXPORT HID_API_CALL hid_set_input_report_buffer_size(hid_device *dev, int buffer_size)
+{
+	if (!dev) {
+		register_global_error("Device is NULL");
+		return -1;
+	}
```
Youw: remove

windows/hid.c
```diff
@@ -1257,6 +1257,24 @@ HID_API_EXPORT const wchar_t * HID_API_CALL hid_read_error(hid_device *dev)
 	return dev->last_read_error_str;
 }
 
+
+int HID_API_EXPORT HID_API_CALL hid_set_input_report_buffer_size(hid_device *dev, int buffer_size)
+{
+	if (!dev) {
+		register_global_error(L"Device is NULL");
+		return -1;
+	}
```
Youw: remove

Youw (commented):
refer to comments, otherwise looks good

linux/hid.c
auxcorelabs: Thanks. done.

mac/hid.c
auxcorelabs: Thanks. done.

netbsd/hid.c
auxcorelabs: Thanks done.

windows/hid.c
auxcorelabs: Thanks. done.

linux/hid.c
auxcorelabs: Have updated the comment to explicitly call out no-op. Thanks

auxcorelabs:
Thanks for reviewing. I have made the changes.

JoergAtGithub:
I think the name of the Win32-API `HidD_SetNumInputBuffers` describes better what it does than `hid_set_input_report_buffer_size`, as it does change the number of buffers and not the size of the bufferin bytes.

auxcorelabs:
Good callout @JoergAtGithub. Have renamed the function to `hid_set_num_input_buffers()` and made the associated changes.

@Youw  - Please let me know if any further changes required.

Thanks both.

mcuee:
@auxcorelabs 
Just wondering if you can enhance the hidtest application to include this new API.

If not, do you have a simple test to share? Thanks.

linux/hid.c
```diff
@@ -1199,6 +1199,21 @@ HID_API_EXPORT const wchar_t * HID_API_CALL  hid_read_error(hid_device *dev)
...
+int HID_API_EXPORT hid_set_num_input_buffers(hid_device *dev, int num_buffers)
+{
+	if (num_buffers <= 0 || num_buffers > HID_API_MAX_NUM_INPUT_BUFFERS) {
+		register_error_str(&dev->last_error_str, "num_buffers out of range");
+		return -1;
+	}
+	/* No-op on Linux hidraw and BSD backends: the kernel manages input
+	   report buffering and there is no userspace queue to resize. The
+	   call is accepted (returns 0) to preserve a consistent cross-platform
+	   API so callers do not need per-backend conditional code. */
```
Youw: I meant having this section in the .h as a `NOTE` section.
not much point having it here as verbose versino, specifically no point mentioning other backends from within (like mentioning BSD backend inside of the hidraw implementation)

auxcorelabs: @Youw - have moved this to .h file as a Note section. 

@mcuee  - have updated the hidtest app to include the new API.

Youw (approved):
So far so good.

linux/hid.c
```diff
@@ -1199,6 +1199,17 @@ HID_API_EXPORT const wchar_t * HID_API_CALL  hid_read_error(hid_device *dev)
 	return dev->last_read_error_str;
 }
 
+
+int HID_API_EXPORT hid_set_num_input_buffers(hid_device *dev, int num_buffers)
+{
+	if (num_buffers <= 0 || num_buffers > HID_API_MAX_NUM_INPUT_BUFFERS) {
+		register_error_str(&dev->last_error_str, "num_buffers out of range");
```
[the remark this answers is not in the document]
Youw: agree

netbsd/hid.c
```diff
@@ -955,6 +955,17 @@ HID_API_EXPORT const wchar_t* HID_API_CALL hid_read_error(hid_device *dev)
 	return dev->last_read_error_str;
 }
 
+
+int HID_API_EXPORT HID_API_CALL hid_set_num_input_buffers(hid_device *dev, int num_buffers)
+{
+	if (num_buffers <= 0 || num_buffers > HID_API_MAX_NUM_INPUT_BUFFERS) {
+		register_error_str(&dev->last_error_str, "num_buffers out of range");
```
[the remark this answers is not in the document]
Youw: agree

hidapi/hidapi.h
```diff
@@ -423,6 +423,50 @@ extern "C" {
...
+		    @param dev A device handle returned from hid_open().
+		    @param num_buffers The desired number of input report buffers.
+		        Must be in range [1, HID_API_MAX_NUM_INPUT_BUFFERS].
+
+		    @returns
+		        0 on success, -1 on failure (invalid parameters or
+		        backend-specific error). Call hid_error(dev) for
+		        details where supported.
+		*/
+		int HID_API_EXPORT HID_API_CALL hid_set_num_input_buffers(hid_device *dev, int num_buffers);
```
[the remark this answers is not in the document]
Youw: minor: need update PR description

hidapi/hidapi.h
```diff
@@ -423,6 +423,50 @@ extern "C" {
 		*/
 		int  HID_API_EXPORT HID_API_CALL hid_set_nonblocking(hid_device *dev, int nonblock);
 
+		/** @brief Upper bound for hid_set_num_input_buffers().
+
+		    Values passed above this limit are rejected by
+		    hid_set_num_input_buffers(). Guards against
+		    memory-exhaustion via unbounded input report queue growth.
+		*/
+		#define HID_API_MAX_NUM_INPUT_BUFFERS 1024
```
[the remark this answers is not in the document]
Youw: true - either a PR descriptino or code needs to be updated, I don't really have a preference

mac/hid.c
```diff
@@ -908,7 +910,7 @@ static void hid_report_callback(void *context, IOReturn result, void *sender,
 		/* Pop one off if we've reached 30 in the queue. This
 		   way we don't grow forever if the user never reads
 		   anything from the device. */
-		if (num_queued > 30) {
+		if (num_queued > dev->num_input_buffers) {
 			return_data(dev, NULL, 0);
 		}
```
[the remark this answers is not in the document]
Youw: Someone needs to double-check this, but I'm pretty sure Copilot is right here.
We might even need to consider one step further and switch from linked list to a ring buffer as an optimisation

netbsd/hid.c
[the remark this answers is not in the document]
auxcorelabs: Updated.

linux/hid.c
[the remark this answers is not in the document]
auxcorelabs: Updated.

hidapi/hidapi.h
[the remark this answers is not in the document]
auxcorelabs: Updated PR description.

hidapi/hidapi.h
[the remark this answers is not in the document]
auxcorelabs: Updated code.

libusb/hid.c
```diff
@@ -985,7 +989,7 @@ static void LIBUSB_CALL read_callback(struct libusb_transfer *transfer)
 			/* Pop one off if we've reached 30 in the queue. This
 			   way we don't grow forever if the user never reads
 			   anything from the device. */
-			if (num_queued > 30) {
+			if (num_queued > dev->num_input_buffers) {
 				return_data(dev, NULL, 0);
 			}
```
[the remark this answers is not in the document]
auxcorelabs: Updated to ring buffer

mac/hid.c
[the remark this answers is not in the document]
auxcorelabs: Updated to ring buffer.

mac/hidapi_input_ring.h
Youw, line 1: 1)
is this a known implementation for such a ring buffer?
I'd love to at least basic unit-testing for such kind implementation.

2) We're still having a dynamic allocation each time we try to push/pop, which kind of defeats the purpose of a ring buffer in a first place. Since we know the maximum size of an input report (from a HID Report descriptor), we can pre-allocate a single continues buffer to fit at most `num_input_buffers`.

3) Having two idential implementations copy-pasted in two subdirs is not great. Since we already have it in the separate file, lets place it at common place: `../core/hidapi_input_ring.h`

Youw (commented):
This is an impressive amount of new code contributed in a short amount of time.
What AI tool has been used to generate it? I will not believe it was hand-written.

mac/hidapi_input_ring.h
auxcorelabs: (1a) Known implementation. The ring's core shape — fixed slot array of (pointer, length) entries with head/tail wraparound, per-push payload allocation is closer to Linux kernel's drivers/hid/hidraw.c: same struct hidraw_report { __u8 *value; int len } slot, per-push kmalloc (kmemdup = kmalloc + memcpy) in hidraw_report_event, per-pop kfree in hidraw_read.  Substantive differences are context-driven: bitmask wraparound requires a power-of-2 cap vs ours is build-time configurable, spin_lock_irqsave is mandatory in kernel interrupt context vs we run in userspace and can use pthread_mutex, drop-new vs drop-old. Chose drop-oldest for this PR's streaming-device use case.

(1b) Unit testing - will work on this.

(2) Per-push: low idle memory that grows with actual use; small CPU cost per report VS Flat pre-alloc: more memory always, though reducible at build time via -DHID_API_MAX_NUM_INPUT_BUFFERS=N for constrained targets,  allocation-free on the hot path (lower CPU cost). It's a design choice — happy to go with whichever you prefer.

(3) - Thanks for pointing to the common place. Will move to ../core/hidapi_input_ring.h

auxcorelabs:
> This is an impressive amount of new code contributed in a short amount of time. What AI tool has been used to generate it? I will not believe it was hand-written.

This was a relatively large change, developed using a structured workflow with human oversight & guidance. The design and approach/choices were planned and reviewed (via AI-assisted passes and manually) before any code change. Two AI models were then used in complementary roles—one for code drafting and the other for iterative review to refine the result. The output went through human review and feedback, and the final changes were manually validated, including building across available platforms to verify there were no errors before committing.

From my experience, the effectiveness comes primarily from the structured workflow and repeated AI review iterations, followed by human review & guidance.

mac/hidapi_input_ring.h
Youw: 2 - I suggest to pre-allocate the actual number of buffers requested (or default), and not the `MAX_NUM_INPUT_BUFFERS` - lets have that, since we've already started.
It will make the implementation maybe a little less common, and will require more complicated logic on "resize" action (i.e. allocate new buffer, move/copy existing data from old buffer, then free old buffer), but that's pretty much how std::vector<std::array> would work in C++ (except in our implementation, the size of a report is also determined in runtime - once from the report descriptor.

Youw:
>  the effectiveness comes primarily from the structured workflow and repeated AI review iterations, followed by human review & guidance

I generally don't mind as I'm using similar tools and aproaches myself in various other projects.

But since HIDAPI is open-source well-know public repo with many users and this is a first huge contribution to HIDAPI using AI-tools, I'd like to mention this explicitly in a form suggested by https://docs.kernel.org/process/coding-assistants.html, specifically the commit message/PR description attribution (at least I'm planning to include it in the final squash-merge commit into master), e.g.:
Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]

that's why I asked what AI tool/model was used to produce this.

---
I'll make a generic NOTE about this somewhere in the README a bit later.

mac/hidapi_input_ring.h
auxcorelabs: Switched to flat pre-alloc: backing storage is num_input_buffers × slot_size, allocated at device open — no MAX_NUM_INPUT_BUFFERS waste.
                                                                                                                                          
Slot size sourced at runtime per backend:
  - macOS: `kIOHIDMaxInputReportSizeKey` — IOKit pre-parses the HID report descriptor and exposes the max input size as a property, so no parser needed on our side.                                                                                                              
  - libusb: interrupt-IN `wMaxPacketSize` (matches the existing single-packet transfer sizing).
                                                                                                                                          
Note: #728 (Fixes #274) parses the report descriptor and uses the result to size the libusb transfer buffer. Once it lands, the libusb  slot_size source in our PR can switch to the same dev->max_input_report_size field it introduces — should be a relatively small change. I've left a short Enhancement: comment in `hidapi_initialize_device() `flagging the integration point.                
                                                                                                                                          
Unit tests added in tests/test_hidapi_input_ring.c.

auxcorelabs:
> Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]

Thanks for the information. Please find details below:
Assisted-by: Claude Code CLI:claude-opus-4.6
Assisted-by: Claude Code CLI:claude-opus-4.7                                                                                            
Assisted-by: Codex CLI:gpt-5.4

mac/hid.c
```diff
@@ -877,48 +867,20 @@ static void hid_report_callback(void *context, IOReturn result, void *sender,
...
-		if (num_queued > 30) {
-			return_data(dev, NULL, 0);
-		}
+	int push_rc = hidapi_input_ring_push(&dev->input_ring,
+	                                     report,
+	                                     (size_t)report_length);
+	if (push_rc == 0) {
+		pthread_cond_signal(&dev->condition);
+	} else {
+		register_device_error(dev, "input queue allocation failed");
```
Youw: as per thread-safety notes posted on Wiki - `register_device_error` itself is not thread-safe (and not intended to be) and accessing error string from different threads (`hid_report_callback` thread in this case) can cause a data race

core/hidapi_input_ring.h
```diff
@@ -0,0 +1,254 @@
...
+    if (!r || capacity < 1 || slot_size < 1)
+        return -1;
+    if (r->storage)      /* double-init guard */
+        return -1;
+    /* Reject capacity × slot_size that would overflow size_t. */
+    if ((size_t)capacity > SIZE_MAX / slot_size)
+        return -1;
+
+    r->storage = (uint8_t *)malloc((size_t)capacity * slot_size);
+    r->lengths = (size_t *)calloc((size_t)capacity, sizeof(size_t));
```
Youw: I believe this doesn't have to be a separate allocation either. There could be a single `storage` and two non-owning poiinters to `lengths` and `data`

Youw (commented):
This looks great, thanks!

I did a quick review, but I want to go over it again what I have a bit more free time than now.

Thanks for the contribution! I'll have it merged some time soon.

mac/hid.c
auxcorelabs: Done. Dropped register_device_error() from the hid_report_callback thread.

core/hidapi_input_ring.h
auxcorelabs: Done. Updated.

auxcorelabs:
> This looks great, thanks!
> 
> I did a quick review, but I want to go over it again what I have a bit more free time than now.
> 
> Thanks for the contribution! I'll have it merged some time soon.

Thanks!

auxcorelabs:
@Youw - just wanted to check whether you’ve had a chance to review this. Thanks.

mcuee:
Somehow the added test is not built under Windows, only libusb and macOS.

auxcorelabs:
> Somehow the added test is not built under Windows, only libusb and macOS.

@mcuee - The tests/test_hidapi_input_ring.c are unit tests, all of them exercising the ring buffer (hidapi_input_ring.h). The ring buffer only links into mac and libusb backends — Windows wraps HidD_SetNumInputBuffers natively (our hid_set_num_input_buffers is a thin wrapper there). Hence these tests were gated to only libusb and mac.                                                       
                                                                                                                                                                             
Since hidtest/test.c builds everywhere, that's where symbol-availability check and a runtime smoke call to hid_set_num_input_buffers were added.

mcuee:
@auxcorelabs 

I see. Thanks for the explanation.

PR #813, opened 2026-05-21
Protect Against Potential Memory L eak and IWYU Cleanups

JHartzer:
## Description

This PR resolves a potential memory leak identified via static analysis in the Linux backend. It also implements several IWYU header fixes.

## Problems Identified

1. Linux Memory Leak: There is a potential memory leak in `linux/hid.c` `parse_uevent_info` where multiple matching `HID_UNIQ` or `HID_NAME` keys in a `uevent` file could overwrite previously allocated buffers without freeing them first.
2. Transitive Include Fragility: Several source files (`libusb/hid.c` and `mac/hid.c`) relied on transitive inclusions for standard types like `uint8_t`, standard functions like `calloc`, and error definitions like `errno`. If parent headers are refactored, these sources will break.
3. Unused Headers: Unused headers (such as `<sys/utsname.h>` and `<linux/version.h>`) were included in multiple places, unnecessarily increasing compile times.

## Proposed Changes

- Linux Memory Leak Fix
   - Updated the Linux backend `linux/hid.c` to safely free any previously allocated memory before saving a new device property. This stops memory from leaking if a device's info key (like name or serial) is parsed more than once.
- IWYU (Include-What-You-Use) Header Cleanups
   - Added missing headers (like `<stdint.h>`, `<stdlib.h>`, and `<errno.h>`) across the test utility, macOS, and LibUSB backends so they don't crash or break when parent files are updated.
   - Removed unused headers (like `<sys/utsname.h>` and `<linux/version.h>` ) that were cluttering the code and slowing down compile times.

### 1. Automated Compilation

Clean out-of-source builds were configured and executed using CMake:

```
cmake -B build -S . -DHIDAPI_WITH_LIBUSB=OFF -DHIDAPI_BUILD_HIDTEST=ON
cmake --build build
```

- Result: All targets compiled with zero warnings and zero errors.

### 2. Static Analysis

A verification pass using clang-tidy was conducted on the Linux backend:

```clang-tidy -p build-tidy linux/hid.c```

- Result: The Potential memory leak warning has been successfully resolved.

### 3. Functional Verification

The compiled test utility `hidtest_hidraw` was run successfully. It executed the `hid_enumerate()` discovery loop using our modified `parse_uevent_info` implementation, successfully extracting and displaying all connected USB device attributes

### CODE REVIEW

File: hidtest/test.c [hunk 1]
```diff
@@ -19,7 +19,6 @@
 #include <wchar.h>
 #include <string.h>
 #include <stdlib.h>
-
 #include <hidapi.h>
```
Youw, line 23: I don't see anything wrong if we separate visually system headers and our library

Youw (approved):
One comment, everything else looks good, thanks

hidtest/test.c
JHartzer: Yes that makes sense! I've reverted it now. Thanks for the review!

JHartzer:
Hey @Youw! I just pushed up a suggested fix for the [windows-cmake](https://github.com/libusb/hidapi/actions/runs/27485969769/job/81242157882#step:11:23) build failures. It looks like `windows-latest` runner was updated to a newer Visual Studio version (in this case, Visual Studio 2026 / VS 18), so the hard-coded VS path didn't exist. This seems to be a standard, flexible way to find the Visual Studio installation path.

Youw:
Thanks.

I'd like to keep these things separated. Let this PR be only about the code and not about the CI. Temporary failed MSVC build will not stop me from merging it (CI issue is not caused by the PR, so it is fine).

Regarding MSVC build fix/improvement - I agree, your suggested fix is so much better than the hard-coded version that is now, but I'd rather make even one more step further, and use [setup-msvc-dev](https://github.com/marketplace/actions/setup-msvc-developer-command-prompt?version=v4.0.0) Github Action.
But thatshould be a separate PR a separte PR.

Youw (approved):

PR #817, opened 2026-06-14
ci: use setup-msvc-dev action instead of a hard-coded vcvars path

Youw:
## Summary

The `windows-cmake` job set up the MSVC environment by `call`-ing a hard-coded path:

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat
```

in its NMake MSVC, NMake clang-cl and Meson steps. The hosted `windows-latest` image no longer has Visual Studio under that **Enterprise** path, so as of recently every one of those steps fails at the first `call`:

```
The system cannot find the path specified.
```

(`windows-msbuild` is unaffected — it doesn't call `vcvars`.)

This sets up the MSVC developer environment once per job with the [setup-msvc-dev](https://github.com/marketplace/actions/setup-msvc-developer-command-prompt) action — it locates the installed toolchain via `vswhere` (no hard-coded path or VS edition) — and removes all seven `vcvars64.bat` calls.

Since that environment is job-wide, the **MinGW** build (which must use `gcc`, not `cl`) is split out into its own **`windows-mingw`** job that runs without the MSVC environment. The MSVC, NMake, clang-cl and Meson builds stay in `windows-cmake` with the action.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -193,6 +193,10 @@ jobs:
         choco install ninja
         pip3 install meson
         refreshenv
+    - name: Set up MSVC developer environment
+      uses: TheMrMilchmann/setup-msvc-dev@v4
+      with:
+        arch: x64
     - name: Configure CMake MSVC
       shell: cmd
       run: |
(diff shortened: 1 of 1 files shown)


Youw:
Heads-up: #813 also fixes this same Windows CI break in `builds.yml`, via a different mechanism — it adds a `vswhere`-based step that exports `VCVARS_PATH` and replaces the hard-coded path with `call "%VCVARS_PATH%"`. This PR instead drops the manual `vcvars` calls entirely and sets the environment up once per job with the `setup-msvc-dev` action.

So the two overlap on `builds.yml`: whichever merges first, the other's `builds.yml` change should be dropped on rebase. #813's other changes (the `linux/hid.c` memory-leak fix and the IWYU header cleanups) are unrelated to CI and unaffected.

(Note: #813 also touches `mac/hid.c`, which overlaps #816 — worth coordinating that rebase too.)

_Drafted with Claude Code._

.github/workflows/builds.yml
```diff
@@ -306,6 +304,25 @@ jobs:
...
 
+    - name: Check Meson build
+      shell: cmd
+      run: |
+        meson setup build_meson hidapisrc
+        cd build_meson
+        ninja
+
+
+  windows-mingw:
```
Youw:
```suggestion
  windows-cmake-mingw:
```

.github/workflows/builds.yml
Youw:
```suggestion
  windows-cmake-msvc:
```

PR #819, opened 2026-06-14
appveyor: derive the build version from the repo

Youw:
## Summary

The AppVeyor **"Build version format"** is currently hard-coded in the project's web settings. This moves version control into the repo:

- adds `version: '{build}'` to `.appveyor.yml` (a `version:` in the file overrides the UI setting), and
- in a `before_build` step, replaces it with a version derived from the repo — the `VERSION` file + the short commit hash + the build number:

```
0.16.0-1a2b3c4.42
```

`Update-AppveyorBuild` runs after checkout (so the `VERSION` file is available); `APPVEYOR_REPO_COMMIT` provides the commit and `APPVEYOR_BUILD_NUMBER` keeps each build's version unique even if the same commit is built more than once.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,3 +1,9 @@
+# The build version is derived from the repo here, which overrides the
+# hard-coded "Build version format" in the AppVeyor project settings. The
+# {build} placeholder is replaced in before_build with the version read from
+# the VERSION file plus the short commit hash (see below).
+version: '{build}'
+
 environment:
   matrix:
   - BUILD_ENV: msbuild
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -6,6 +12,15 @@ environment:
     arch: Win32
   - BUILD_ENV: cygwin
 
+# Derive the build version from the repo: <VERSION>-<short commit>.<build number>,
+# e.g. 0.16.0-1a2b3c4.42. The build number keeps it unique even if the same
+# commit is built more than once.
+before_build:
+  - ps: |
+      $base  = (Get-Content -Raw "$env:APPVEYOR_BUILD_FOLDER/VERSION").Trim()
+      $short = $env:APPVEYOR_REPO_COMMIT.Substring(0, 7)
```
Youw: the hard-coded `7` might bite us in general case. Is there a way to ask git to give the short version?

Youw: Done in `bd64102` — switched to `git rev-parse --short HEAD`, so git picks the (unambiguous) short length itself instead of a hard-coded 7. Verified on AppVeyor: the build set the version to `0.16.0-bd64102.2082`.

_Drafted with Claude Code._

Issue #793, opened 2026-04-24
Hotplug: HID_API_HOTPLUG_ENUMERATE should honour callback return value


Youw:
## Summary

The hotplug callback documentation added in #790 specifies:

> Return 0 to keep the callback registered. Return any non-zero value to have HIDAPI deregister the callback; no further events for this handle will be delivered, and the handle is freed as if hid_hotplug_deregister_callback() had been called. **This applies to both live events and to the synthetic \"arrived\" events delivered during registration when HID_API_HOTPLUG_ENUMERATE is set.**

The current implementation discards the return value during the synthetic initial-enumeration pass. In \`libusb/hid.c\` for example: https://github.com/libusb/hidapi/blob/connection-callback/libusb/hid.c#L1393

\`\`\`c
(*hotplug_cb->callback)(hotplug_cb->handle, device, HID_API_HOTPLUG_EVENT_DEVICE_ARRIVED, hotplug_cb->user_data);
\`\`\`

— return value is assigned to nothing; the callback cannot self-deregister during initial enumeration.

## Why this matters

The natural idiom \"find the first device matching this VID/PID, react to it, and stop listening\" — \`return 1\` on match — silently fails when \`HID_API_HOTPLUG_ENUMERATE\` delivers the first match synchronously during \`hid_hotplug_register_callback()\`. Users write the obvious thing and get unexpected extra invocations.

## Proposed fix

Treat the synthetic arrival events identically to live events: capture the return value, mark the callback for removal if non-zero, and stop delivering further synthetic events for it. All four backends share the same pattern.

## Scope

- \`libusb/hid.c\`, \`linux/hid.c\`, \`mac/hid.c\`, \`windows/hid.c\`
- hidtest: consider a test case that self-deregisters from an ENUMERATE event.

## Related

- Documentation: #790
- Original RFC: #674

---
_Drafted with Claude Code._

Issue #792, opened 2026-04-24
Hotplug: device->next must be NULL inside callback


Youw:
## Summary

The hotplug callback documentation added in #790 specifies:

> The device->next pointer is always NULL. Each callback invocation describes exactly one device; compound or composite devices that expose multiple interfaces produce multiple callback invocations (typically delivered in quick succession).

The current implementation does not enforce this on all backends. The libusb backend has an explicit TODO acknowledging it — https://github.com/libusb/hidapi/blob/connection-callback/libusb/hid.c#L1123 — and when the enumerated list is iterated each entry is passed to the callback with its sibling \`next\` pointer still set, exposing the internal list layout to user code.

## Proposed fix

In each backend's callback dispatch path, copy the \`hid_device_info\` into a local struct with \`.next = NULL\` before invoking the user callback:

\`\`\`c
struct hid_device_info info_copy = *info;
info_copy.next = NULL;
callback(handle, &info_copy, event, user_data);
\`\`\`

No strings need to be duplicated (they remain valid for the duration of the callback). This prevents user code from traversing HIDAPI's internal list and keeps the public contract honest.

## Scope

- \`libusb/hid.c\` (\`hid_internal_invoke_callbacks\` and the \`HID_API_HOTPLUG_ENUMERATE\` path in \`hid_hotplug_register_callback\`).
- \`linux/hid.c\`, \`mac/hid.c\`, \`windows/hid.c\` — same audit.

## Related

- Documentation: #790
- Original RFC: #674

---
_Drafted with Claude Code._

Issue #791, opened 2026-04-24
Hotplug: device->next must be NULL in callback


Youw:
Tracking from #790 / #674.

### Contract (documented in #790)

> The `device->next` pointer is always NULL. Each callback invocation describes exactly one device; compound or composite devices that expose multiple interfaces produce multiple callback invocations (typically delivered in quick succession).

### Current state

Backends invoke the user callback with a `hid_device_info*` whose `next` field still points into the backend-internal enumeration list. Example: [libusb/hid.c:1123](https://github.com/libusb/hidapi/blob/connection-callback/libusb/hid.c#L1123) already carries a \`TODO\` for this:

\`\`\`c
/* TODO: possibly make the \`next\` field NULL to match the behavior on other systems */
\`\`\`

The other backends are similarly inconsistent.

### Proposed fix

Before dispatching the callback, copy the \`hid_device_info\` into a local struct on the dispatcher's stack with \`.next = NULL\`, then pass a pointer to that local struct into the callback. Do not deep-copy the string fields — they are still owned by HIDAPI and remain valid for the duration of the callback.

This avoids exposing the internal list layout to user code and gives one consistent cross-backend contract.

### Backends to touch

- [ ] \`libusb/hid.c\` — \`hid_internal_invoke_callbacks\`
- [ ] \`linux/hid.c\`
- [ ] \`mac/hid.c\`
- [ ] \`windows/hid.c\`

### Related

- #674 (hotplug implementation)
- #784 (original doc patch, superseded)
- #790 (documentation of the target contract)

Issue #238, opened 2021-01-26
Add hotplug support


mcuee:
One of the main missing feature of hidapi is the hotplug support.

mcuee:
Technically Windows codes are out there. For Linux you can consider libudev. For macOS there are similar codes as well. libusb has hotplug support under Linux and macOS but not Windows.

mcuee:
For Windows, not so sure pywinusb can shed some light.
https://github.com/rene-aguirre/pywinusb

Or the codes here.
http://janaxelson.com/hidpage.htm

I think Hidsharp also supports hotplug.
https://github.com/IntergatedCircuits/HidSharp

Youw:
I'd say that hotplug support for each platform is pretty much straight-forward (I've done it for macOS and Windows several years ago in a propriatary project)
same is true for linux (udev) and libusb backends

it just has to be done

mcuee:
#299 seems to work pretty well under Windows. The API function names are now similar to libusb and I think they are fine.

807183087:
Hi,sir. I found  a little error about nullptr in https://github.com/DJm00n/hidapi/blob/device_notify/windows/hid.c#L1036
![before change](https://user-images.githubusercontent.com/30239482/147877780-79c574e0-5f6f-44e0-9d5b-47ec8b5f4ff3.png)
![after change](https://user-images.githubusercontent.com/30239482/147877783-1c054683-85c9-4f92-ad21-54cfffc5987f.png)

807183087:
I noticed, the plugged devices before my testing program startup also can be callback.
I think, hotplug just processes the hot plug events, but not existed devices.

Youw:
> I think, hotplug just processes the hot plug events, but not existed devices.

Once hotplug is registered - it has to report all existing devices.

The likely alternative you probably want to have is:
1) enumerate existing devices;
2) register callback for new devices;

Or the other way around (first register callback, then enumerate existing devices).

In either case you're having a race condition if the device is plugged at the moment you're registering a callback (and that is happening more often than you could imagine).
You either miss a device or have it reported twice.

The only 100% reliable way to use device hotplug events - is to use _only hotplug callback_ events to get a list of all devices (existing and new).
And not use things like `hid_enumerate` at all.

807183087:
yes.
regeste callback like this will not receive existed devices:

[...]

PR #822, opened 2026-07-13
linux: deliver hotplug ENUMERATE pass asynchronously on the event context

Youw:
Implements the hotplug contract documented in `hidapi/hidapi.h` (#790) for the linux/hidraw backend.

## What changed

**Asynchronous ENUMERATE pass.** `hid_hotplug_register_callback()` no longer runs the `HID_API_HOTPLUG_ENUMERATE` initial pass synchronously on the calling thread. Instead it takes a deep-copied, registration-time snapshot of the matching entries of the internal device cache (only when `HID_API_HOTPLUG_EVENT_DEVICE_ARRIVED` is requested), stores it on the callback record (`replay`), and the udev monitor thread delivers it as synthetic "arrived" events:

- never on an application thread and never from within the register call itself;
- before any live events for that callback вЂ” `hid_internal_invoke_callbacks()` flushes a callback's pending snapshot before handing it a live event, and the monitor thread also flushes pending snapshots at the top of its loop (the existing 5 ms poll timeout caps the latency, no extra wakeup fd needed);
- exactly once per device connection: the snapshot is taken under the hotplug mutex, so a device arrival is reported either by the pass or as a live event, never both;
- a non-zero callback return stops the remainder of the pass and deregisters the callback (undelivered snapshot entries are freed), same as for live events.

This also covers registrations made while the monitor thread is already running, including registrations made from within another hotplug callback.

**`device->next` is now NULL for every invocation.** The udev arrival path used to pass the chained multi-usage list to callbacks (the old TODO); each entry is now delivered on its own, keeping the internal cache intact for later "left" matching.

**Error strings and out-parameter.** Every failure path of register/deregister (NULL callback, empty/unknown events bits, unknown flags bits, allocation/udev/thread-start failures, unknown or stale deregister handle) now sets the global error string retrievable via `hid_error(NULL)`, and register sets `*callback_handle` to 0 on failure before any error return.

**Monitor-thread lifecycle fixes.** Previously, when the last callback removed itself on the monitor thread (non-zero return or in-callback deregister), the thread exited on its own, was never joined, and freed the shared device list / udev monitor unlocked вЂ” racing a subsequent first registration that re-creates them. Now the exiting thread releases nothing; joining and monitoring-context teardown are centralized in `hid_internal_hotplug_cleanup()` and a new first registration reaps the previous thread before re-creating the context.

## Design decisions / deviations

- `hid_enumerate(0, 0)` inside the first registration may set the "No HID devices found" global error on an empty system; register clears it, since an empty system is not a registration failure. A *genuine* enumeration failure does fail the registration.
- The initial pass is all-or-nothing: if the snapshot cannot be copied in full, the registration is unwound and fails, rather than delivering a partial device list (each connection must be reported exactly once вЂ” never "neither").

## Verification

- `gcc -fsyntax-only -Wall -Wextra -pedantic -Werror -Wformat-signedness` and `g++ -fsyntax-only -Wall -Wextra -Werror` clean; full CMake builds (C and `HIDAPI_BUILD_AS_CXX=ON` with `-Werror`) pass.
- The WSL2 kernel used for testing has no uhid/hidraw support (`CONFIG_UHID` not set), so the #815 virtual-device harness could not drive real udev events. Instead a throwaway whitebox harness (including `linux/hid.c` directly, injecting fake entries into the hotplug device cache and driving the internal arrival/removal handlers) exercised 119 checks: async delivery on the monitor thread (never the registering thread), VID/PID filtering of the snapshot, `next == NULL` on every invocation, exactly-once vs. later-connected devices and vs. a pre-queued add, non-zero return stopping the pass, self-deregistration and nested registration from within a callback, registration from within an in-flight dispatch (both event types, multi-usage devices), tombstoned/stale-handle deregister, handle exhaustion, last-callback self-removal followed by re-registration (thread reaping), `hid_exit()` with an undelivered snapshot, re-initialization after `hid_exit()`, concurrent first registrations, concurrent stale deregistrations, and all argument-validation error strings.
- 10/10 runs clean under ASan+UBSan+LSan (no leaks) and 10/10 clean under TSan.
- Live udev arrival/removal delivery was not exercised against real hardware and still deserves a check on a real machine.

Addresses #792 and #793 for the linux/hidraw backend.

## Review round 1

All findings from the first review round are addressed in `linux: address the hotplug review findings`:

- **Global error-string race.** Register/deregister wrote `last_global_error_str` with no lock (on the validation paths, before the hotplug mutex exists, and on the post-unlock paths), and the monitor thread wrote it too (via `create_device_info_for_device()`); two concurrent, documented-thread-safe calls could double-free it. Every mutation now goes through a file-static mutex in the lowest-level writer (`register_error_str()`), which all the `register_global_error*()` / `register_device_error*()` helpers funnel into. Reading via `hid_error()` remains subject to the documented thread-safety rules.
- **First-initialization race.** Two concurrent first registrations could both observe `mutex_ready == 0` and both `pthread_mutex_init()` the hotplug mutex (undefined behaviour) while resetting the context fields under each other. A static bootstrap mutex now guards the machinery init/teardown block.
- **`hid_init()` race (found while making TSan clean).** A registration implicitly initializes the library twice вЂ” directly, and again inside the initial enumeration вЂ” and `setlocale()` is not thread-safe against itself, so two concurrent registrations raced in `hid_init()`. `hid_init()` now serializes itself.
- **Liveness / publication.** The monitor loop no longer reads any decision state unlocked: it `trylock`s every iteration and checks the callback list (and the pending initial passes) only under the acquired mutex, so a stale read cannot keep it alive or make it exit early. The thread announces its exit under the mutex, and `hid_internal_hotplug_cleanup()` unlocks before `pthread_join()` вЂ” nothing is ever joined while holding the mutex. This removed the previously "excused" races, and TSan is now fully clean.
- **Duplicate arrival.** A device showing up between `udev_monitor_enable_receiving()` and the initial `hid_enumerate()` was in both the snapshot and the queued netlink event, producing a duplicate "arrived", a duplicate cache entry and later a duplicate "left". Arrivals whose path is already cached are now skipped.
- **Dispatch bounding and cache order.** The device cache is updated *before* an event is dispatched, and every dispatch (both event types) is bounded to the callbacks registered when the event started. A callback registered from within a callback therefore observes the device exactly once, through its own snapshot, instead of also receiving the in-flight event вЂ” and, for a multi-usage device, all of its usage entries are added to (or removed from) the cache before any of them is dispatched, so such a callback can neither miss a device nor see a "left" with no matching "arrived".
- **All-or-nothing snapshot.** A failed snapshot copy now frees the partial snapshot, unwinds the registration completely (leaving `*callback_handle` at 0), sets the error string and returns -1. A partial `hid_internal_copy_device_info()` result is likewise rejected instead of reaching a callback. A genuine `hid_enumerate()` failure now fails the registration, and is distinguished from an empty system.
- **libudev setup and fd 0.** Every libudev call in the registration path is checked and fails the registration with an error string. `0` is a valid file descriptor: `-1` is now the sentinel, and the checks test for `< 0`.
- **Handle exhaustion.** `next_handle++` at `INT_MAX` was signed-overflow undefined behaviour, and wrapping back to 1 could collide with a live handle. Registration now fails with an error instead of wrapping.
- **Tombstoned handle.** Deregistering a handle that was already deregistered from within a callback (and is only awaiting its postponed removal) now reports "not found" (-1) instead of 0.
- **Early return in deregister.** Deregistration now reaps a previously self-exited monitor thread before returning early on an empty callback list, instead of leaving the monitoring context behind until the next registration or `hid_exit()`.
- **Robustness.** The udev action and devnode are checked for NULL, and `poll()` replaces `select()` for the monitor file descriptor (no `FD_SETSIZE` undefined behaviour).

---

The narrative above traces the implementation; the merged design is: a udev-monitor thread delivering an asynchronous, registration-time replay snapshot; a process-lifetime recursive mutex; and a **retired-thread-list** lifecycle in which every monitor generation ever created is tracked and joined before `hid_exit()` returns (identity is a monotonic token, never a post-join `pthread_t`). On an unrecoverable udev-monitor failure the backend **stops cleanly** — no fabricated `DEVICE_LEFT`, cache and open handles untouched, new registrations refused. (An earlier monitor auto-recovery attempt was dropped as unfixable; recorded as a possible future enhancement in #828.)

Implements the `device->next == NULL` (#791, #792) and ENUMERATE return-value (#793) contract for hidraw. Part of hotplug support (#238); contract documented in #790. Follow-up: #828.

### DIFF
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -1202,31 +1202,30 @@ static void hid_internal_hotplug_retire_current(void)
 	hid_hotplug_context.thread_id = 0;
 }
 
-/* Non-zero when the caller runs on one of HIDAPI's monitor threads: the current
-   generation, or a retired one that has not been claimed for joining. Called
-   with the mutex held.
+/* The stable id token of the generation this thread IS, or 0 on any other
+   thread. Written by hotplug_thread() as its first action and never cleared, so
+   it identifies the thread for the whole of its life - including the
+   thread-specific-data destructor phase, which runs AFTER the thread published
+   FINISHED and returned. That phase is exactly where a pthread_t comparison
+   stops being usable: a reaper may already have claimed the entry
+   (being_joined) and be sitting in pthread_join(), which returns only once
+   those destructors are done - so being_joined does NOT mean the thread has
+   terminated, and its pthread_t must not be touched either way. The id tokens
+   are monotonic and start at 1, so 0 is never a valid generation; thread-locals
+   are zero-initialized, and a recycled OS thread gets a fresh (zero) one. */
+static __thread unsigned long hid_hotplug_thread_self_id;
+
+/* Non-zero when the caller runs on one of HIDAPI's monitor threads, whatever
+   the state of that thread's generation - running, finished, retired or being
+   joined. Needs no lock: it only reads this thread's own thread-local.
 
    mutex_in_use alone does not catch every such call: it only covers a call made
    from within a dispatch. A thread-specific-data destructor armed by a user
-   callback runs on the monitor thread AFTER the dispatch unwound - and after the
-   thread published FINISHED - and may re-enter the library from there.
-   An entry another reaper claimed (being_joined) is skipped: that thread has
-   terminated, so it cannot be the caller, and its pthread_t must not be
-   inspected once its joiner started (it may already have been invalidated). */
+   callback runs on the monitor thread after the dispatch unwound - and after the
+   thread published FINISHED - and may re-enter the library from there. */
 static int hid_internal_on_monitor_thread(void)
 {
-	if (hid_hotplug_context.thread_state != HID_HOTPLUG_THREAD_NONE
-	    && pthread_equal(pthread_self(), hid_hotplug_context.thread)) {
-		return 1;
-	}
-
-	for (struct hid_hotplug_monitor_thread *entry = hid_hotplug_context.retired; entry != NULL; entry = entry->next) {
-		if (!entry->being_joined && pthread_equal(pthread_self(), entry->thread)) {
-			return 1;
-		}
-	}
-
-	return 0;
+	return hid_hotplug_thread_self_id != 0;
 }
 
 /* Reaps monitor threads: first RETIRES the current generation if it has finished
@@ -1485,15 +1484,19 @@ static void hid_internal_hotplug_exit(void)
 		   register/deregister/hid_exit() from an application thread to join: the
 		   same bounded degradation as hid_exit() from within a callback.
 
-		   A being-joined entry is counted without inspecting its pthread_t (its
-		   thread terminated, so it cannot be the caller anyway) - a pthread_t is
-		   never touched after a join was started on it. */
+		   Self is recognized by the generation id token this thread carries in a
+		   thread-local (hid_hotplug_thread_self_id), NOT by pthread_equal(): the
+		   caller's own entry may already be claimed by a reaper sitting in
+		   pthread_join() - which returns only once this thread's destructors are
+		   done, so being_joined does NOT mean the thread has terminated - and a
+		   claimed pthread_t must not be inspected. Counting that entry as
+		   somebody else's would deadlock both threads: this one would wait for a
+		   broadcast that only its own join completion can send. */
 		if (hid_hotplug_context.retired != NULL) {
 			int others = 0;
 
 			for (struct hid_hotplug_monitor_thread *entry = hid_hotplug_context.retired; entry != NULL; entry = entry->next) {
-				if (entry->being_joined
-				    || !pthread_equal(pthread_self(), entry->thread)) {
+				if (entry->id != hid_hotplug_thread_self_id) {
 					others = 1;
 					break;
 				}
@@ -2032,6 +2035,13 @@ static void* hotplug_thread(void* user_data)
 	   the same: an unsynchronized read here would race the setup of the next
 	   monitoring context (whose registration this thread cannot have seen). */
 	pthread_mutex_lock(&hid_hotplug_context.mutex);
+	/* Learn our own generation id, for the join-safe self-identification the
+	   library needs once this thread's pthread_t may have been claimed for a
+	   join (see hid_hotplug_thread_self_id). The value read here is ours: the
+	   creating registration holds the mutex continuously from pthread_create()
+	   until after it published this id, and a newer generation cannot be
+	   created before this one publishes FINISHED below. */
+	hid_hotplug_thread_self_id = hid_hotplug_context.thread_id;
 	monitor_fd = hid_hotplug_context.monitor_fd;
 	pthread_mutex_unlock(&hid_hotplug_context.mutex);
 
@@ -2376,7 +2386,9 @@ int HID_API_EXPORT HID_API_CALL hid_hotplug_register_callback(unsigned short ven
 			if (monitor_error && !quiet) {
 				register_global_error(monitor_error);
 			}
-			/* on enumerate_failure the global error is already set by the enumeration */
+			/* on enumerate_failure the enumeration has set the global error -
+			   unless quiet suppressed it, in which case this caller is an
+			   internal thread, which must not read it either */
 			return -1;
 		}
 
--- a/linux/hid.c
+++ b/linux/hid.c
@@ -661,7 +661,15 @@ next_line:
 
 /* quiet: don't touch the global error string - for the callers on HIDAPI's
    internal monitor thread, which never writes it (see hidapi.h) */
-static struct hid_device_info * create_device_info_for_device(struct udev_device *raw_dev, int quiet)
+/* Build the hid_device_info chain (one entry per usage) for a single udev device
+   node. Returns NULL both for benign exclusions (no HID parent, an unparseable
+   uevent, an unhandled bus type - exactly what hid_enumerate() would also skip)
+   and for genuine resource failures. When `failure` is non-NULL it is set to 1
+   ONLY in the latter case, so a caller that needs an authoritative, complete
+   enumeration (the monitor-thread reconcile) can tell a device that is
+   legitimately absent from one that merely failed to materialize. It is never
+   cleared here, so callers accumulate it across a whole enumeration. */
+static struct hid_device_info * create_device_info_for_device(struct udev_device *raw_dev, int quiet, int *failure)
 {
 	struct hid_device_info *root = NULL;
 	struct hid_device_info *cur_dev = NULL;
@@ -721,8 +729,14 @@ static struct hid_device_info * create_device_info_for_device(struct udev_device
 
 	/* Create the record. */
 	root = (struct hid_device_info*) calloc(1, sizeof(struct hid_device_info));
-	if (!root)
+	if (!root) {
+		/* A genuine resource failure, unlike the benign NULL returns above:
+		   flag it for callers that require an all-or-nothing enumeration. */
+		if (failure) {
+			*failure = 1;
+		}
 		goto end;
+	}
 
 	cur_dev = root;
 
@@ -855,8 +869,14 @@ static struct hid_device_info * create_device_info_for_device(struct udev_device
 			struct hid_device_info *tmp = (struct hid_device_info*) calloc(1, sizeof(struct hid_device_info));
 			struct hid_device_info *prev_dev = cur_dev;
 
-			if (!tmp)
+			if (!tmp) {
+				/* Out of memory mid-device: the returned chain would be
+				   incomplete, so flag it for all-or-nothing callers. */
+				if (failure) {
+					*failure = 1;
+				}
 				break;
+			}
 			cur_dev->next = tmp;
 			cur_dev = tmp;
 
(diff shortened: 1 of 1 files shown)


Youw:
## Full implementation review (fresh pass, head `72b00d7`)

Traced all hotplug paths against the contract in `hidapi/hidapi.h`: registration (first and subsequent), the replay-snapshot machinery, live dispatch, tombstoning/postponed removal, the monitor-thread loop, the retire/reap/join lifecycle, and `hid_exit` teardown. Each candidate finding was adversarially self-checked before being reported. Overall the implementation is in very good shape — the core contract (registration-time snapshot, exactly-once XOR delivery, replay-before-live ordering, nonzero-return deregistration, deregister-blocks-on-in-flight, handle monotonicity, monitor-death stop-cleanly, retired-thread-list join lifecycle) holds up under adversarial reading. **3 findings survived: 1 major, 2 minor. No blockers.**

### [major] `linux/hid.c:1412-1416` — `hid_exit()` from a TSD destructor on the finished monitor thread self-deadlocks and permanently bricks the hotplug API

The comment at 1412-1413 asserts *"hid_exit() is never the monitor thread, so no retired entry is ever 'self' here"* — but the file explicitly engineers for thread-specific-data destructors re-entering HIDAPI on the just-finished monitor thread (986-988, 2266-2271 support *register* from exactly that context), and `hidapi.h`'s UB carve-out covers only `hid_exit` "from within the callback itself", not a TSD destructor.

Trace: monitor thread publishes FINISHED, its start routine returns, a user TSD destructor calls `hid_exit()` → `hid_internal_hotplug_exit()`: `exiting` is 0, `mutex_in_use` is 0, so it sets `exiting = 1` and calls the reaper, which retires the generation onto `retired` and then *skips it* (`pthread_equal(pthread_self(), ...)`, line 1205). Back at 1414, `retired != NULL` → `pthread_cond_wait()` — forever: the only broadcast source is a join completion (1230), but no other thread can ever reach the reaper, because `register` (2106), `deregister` (2416), and a second `hid_exit` (1356) all bail out early on `exiting == 1`, which is now stuck at 1. The destructor thread hangs, can never be joined, and every future hotplug call fails with "hid_exit() is in progress" for the life of the process.

**Suggested fix:** in the `hid_internal_hotplug_exit()` wait loop, before waiting, scan `retired`: if every remaining entry that is not `being_joined` is `pthread_equal` to self, break out instead of waiting (leaving the self entries for the next register/deregister/`hid_exit` to reap — the same documented bounded-leak degradation used for the in-callback case), so `exiting` resets to 0 and the API stays usable. Only wait on `thread_cond` when a non-self entry is actually being joined by another reaper. Correct the 1412-1413 comment.

### [minor] `linux/hid.c:2080` — `quiet` detection misses the TSD-destructor path: register/deregister on the (finished) monitor thread write the global error string

`quiet` is derived solely from `mutex_in_use` (2080; 2406 in deregister). A registration from a TSD destructor running on the just-finished monitor thread — a path the file explicitly supports (2266-2271) — has `mutex_in_use == 0`, so `quiet == 0`: it calls `hid_init()` (2121, clears the global error) and on failure paths writes `last_global_error_str` (2256, 2303, 2323, plus `register_global_error` inside `hid_internal_enumerate` reached from 2243). That violates the invariant that internal threads never write the global error string — an application thread reading `hid_error(NULL)` (unlocked read, 2895-2897) cannot serialize against a write it doesn't know is happening. The comment at 2205-2210 is false for exactly this case.

**Suggested fix:** under the mutex, extend the check: `quiet = mutex_in_use || on_monitor_thread()`, where `on_monitor_thread()` compares `pthread_self()` via `pthread_equal` against the current generation's thread (when `thread_state != NONE`) and against every `retired` entry that is not `being_joined` (a being-joined thread has terminated and cannot be self; skipping it preserves the never-inspect-a-joined-`pthread_t` discipline). Fix the 2205-2210 comment.

### [minor] `linux/hid.c:746,885-892` — unchecked `strdup`/`wcsdup` results defeat the all-or-nothing `failure` flag and can cache a NULL-path device

`create_device_info_for_device()` sets `*failure` only on the two `calloc` failures (736-739, 873-879), but `cur_dev->path = dev_path ? strdup(dev_path) : NULL` (746) and the strdup/wcsdup calls in the additional-usage loop (885-892) fail silently under OOM. Consequences: (a) the registration-time snapshot (2243), whose stated purpose is to *fail rather than arm the callbacks against an incomplete device set*, can succeed with silently degraded entries; (b) on the monitor-thread arrival path a `path == NULL` entry enters the cache, its ARRIVED is dispatched, but `hid_internal_hotplug_process_removal` (1838) and the duplicate check (1741) can never match it — an "arrived" with no possible "left", and broken same-node dedup, contradicting the exactly-once contract the flag was added to protect.

**Suggested fix:** apply the same partial-copy check already used in `hid_internal_copy_device_info` (1646-1652): after filling an entry, if any source string was non-NULL but its copy is NULL, set `*failure` (and on the monitor-thread arrival path drop the whole device chain instead of caching it).

---

Verified sound (checked, no issues): exactly-once delivery (monitor armed before initial enumeration, path-dedup, `dispatch_bound` handle-ordering for nested registrations); replay-before-live per-callback flush; all snapshot-free paths on deregister/tombstone/exit; thread lifecycle (pre-allocated retirement nodes, joins outside the mutex, self/double-join prevention, no lost wakeups); lock ordering (hotplug mutex → `hid_init_mutex` → `error_str_mutex`); EINTR/ENOBUFS/POLLERR handling with the 100-poll limit; monitor-death stop-cleanly semantics; udev-unavailable failure path.

_Drafted with Claude Code._

Youw:
## Re-review of fix commits `97760f9` + `18434e0`

Fresh delta reviews (independent reviewers, one per round) verified the fixes against all three findings and hunted for newly introduced defects.

**Round 1 (`97760f9`):** the unchecked-strdup finding was verified FIXED end-to-end (checks trip only on non-NULL-source → NULL-copy; `utf8_to_wchar_t` returns NULL only on allocation failure; the USB-with-parent attribute-absent exemption is scoped exactly right; the arrival path drops the whole chain on failure and `process_arrival(NULL)`/uncached-path removals are clean no-ops). The TSD-destructor deadlock and quiet-detection fixes were verified for the primary scenario, but the reviewer found a **residual variant**: a `being_joined` *self* entry was still misclassified as "others" — `being_joined` is set *before* `pthread_join`, and the join spans the target's TSD-destructor phase, so "being joined ⇒ terminated ⇒ cannot be self" was a false premise. A monitor thread whose entry was concurrently claimed by an app-thread reaper could still deadlock permanently in its destructor-phase `hid_exit`, and the same window defeated the `quiet` detection.

**Round 2 (`18434e0`)** replaces `pthread_equal`-based self-detection with a thread-local generation id, and the round-2 review confirms **RESIDUAL CLOSED, 0 new defects**:
- Publication ordering holds: the registering thread holds the mutex continuously from `pthread_create` through id publication, and the new thread's first action is a blocking mutex lock, so its stash happens-after publication; a competing overwrite requires FINISHED, which the thread publishes strictly after its own stash — in every interleaving, including the post-create registration-failure unwind.
- The exit scan classifies purely by id: the caller's own entry is recognized as self regardless of join state, falls through, resets `exiting`, and never frees anything (the only free is in the join claimant's post-join block — no double-free). App-thread `hid_exit` still reaps and joins every non-self entry, waiting only on claimed entries whose claimant is guaranteed to broadcast; the cross-generation two-destructor variant resolves via the `exiting` early-return.
- `hid_internal_on_monitor_thread()` reduces to a thread-identity check with correct semantics for both `quiet` call sites; the one remaining `pthread_equal` (reap self-skip) is short-circuit-guarded by `!being_joined`, so a claimed `pthread_t` is never inspected. Types are uniform (`unsigned long` throughout); thread-locals are zero-initialized per C11, so recycled OS threads cannot inherit a stale id.

Both commits build green across the CI matrix. One pre-existing doc nit surfaced in passing (the cleanup comment about a slot staying FINISHED describes branches that can no longer occur with the pre-allocated retirement node) — cosmetic only, noted for a future pass.

PR #822 is review-clean from my side.

_Drafted with Claude Code._

Youw:

Two independent reviews (Claude Fable 5.1 and gpt-6-astra), merged and deduplicated: **28 findings** (1 major, 6 minor, 7 nit, 14 note).

- 179dbfb - lifecycle and error-handling findings (round 1)
- bec001f - named `timespec` for the poll pacing (the C++ build rejected the compound-literal address)
- b077b35 - join hardening (a failed join no longer spins `hid_exit()`) and only transport-level `udev_monitor_receive_device()` errors are treated as fatal

Final adversarial verification (Fable, on the final state): 14 fixed; 1 partial - linux-4, the reused-devnode window inside the arm-then-snapshot interval is documented (README and the `process_arrival()` comment) rather than fenced by syspath/seqnum, the fallback the brief allowed; 1 not fixed - linux-18, no TSan stress regression was added (note-level). The two notes the verifier raised (the `hid_exit()` spin after a failed join, per-message udev parse errors killing the monitor) are the ones fixed in b077b35.

CI: green on b077b35; the device-backed suites pass on the aggregate branch (#830).

_Drafted with Claude Code._

PR #831, opened 2026-07-26
cmake: provide generic hidapi pkg-config alias

### LINKED ISSUE #638: `hidapi` is not available via pkg-config on BSD systems when built with CMake
Apparently CMake build scripts generate `hidapi-libusb` package (same way as on Linux), but not `hidapi` as autotools used to do it.
Probably best to provide `hidapi` package always, even when more than one backend is available.

giri256:
Provide a generic `hidapi.pc` package for the selected default backend while continuing to install each enabled backend's own pkg-config package.

The generic `pc/hidapi.pc.in` template now accepts an explicit library name and is reused by both CMake and Autotools. Backend-specific packages are generated from one parameterized `pc/hidapi-backend-specific.pc.in` template, including dedicated `hidapi-darwin.pc` and `hidapi-winapi.pc` packages. Autotools preserves the existing package names, descriptions, linker flags, and direct `config.status pc/hidapi-*.pc` outputs for compatibility with existing build systems.

Linux CI now verifies that `hidapi.pc` is installed and uses a small standalone Autotools consumer to compile and link through `pkg-config hidapi` when both hidraw and libusb backends are present.

Tests:

* CMake configure/build/install with both Linux backends, plus the standalone pkg-config consumer
* Autotools bootstrap/configure/build/install, direct `config.status` regeneration, and the standalone pkg-config consumer
* Synthetic Darwin, FreeBSD, Windows, and NetBSD pkg-config generation checks
* Native Windows CMake build/install with Visual Studio 2022
* Full PR CI matrix, including the new Ubuntu pkg-config compile/link step
* `git diff --check` and whitespace-insensitive diff audit

Fixes: #638

Assisted-by: ChatGPT:GPT-5.6-Thinking
Assisted-by: codex:gpt-5.6-sol


Youw:
Thanks for looking into this!
I'll update this a bit per my taste, to generalize the solution.

Youw (approved):
Try if this works for you.

Youw:
@giri256 once you confirm latest changes didn't break anything for your flow - I'll have this merged

PR #832, opened 2026-07-28
libusb: add configurable write and report timeouts

mbcinergy:
Add libusb-specific timeout setters for `hid_write()`, `hid_send_output_report()`, and `hid_send_feature_report()`.

Each operation keeps its existing 1000 ms default. A timeout of 0 waits indefinitely, matching libusb semantics. Control-endpoint fallback writes use the configured `hid_write()` timeout, while direct output-report and feature-report operations use their respective settings.

### DIFF
--- a/libusb/hid.c
+++ b/libusb/hid.c
@@ -145,6 +145,8 @@ struct hid_device_ {
 	wchar_t *last_read_error_str;
 
 	unsigned int write_timeout_ms;
+	unsigned int send_output_report_timeout_ms;
+	unsigned int send_feature_report_timeout_ms;
 };
 
 static struct hid_api_version api_version = {
@@ -168,6 +170,8 @@ static hid_device *new_hid_device(void)
 
 	dev->blocking = 1;
 	dev->write_timeout_ms = 1000;
+	dev->send_output_report_timeout_ms = 1000;
+	dev->send_feature_report_timeout_ms = 1000;
 
 	hidapi_thread_state_init(&dev->thread_state);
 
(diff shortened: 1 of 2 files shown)


Youw:
Thanks for the contribution!
This kind of mirrors the `hid_winapi_set_write_timeout`, except `hid_winapi_set_write_timeout` only changes the timeout for `hid_write` while `hid_libusb_set_write_timeout` changes timtout for all of `hid_write`, `hid_send_feature_report` and `hid_send_output_report`.

I get that most likely it doesn't make much sense to configure those separately (does it?), but I don't like the idea of having a similar configuration function (with similar name) but with slightly different behavior. And updating Windows implementation to update all 3 timeouts is not as trivial as one might expect.

@mbcinergy what are you thoughs on this?

mbcinergy:
I thought it made sense to use it for all writing functions, as the name suggests. The function i need it in is actually `hid_send_output_report`, which for libusb backend is also used by `hid_write`. So they are more or less connected together functionality wise. 

I don't think it makes sense to use different timeouts for `hid_send_feature_report` either. We could leave it at fixed 1000ms, but IMO that would be strange behaviour as well.

With winapi backend there is another thing, because the `hid_send_output_report` functions uses `HidD_SetOutputReport` which has a fixed 5000ms timeout. This is already different than the default 1000ms for libusb backend. Not sure how you could change that.

We could add a separate timeout parameter for the reading functions, but that might be something for another PR. I have left the reading functions with the default 1000ms for now (and we already have `hid_read_timeout`).

Youw:
> functions uses HidD_SetOutputReport which has a fixed 5000ms timeout. ... Not sure how you could change that.

That can be changed by using DeviceIOControl directly, instead of using HidD_SetOutputReport  wrapper, but that would be a separate change

> hid_send_output_report, which for libusb backend is also used by hid_write.

I don't believe that's the case. It is for macOS backend, but for libusb backend all 3 functions have their own implementation with a separate `libusb_control_transfer` invocation


---
Lets introduce 3 separate:
- `hid_libusb_set_write_timeout`
- `hid_libusb_set_send_output_report_timeout`
- `hid_libusb_set_send_feature_report_timeout`

This way it will be a 100% clear what is going on and what it controls.
And we can match the implementation for other backends in the future.

mbcinergy:
for libusb backend the `hid_write` uses the `hid_send_output_report` function when there is no interrupt out endpoint. 

I'm fine with adding separate timeout functions, but in this case `hid_write` would actually use the output_report_timeout instead when there is no interrupt out endpoint.

Youw:
> uses the hid_send_output_report function when there is no interrupt out endpoint

Ah right, that case...
Then lets extract `static int hidapi_internal_send_output_report` which explicitly accepts the timeout parameter. And use it from both `hid_write` and `hid_send_output_report`

Youw (approved):
This looks fine. I'll run AI review before merging

Youw:
Interesting fact: I presume this entire code was handwritten by a human (@mbcinergy), but I asked codex to update the PR description and make a merge-commit, so it did and added codex annotation to the commit message body.
That is technically correct, but no code was actually touched by codex in this instance.

mbcinergy:
Yes it was written by a simple mortal human being (a.k.a. me). Thanks for the quick merge!

